PR #24, opened 2018-08-23
feat(table): refactor table to new styles

zvonimirfras:


src/table/table.component.ts
```diff
@@ -328,12 +341,11 @@ import { getScrollbarWidth } from "../common/utils";
...
 	/**
 	 * Size of the table rows.
-	 * (size `"default"` is being deprecated as of neutrino v1.2.0, please use `"md"` instead)
 	 *
-	 * @type {("sm" | "md" |"default" | "lg")}
+	 * @type {("sm" | "md" | "lg")}
 	 * @memberof Table
 	 */
-	@Input() size: "sm" | "md" |"default" | "lg" = "md";
+	@Input() size: "sm" | "md" | "lg" = "md";
```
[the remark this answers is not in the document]
zvonimirfras: It is, I have all 3 in stories.

src/table/table-model.class.ts
```diff
@@ -285,11 +319,20 @@ export class TableModel {
 		const rri = this.realRowIndex(index);
 		this.data.splice(rri, 1);
 		this.rowsSelected.splice(rri, 1);
+		this.rowsExpanded.splice(rri, 1);
 		this.rowsContext.splice(rri, 1);
```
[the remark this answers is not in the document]
zvonimirfras: This function is fairly short and the same code isn't repeated anywhere else so there's no real benefit from it now.

With the new new designs and feature set, we might need to refactor model as well to catch up with the features and then it's worth looking at.

src/table/table.component.ts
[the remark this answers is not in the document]
cal-smith: Must have missed that, cool :+1:

src/table/table.stories.ts
```diff
@@ -0,0 +1,268 @@
...
+	sort(model, index: number) {
+		if (model.header[index].sorted) {
+			// if already sorted flip sorting direction
+			model.header[index].ascending = model.header[index].descending;
+		}
+		model.sort(index);
+	}
+
+	addRow() {
+		const lastRowCopy = JSON.parse(JSON.stringify(this.model.row(this.model.data.length - 1)));
```
theiliad: Should we move this into a Tools file?

This is what I'm doing in charts
image

src/table/table.stories.ts
```diff
@@ -0,0 +1,268 @@
...
+		this.model.data = [
+			[
+				new TableItem({data: "Name 1", expandedData: "No template"}),
+				new TableItem({data: {name: "Lessy", link: "#"}, template: this.customTableItemTemplate})
+			],
+			[new TableItem({
+				data: "Name 3",
+				expandedData: {name: "In", surname: "Template"},
+				expandedTemplate: this.customTableItemTemplate
+			}), new TableItem({data: "swer"})],
```
theiliad: New line after [ & before ] for all entries?

src/table/table.stories.ts
```diff
@@ -0,0 +1,268 @@
...
+				[striped]="striped">
+			</app-custom-table>
+		`,
+		props: {
+			size: selectV2("size", {Small: "sm", Normal: "md", Large: "lg"}, "md", "table-size-selection"),
+			showSelectionColumn: boolean("showSelectionColumn", true),
+			striped: boolean("striped", true)
+		}
+	}));
+
```
theiliad: Can also do this to distinguish each story more:
image

src/checkbox/checkbox.component.ts
```diff
@@ -56,25 +54,23 @@ export class CheckboxChange {
...
+			[required]="required"
+			[checked]="checked"
+			[disabled]="disabled"
+			[indeterminate]="indeterminate"
+			[attr.aria-label]="ariaLabel"
+			[attr.aria-labelledby]="ariaLabelledby"
+			[attr.aria-checked]="indeterminate ? 'mixed' : checked"
+			(change)="onChange($event)"
+			(click)="onClick($event)">
+		<label [for]="id" class="bx--checkbox-label"><ng-content></ng-content></label>
```
theiliad:
```html
<label [for]="id" class="bx--checkbox-label">
   <ng-content></ng-content>
</label>
```
?

src/table/table.stories.ts
zvonimirfras: Can, shouldn't in this case. This follows what other stories in this repo do.

We can consider refactoring all of them.

PR #25, opened 2018-08-27
Added documentation to input component

vRuslan:

Added documentation to input component 

#### Changelog

**New**

* 

**Changed**

* src/input/input.directive.ts

**Removed**

*

### DIFF
--- a/src/input/input.directive.ts
+++ b/src/input/input.directive.ts
@@ -1,7 +1,7 @@
 import { Directive, HostBinding } from "@angular/core";
 
 /**
- * A Component for applying styling to a input element.
+ * A directive for applying styling to an input element.
  *
  * Example:
  *
--- a/src/input/input.directive.ts
+++ b/src/input/input.directive.ts
@@ -1,6 +1,7 @@
 import { Directive, HostBinding } from "@angular/core";
+
 /**
- * A convinence directive for applying styling to a input element.
+ * A Component for applying styling to a input element.
  *
  * Example:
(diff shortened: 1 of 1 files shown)


src/input/input.directive.ts
```diff
@@ -1,5 +1,15 @@
 import { Directive, HostBinding } from "@angular/core";
-
+/**
+ * A convinence directive for applying styling to a input element.
```
magedhennawy: `A convinence directive` is not necessary. 
Maybe replace that with `A component`

src/input/input.directive.ts
```diff
@@ -1,5 +1,15 @@
 import { Directive, HostBinding } from "@angular/core";
-
```
magedhennawy: Can we have an empty line here?

src/input/input.directive.ts
magedhennawy: There's a typo in all documentation that has `A convinence directive`. 

I will accept this and change the rest after it goes through

vRuslan: Ok I'll change

src/input/input.directive.ts
```diff
@@ -1,5 +1,16 @@
 import { Directive, HostBinding } from "@angular/core";
 
+/**
+ * A Component for applying styling to a input element.
```
zvonimirfras: Should be `A directive for applying styling to an input element.`

PR #26, opened 2018-08-28
chore: Improve code style

zvonimirfras:


asudoh:
Nice change! Weighing in a bit, any thoughts on automating this thing? Seems that Prettier 1.4 has TS support - Does thing worth considering? Thanks!

PR #27, opened 2018-08-31
Added a step to set up fonts

stanislavgeorgiev:
Added a new step to redefine the default fonts path to prevent any compile warnings and also prevent serving fonts from a CDN.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,14 +11,13 @@ $ cd my-project
 $ npm i --save-dev carbon-components-angular carbon-components @ngx-translate/core
 ```
 
-Then we need to include carbon-components in `src/styles.scss` and set our fonts path:
+Then we need to include carbon-components in `src/styles.scss`:
 
 ```scss
-$font-path: '/assets/fonts/';
 @import "~carbon-components/scss/globals/scss/styles.scss";
 ```
 
-And copy the fonts files from `node_modules/carbon-components/src/globals/fonts` into our app's `assets/fonts` folder.
+> *Note:* For offline usage we'll need to set `$font-path: '~carbon-components/src/globals/fonts';` at the very top of our `src/styles.scss`. This will copy the fonts to our `dist` folder upon successful build. If you like the fonts to be a part of your `assets` folder and not polute the `dist` folder then copy the fonts from `node_modules/carbon-components/src/globals/fonts` into our app's `src/assets/fonts` folder and add `$font-path: '/assets/fonts/';` at the very top of our `src/styles.scss`.
 
 Then set up our translations in `src/app/app.module.ts`:
 
--- a/README.md
+++ b/README.md
@@ -11,13 +11,16 @@ $ cd my-project
 $ npm i --save-dev carbon-components-angular carbon-components @ngx-translate/core
 ```
 
-Then we need to include carbon-components in `src/styles.scss`:
+Then we need to include carbon-components in `src/styles.scss` and set our fonts path:
 
 ```scss
+$font-path: '/assets/fonts/';
 @import "~carbon-components/scss/globals/scss/styles.scss";
 ```
 
-And set up our translations in `src/app/app.module.ts`:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,13 +11,16 @@ $ cd my-project
...
-Then we need to include carbon-components in `src/styles.scss`:
+Then we need to include carbon-components in `src/styles.scss` and set our fonts path:
 
 ```scss
+$font-path: '/assets/fonts/';
 @import "~carbon-components/scss/globals/scss/styles.scss";
 ```
 
-And set up our translations in `src/app/app.module.ts`:
+And copy the fonts files from `node_modules/carbon-components/src/globals/fonts` into our app's `assets/fonts` folder.
```
theiliad: There should be a better way thank copying directly from node_modules?

stanislavgeorgiev:
I can add instructions on how to reference them directly from node_modules.
The fonts will end up copied over directly to the /dist folder though and
create a lot of clutter and that's why I am suggesting this method.

I could include both and explain pros and cons if you like.

Thanks,
Stan.

theiliad:
@stanislavgeorgiev fair point. It'd be good to find out whether `carbon-components` has an alias for fonts, or whether we can add that

README.md
cal-smith: That's not a bad way ... It should be consistent, but I'm wondering if adding a "note: for offline usage use `@ibm/plex` and load fonts from there" or something would be best?

asudoh:
Some background - The default `$font-path` is totally fine (regardless of the warning in discussion here) for evaluation/demo purposes. Carbon’s biggest consumers have their own CDN and adjust `$font-path` accordingly, and we recommend all consumer applications to do the same. Hope it helps!

README.md
theiliad: Would `~carbon-components/src/globals/fonts` work?

stanislavgeorgiev:
Our product doesn't use CDN as it is hosted by the client and not on a
cloud thus we need to embed all required resources with the product and
assume that there is no access to the Internet.

I now see the other use cases and I will update the instructions to include
a hosted option with CDN (default) and an offline mode with copying files.
I think this maybe a candidate for a wiki page instead of dumping it all in
the readme. What do you think?

Thanks,
Stan.

On Sep 2, 2018 6:10 PM, "Akira Sudoh" <notifications@github.com> wrote:

Some background - The default $font-path is totally fine (regardless of the
warning in discussion here) for evaluation/demo purposes. Carbon’s biggest
consumers have their own CDN and adjust $font-path accordingly, and we
recommend all consumer applications to do the same. Hope it helps!

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/IBM/carbon-components-angular/pull/27#issuecomment-417963292>,
or mute the thread

theiliad:
What about `~carbon-components/src/globals/fonts`?

asudoh:
@stanislavgeorgiev While CDN usage is recommended for on-prem apps, too, hosting fonts (and other contents) on a non-CDN server is fine. BTW another consideration defining `$font-path` is cache busting (e.g. ensuring that the font path contains the associated Carbon version).

README.md
cal-smith: Yeah I think the default/encouraged use case should be to use a CDN, and this text should look something like:

> **note:** For offline usage we'll need to copy the fonts from `node_modules/carbon-components/src/globals/fonts` into our app's `assets/fonts` folder. and set `$font-path: '/assets/fonts/';` before our carbon components import.

cal-smith:
IIRC The concern here is for apps that are actually fire-walled from the internet completely ... for that case they fonts would definitely need to be available locally. I'm not going to say it's totally an edge case, but we definitely want to encourage people to use the CDN when they can.

stanislavgeorgiev:
@cal-smith I like your "offline usage" and I moved the instructions into a separate *Note*.

PR #34, opened 2018-09-01
Update to valid SPDX license

theiliad:
https://spdx.org/licenses/


cal-smith:
Ah interesting, good catch!

PR #37, opened 2018-09-04
Add new tiles module with basic and clickable tile components

markewallace:
Closes IBM/carbon-components-angular#36

Add new tiles module with basic and clickable tile components

#### Changelog

**New**

* Add new tiles module with basic and clickable tile components

**Changed**

* Fixed `package.json` so spec files are ignored when linting


src/tiles/clickable-tile.component.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    expect(directiveEl.attributes["href"]).toBe("https://angular.carbondesignsystem.com/");
+  });
+});
+
+@Component({
+	selector: "test-cmp",
+	template: `
+	<ibm-clickable-tile disabled="true" href="https://angular.carbondesignsystem.com/">
+		Test Clickable Tile
+	</ibm-clickable-tile>`,
```
theiliad: We should indent all the templates consistently

src/tiles/clickable-tile.component.ts
```diff
@@ -0,0 +1,73 @@
...
+    data-tile="clickable"
+    [ngClass]="{'bx--tile': true, 'bx--tile--clickable': true, 'bx--tile--is-clicked': clicked}"
+    tabindex="0"
+    (click)="handleClick($event)"
+    (keydown)="handleKeyDown($event)"
+    [href]="href"
+    [attr.aria-disabled]="disabled">
+    <ng-content></ng-content>
+  </a>`,
+	styles: ["a[aria-disabled=\"true\"] { opacity: 0.5; pointer-events: none; }"],
```
cal-smith: This feels like something that should get upstreamed to carbon-components ... @asudoh thoughts?

cal-smith: Also this would be more readable as a template string over multiple lines

src/tiles/clickable-tile.component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+	ngOnInit() { }
+
+	handleClick(event) {
+		this.clicked = !this.clicked;
+	}
+
+	handleKeyDown(event) {
+		if (event.which === ClickableTile.CARRIAGE_RETURN ||
+			event.which === ClickableTile.SPACE) {
```
cal-smith: [Use `event.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) with strings instead ... [`event.which` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which)

src/tiles/clickable-tile.component.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @export
+ * @class ClickableTile
+ * @implements {OnInit}
+ */
+@Component({
+	selector: "ibm-clickable-tile",
+	template: `
+  <a
+    data-tile="clickable"
```
cal-smith: we don't need the data attributes since we don't use the carbon js directly

src/tiles/tile.component.ts
```diff
@@ -0,0 +1,35 @@
...
+    <div class="bx--tile">
+      <ng-content></ng-content>
+    </div>`,
+	styles: [],
+})
+export class Tile implements OnInit {
+
+	constructor() { }
+
+	ngOnInit() { }
```
cal-smith: Remove the empty `ngOnInit` and `constructor`

src/tiles/clickable-tile.component.ts
```diff
@@ -0,0 +1,73 @@
...
+	 * @type {boolean}
+	 * @memberof ClickableTile
+	 */
+	@Input() disabled = false;
+
+	clicked = false;
+
+	constructor() { }
+
+	ngOnInit() { }
```
cal-smith: We can remove the empty `ngOnInit` and `constructor`

src/tiles/tile.component.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @class Tile
+ * @implements {OnInit}
+ */
+@Component({
+	selector: "ibm-tile",
+	template: `
+    <div class="bx--tile">
+      <ng-content></ng-content>
+    </div>`,
+	styles: [],
```
cal-smith: remove the empty `styles` block

package.json
```diff
@@ -8,7 +8,7 @@
     "storybook": "start-storybook -p 6006",
     "docs:build": "compodoc src -p tsconfig.json",
     "docs:server": "compodoc src -p tsconfig.json -s -w",
-    "lint": "tslint -e spec.ts src/**/*.ts && stylelint src/**/*.scss --config .stylelintrc.yml",
+    "lint": "tslint -e 'src/**/*spec.ts' src/**/*.ts && stylelint src/**/*.scss --config .stylelintrc.yml",
```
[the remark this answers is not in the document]
markewallace: Yes without them the `spec.js` files are included in the linting

src/tiles/clickable-tile.component.spec.ts
markewallace: Not clear what I need to change. I used the following file as a reference and my template looks consistent: https://github.com/IBM/carbon-components-angular/blob/master/src/radio/radio.component.spec.ts#L49

src/tiles/clickable-tile.component.ts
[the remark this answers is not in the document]
markewallace: Adding `bx--link` to the list of styles

src/tiles/clickable-tile.component.spec.ts
markewallace: Think this may have been caused by a spaces vs. tabs issue. Is it OK now?

src/tiles/clickable-tile.component.ts
asudoh: @cal-smith Makes sense, as long as our designers agree on “disabled clickable tile” feature. Feel free to send a GH issue our way!

markewallace:
@zvonimirfras @cal-smith @theiliad @magedhennawy this is ready for another review

src/tiles/clickable-tile.component.spec.ts
theiliad: Yes, I commented on this as an example for consistent indenting 🙂

package.json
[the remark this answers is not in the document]
theiliad: Just to confirm, without the single quotation marks `src/**/*spec.ts` would also effect `spec.js` files? Are you sure?

src/tiles/clickable-tile.component.ts
```diff
@@ -0,0 +1,70 @@
...
+ * @export
+ * @class ClickableTile
+ * @implements {OnInit}
+ */
+@Component({
+	selector: "ibm-clickable-tile",
+	template: `
+	<a
+		[ngClass]="{
+			'bx--link': true,
```
zvonimirfras: We normally provide only classes that are conditionally assigned in `ngClass` and other in "regular" class. It kinda says: these classes are programmatically updated.

So you'd have

```html
class="bx--link bx--tile bx--tile--clickable"
[ngClass]="{
	'bx--tile--is-clicked': clicked
}"
```

or something along those lines.

src/tiles/clickable-tile.component.ts
```diff
@@ -0,0 +1,70 @@
...
+		tabindex="0"
+		(click)="onClick($event)"
+		(keydown)="onKeyDown($event)"
+		[href]="href"
+		[attr.aria-disabled]="disabled">
+		<ng-content></ng-content>
+	</a>`
+})
+export class ClickableTile {
+	static CARRIAGE_RETURN = "Enter";
```
zvonimirfras: These are no longer magic values (like `13` was) so are we getting any benefit from assigning it to static variables?

src/tiles/clickable-tile.component.ts
```diff
@@ -0,0 +1,70 @@
...
+
+	/**
+	 * Sets the `href` attribute on the `ibm-clickable-tile` element.
+	 * @type {string}
+	 * @memberof ClickableTile
+	 */
+	@Input() href = "";
+
+	/**
+	 * Sets the `aria-disabled` attribute on the `ibm-clickable-tile` element.
```
zvonimirfras: This description is not really needed but if you wanna keep it, put it after "Set to `true` to disable the clickable tile."
That is what people really need to know about this input when they read the docs.

src/tiles/tiles.stories.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { TilesModule } from "../";
+
+storiesOf("Tiles", module)
+	.addDecorator(
+		moduleMetadata({
+			imports: [
+				TilesModule,
+				TranslateModule.forRoot()
+			],
```
zvonimirfras: Unnecessary `,`

src/tiles/clickable-tile.component.ts
cal-smith: Seconded, the strings alone are pretty self-explanatory I feel

markewallace:
@cal-smith @zvonimirfras latest feedback has been addressed

src/tiles/clickable-tile.component.ts
```diff
@@ -0,0 +1,63 @@
...
+		<ng-content></ng-content>
+	</a>`
+})
+export class ClickableTile {
+	/**
+	 * Sets the `href` attribute on the `ibm-clickable-tile` element.
+	 * @type {string}
+	 * @memberof ClickableTile
+	 */
+	@Input() href = "";
```
zvonimirfras: This should be just `@Input() href: string;`.

With empty string on a page with url like "https://example.com/folder/#/stuff", it points to "https://example.com/folder/"
If it's not initialized, it won't be populated and won't point anywhere, which seems like the intent here.

markewallace:
@zvonimirfras OK done

zvonimirfras:
@cal-smith @theiliad you also requested changes to this at some point, want to take a second look after requests have been addressed?

PR #58, opened 2018-09-12
Add a breadcrumb and breadcrumb item components

markewallace:
Closes IBM/carbon-components-angular#57

Add breadcrumb and breadcrumb item components

Changelog

New
- Add breadcrumb and breadcrumb item components
- Support creation of an overflow menu when the number of breadcrumb items exceeds a threshold

Changed


markewallace:
@cal-smith @zvonimirfras @magedhennawy please review

src/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component } from "@angular/core";
+
+@Component({
+	selector: "ibm-breadcrumb",
+	template: `
+  <nav class="bx--breadcrumb bx--breadcrumb--no-trailing-slash" aria-label="breadcrumb">
+    <ng-content></ng-content>
```
zvonimirfras: `Breadcrumb` should support multiple tiers workflow without developers having to do anything.

multiple tiers breadcrumb

Last item in http://www.carbondesignsystem.com/components/breadcrumb/usage

src/breadcrumb/breadcrumb.item.directive.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Example:
+ *
+ * ```html
+ * <div ibmBreadcrumbItem/>
+ * ```
+ *
+ * See the [vanilla carbon docs](http://www.carbondesignsystem.com/components/text-input/code) for more detail.
+ */
+@Directive({
```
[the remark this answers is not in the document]
markewallace: @cal-smith the reason I opted for a directive is because the carbon styling didn't work when the DOM contains the component element. 

Using a directive I have DOM like this:
```
<ibm-breadcrumb>
  <nav class="bx--breadcrumb">
    <div ibmbreadcrumbitem="" class="bx--breadcrumb-item">
      <a class="bx--link" href="#"> Breadcrumb 1 </a>
    </div>
  </nav>
</ibm-breadcrumb>
```

Using a component I get this:
```
<nav class="bx--breadcrumb">
  <ibm-breadcrumb-item href="#">
    <div class="bx--breadcrumb-item">
      <a class="bx--link" href="#"></a>
    </div>
  </ibm-breadcrumb-item>
</nav>
```
With the above DOM the Carbon styling doesn't work

src/breadcrumb/breadcrumb.component.ts
markewallace: @zvonimirfras the React version doesn't have this built into the `breadcrumb` component. I think it's up to the user of the component to follow this usage pattern. Also we would need to add the overflow component first. This is a component we need so we can work on this and then look at putting together a sample showing this usage pattern. Does that sound OK as a plan?

src/breadcrumb/breadcrumb.item.directive.ts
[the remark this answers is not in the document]
cal-smith: Interesting, they aren't usually so picky ... In that case would removing the inner div and applying the styles directly to `ibm-breadcrumb-itm` work? You have also have to set the `display` property in the host listener, if `bx--breadcrumb-item` doesn't specify it

src/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,18 @@
...
+
+@Component({
+	selector: "ibm-breadcrumb",
+	template: `
+	<nav class="bx--breadcrumb"
+		[ngClass]="{
+			'bx--breadcrumb--no-trailing-slash' : noTrailingSlash
+		}"
+		[attr.aria-label]="ariaLabel">
+			<ng-content></ng-content>
```
zvonimirfras: Breadcrumbs should support multiple tiers without work required from the developer.

It's not clear to me how this accomplishes that?

screen shot 2018-09-13 at 10 46 36 am

src/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,18 @@
...
+	</nav>`,
```
zvonimirfras: Runaway `,`

src/breadcrumb/breadcrumb.component.ts
markewallace: @zvonimirfras the React implementation doesn't handle the overflow for you, see https://github.com/IBM/carbon-components-react/blob/master/src/components/Breadcrumb/Breadcrumb.js
But I agree it makes sense that we just implement this once.

zvonimirfras: Doesn't mean it shouldn't 😃 

It would be ok to merge the incomplete functionality if it's clear how that functionality paves the way towards _adding_ additional functionality (that we know we want) while keeping the same usage pattern/api for the user.

i.e. you have a table, you know that making it zebra striped can be done by adding classes to the table, so you know you can later add an input to the component that controls that.

What would that be in this case?
I was entertaining an idea (not saying it's better or worse than any other ideas) of user providing an array of items (model) to the component that we can then add rendering for. I can see a clean progression from rendering everything in that array with a forEach, to controlled rendering that has the required functionality, at some point in future.
_(It's also a straightforward way to programmatically update your breadcrumbs based on user interaction, which a user could do with current approach by adding a forEach on their own, not a huge deal)_

Reason why I'm bringing this up again is that it's not entirely clear to me, how would we add desired functionality to the component, with a clean implementation, with current approach?

And is that really a superior experience for the user?

I'm asking this from genuine curiosity and care for our users.

cal-smith: We could potentially capture the ibm-breadcrumb components and conditionally render the additional levels into an overflow menu ... I'd prefer that layout over passing an array directly

So

```
<ibm-breadcrumb>
  <ibm-breadcrumb-item>one</ibm-breadcrumb-item>
  <ibm-breadcrumb-item>two</ibm-breadcrumb-item>
  <ibm-breadcrumb-item>three</ibm-breadcrumb-item>
  <ibm-breadcrumb-item>four</ibm-breadcrumb-item>
  <ibm-breadcrumb-item>five</ibm-breadcrumb-item>
</ibm-breadcrumb>
```

Where two and three are in an overflow menu. 

Now, I'd consider this an enhancement, and ultimately if there's no reasonable way to implement it, then an array is a good fallback.

src/breadcrumb/breadcrumb.item.directive.ts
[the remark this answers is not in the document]
markewallace: OK, I have this working now using `ng-template`. Going to investigate supporting the overflow menu before I update the PR.

src/breadcrumb/breadcrumb.component.ts
zvonimirfras: Absolutely agreed. That _would_ be a solution and I'd consider it an enhancement.

The concern here is that if there's no reasonable way to implement it, then we have to break the interface/api.

Now is the time to make that choice, before the code is merged.

Hence the question.

I'm also considering how they would use the component. When would they actually explicitly want to write breadcrumbs out like this?
I always considered them something that's programmatically set (and forget) and not something you would explicitly type into every page. If that's the case, are we just making them to the extra step of writing out the forEach loop into the contents of the breadcrumb?

markewallace: @cal-smith @zvonimirfras I can see the following use cases:

1. Basic form where I want to manually add a breadcrumb to the top of the page. This is what we have in Assess based on our current designs, each static bread crumb contains only two items.
1. Dynamic form where I have an array of objects which contain the href/text for each breadcrumb I want to display. The array could be long so I would want the automatic overflow behaviour.

So I would propose we phase the implementation accordingly with the first step being the Basic form. I can still have an overflow menu with the Basic form but I need to include the components manually as follows:

```html
<ibm-breadcrumb [noTrailingSlash]="noTrailingSlash">
	<ibm-breadcrumb-item href="#">
		Breadcrumb 1
	</ibm-breadcrumb-item>
	<ibm-breadcrumb-item href="#">
		Breadcrumb 2
	</ibm-breadcrumb-item>
	<ibm-breadcrumb-item>
		<ibm-overflow-menu>
			<ibm-overflow-menu-option>
				<ibm-breadcrumb-item href="#">
					Breadcrumb 3
				</ibm-breadcrumb-item>
			</ibm-overflow-menu-option>
			<ibm-overflow-menu-option>
				<ibm-breadcrumb-item href="#">
					Breadcrumb 4
				</ibm-breadcrumb-item>
			</ibm-overflow-menu-option>
			<ibm-overflow-menu-option>
				<ibm-breadcrumb-item href="#">
					Breadcrumb 5
				</ibm-breadcrumb-item>
			</ibm-overflow-menu-option>
		</ibm-overflow-menu>
	</ibm-breadcrumb-item>
	<ibm-breadcrumb-item>
		Breadcrumb 6
	</ibm-breadcrumb-item>
</ibm-breadcrumb>
```

I have updated my PR to support the Basic form but I have found a couple of bugs:

1. This refers to the table but the same happens when I add the overflow to the breadcrumb. https://github.com/IBM/carbon-components-angular/issues/66
1. The overflow icon doesn't rotate when nested in a breadcrumb. This can be reproduced in vanilla carbon https://github.com/carbon-design-system/carbon/issues/42

zvonimirfras: There seems to be a bug in the overflow menu. We're investigating it...

Users shouldn't have to do that manually. It forces them to add logic to distinguish how many items are in there and provide different implementations. Component should do that for them.

markewallace:
@zvonimirfras @cal-smith dynamic overflow is working in the latest PR. There is one limitation however i.e. the text for the breadcrumb must be specified as an attribute. Let me know what you think of this approach. I'll work on some more unit tests to cover the dynamic scenario tomorrow.

src/breadcrumb/breadcrumb-item.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Component({
+	selector: "ibm-breadcrumb-item",
+	template: `
+	<a class="bx--link"
+		href="{{href}}"
+		*ngIf="href; else content">
+		<ng-container *ngTemplateOutlet="content"></ng-container>
+	</a>
+	<ng-template #content>
```
cal-smith: if you exposed this through a property, you could pass this into the dynamic component though an input. Then you can keep the same usage for dynamic and non-dynamic breadcrumbs.

markewallace: @cal-smith I haven't been able to get the text content for `<ng-template #content>`

I tried adding this property `@ViewChild('content', {read: ElementRef}) content: ElementRef;`

When I try `this.content.nativeElement.textContent` it is en empty string.

src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+		}).compileComponents();
+	}));
+
+	beforeEach(() => {
+		fixture = TestBed.createComponent(BreadcrumbComponent);
+		component = fixture.componentInstance;
+		fixture.detectChanges();
+	});
+
+	it("should create", () => {
```
zvonimirfras: it should be `"should work"` to be consistent with other components. It also reads better if it fails :)

src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+		expect(component).toBeTruthy();
+	});
+
+	it("should create a breadcrumb with a single item", () => {
+		const testFixture: ComponentFixture<BreadcrumbComponentTestComponent> =
+			TestBed.createComponent(BreadcrumbComponentTestComponent);
+		testFixture.detectChanges();
+
+		const directiveEl = testFixture.debugElement.query(By.directive(BreadcrumbComponent));
```
zvonimirfras: `directiveElement` please. General guideline is to use full words whenever possible. Extra length will be compiled away anyway but readability stays 💪

src/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,75 @@
...
+	@Input() overflowThreshold = 4;
+
+	@ContentChildren(BreadcrumbItemComponent) items !: QueryList<BreadcrumbItemComponent>;
+
+	first: BreadcrumbItemComponent = null;
+	secondLast: BreadcrumbItemComponent = null;
+	last: BreadcrumbItemComponent = null;
+	overflowItems: BreadcrumbItemComponent[] = [];
+
+	noOverFlow() {
```
zvonimirfras: "overflow" is one word _(the excess or surplus not able to be accommodated by an available space)_, so it should be `noOverflow`

zvonimirfras: This is also a good candidate for a getter (`get noOverflow()...`) and probably be reversed `shouldOverflow`

src/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,75 @@
...
+export class BreadcrumbComponent implements AfterContentInit {
+	@Input() noTrailingSlash = false;
+
+	@Input() ariaLabel: string;
+
+	@Input() overflowThreshold = 4;
+
+	@ContentChildren(BreadcrumbItemComponent) items !: QueryList<BreadcrumbItemComponent>;
+
+	first: BreadcrumbItemComponent = null;
```
zvonimirfras: These 4 should really be getters returning parts of `items` array. There are more reasons for that, the most prominent being: what happens when `items` change on the fly?

Getters could use (conditionally) indexes, or `slice`s

src/breadcrumb/breadcrumb.stories.ts
```diff
@@ -0,0 +1,87 @@
...
+			</ibm-breadcrumb-item>
+			<ibm-breadcrumb-item>
+				Breadcrumb 3
+			</ibm-breadcrumb-item>
+		</ibm-breadcrumb>`,
+		props: {
+			noTrailingSlash: boolean("noTrailingSlash", true)
+		}
+	}))
+	.add("Static Overflow", () => ({
```
zvonimirfras: We shouldn't encourage this usecase and this story should be removed now that the full solution is available.

src/breadcrumb/breadcrumb.stories.ts
```diff
@@ -0,0 +1,87 @@
...
+			</ibm-breadcrumb-item>
+			<ibm-breadcrumb-item>
+				Breadcrumb 5
+			</ibm-breadcrumb-item>
+		</ibm-breadcrumb>`,
+		props: {
+			noTrailingSlash: boolean("noTrailingSlash", true)
+		}
+	}))
+	.add("Dynamic Overflow", () => ({
```
zvonimirfras: This is the only story we really need and this should be the basic (unless we really want the basic for some reason).

Bonus for providing a knob controlling the number of items (`ngFor` them).
There's a pythonic `range` function available with `import { range } from "../common/utils";` that will help with that. It can be used as `*ngFor="let i of range(itemCount)"`. You'll need to wrap it to use it in the template:
```typescript
range(stop: number, start = 0, step = 1) {
	return range(stop, start, step);
}
```

By controlling the number of items and a threshold (which should default to the number spec supports (3?)), users can experience the full range of possibilities of the component.

src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+	beforeEach(async(() => {
+		TestBed.configureTestingModule({
+			imports: [FormsModule],
+			schemas: [CUSTOM_ELEMENTS_SCHEMA],
+			declarations: [
+				BreadcrumbComponent,
+				BreadcrumbItemComponent,
+				BreadcrumbComponentTestComponent
+			],
```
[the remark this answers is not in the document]
markewallace: @zvonimirfras I will fix this but the eslint rules for the project have this
```
    "comma-dangle": [
      2,
      {
        "arrays": "always-multiline",
        "objects": "always-multiline",
        "imports": "always-multiline",
        "exports": "always-multiline",
        "functions": "never"
      }
    ],
```

src/breadcrumb/breadcrumb.component.ts
markewallace: Leaving as `noOverflow` for readability

This `if (this. noOverflow)`

Rather than this `if (!this.shouldOverflow)`

src/breadcrumb/breadcrumb-item.component.ts
markewallace: @cal-smith @zvonimirfras I'm still not clear on what changes you want made here. All other feedback has been addressed.

src/breadcrumb/breadcrumb.component.spec.ts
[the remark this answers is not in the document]
zvonimirfras: Oh, that looks wrong... 😮

src/breadcrumb/breadcrumb-item.component.ts
cal-smith: So if you get the `ng-template` TemplateRef with a `ViewChild` (or `ContentChild`?), and change the `Input` to take a `TemplateRef` you can pass down the whole template instead of just the text.

so instead of  `<ng-container *ngTemplateOutlet="content"></ng-container>` you'd want something like `<ng-container *ngTemplateOutlet="getContentTemplate()"></ng-container>` where `getContentTemplate` does something like:

```typescript
@ViewChild('content') template: TemplateRef;
@Input() contentTemplate: TemplateRef; // replaces @Input() text

public getContentTemplate() : TemplateRef {
    if (this.contentTemplate) {
        return this.contentTemplate;
    }
    return this.template;
}
```

cal-smith: Basically let the user pass in a whole template, but by default use our template that wraps `ng-content`

markewallace: @cal-smith I've tried that but the net result is nothing get's displayed inside the breadcrumb items

src/breadcrumb/breadcrumb.component.spec.ts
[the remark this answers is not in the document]
zvonimirfras: @markewallace Where do you see this? I went to fix it and I can't find it.
Project only has tslint.json that I can see. Even when I search the whole project it comes out empty for "comma-dangle"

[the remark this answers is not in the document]
markewallace: @zvonimirfras sorry I had the carbon-components project in the same workspace and opened the wrong file. But still the linter should be able to detect this error and report it when the pre-commit hook run's.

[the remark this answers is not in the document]
zvonimirfras: @markewallace no problem :)
It should, I added the rule in #99 ;)

markewallace:
Following issue needs to be resolved: https://github.com/IBM/carbon-components-angular/issues/130

src/breadcrumb/breadcrumb-item.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+	@ViewChild("content") template: TemplateRef<any>;
+
+	ngOnInit() {
+		if (!this.contentTemplate) {
+			this.contentTemplate = this.template;
+		}
+	}
+
+	showNothing() {
```
cal-smith: I _think_ I get what this is doing, but a quick doc string would help clarify this

src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+		</ibm-breadcrumb-item>
+	</ibm-breadcrumb>`,
+	entryComponents: [BreadcrumbComponent]
+})
+class NoOverflowComponent { }
+
+@Component({
+	selector: "test-overflow-cmp",
+	template: `
+	<ibm-breadcrumb noTrailingSlash={{true}} overflowThreshold={{4}}>
```
cal-smith: these could be passed in as `[noTrailingSlash]="true"` and `[overflowThreshold]="4"`

markewallace:
@cal-smith changes made

src/breadcrumb/breadcrumb.stories.ts
```diff
@@ -0,0 +1,59 @@
...
+			</ibm-breadcrumb-item>
+			<ibm-breadcrumb-item href="#3">
+				Breadcrumb 3
+			</ibm-breadcrumb-item>
+			<ibm-breadcrumb-item href="#4">
+				Breadcrumb 4
+			</ibm-breadcrumb-item>
+		</ibm-breadcrumb>`,
+		props: {
+			noTrailingSlash: boolean("noTrailingSlash", true)
```
zvonimirfras: What do you think of having a `number` prop/knob which would allow changing item count in breadcrumb component?

That way users would be able to see how the count affect the visuals.

```html
<ibm-breadcrumb [noTrailingSlash]="noTrailingSlash">
	<ibm-breadcrumb-item *ngFor="let i of range(itemCount)" href="#{{i + 1}}">
		Breadcrumb {{i + 1}}
	</ibm-breadcrumb-item>
</ibm-breadcrumb>
```

_(`range` is available in `common/utils`)_

We could still keep both stories for convenience but provide knobs for both.

markewallace: @zvonimirfras I have tried that but cannot get that working. I met with @cal-smith and we decided to make this a follow up item, see issue #130

markewallace:
@cal-smith @zvonimirfras 

- I have updated this PR to provide the option of using a model for the bread crumb items so that they can be dynamically configured. 
- I have added two knobs to the story to control the number of items in the model and the overflow threshold
- There is still a problem when the number of items changes whereby the overflow menu displays as empty

I have created this PR https://github.com/IBM/carbon-components-angular/pull/264 to test the overflow menu behaviour with a dynamic number of items and this seems to work.

Any ideas as to why it doesn't work in the breadcrumb case?

cal-smith:
We could try using the directive directly, instead of through the component helper ... since we don't need most of the functionality from the component

cal-smith:
I'm good with it as is ... we can get some more eyes on some of the weird parts then

markewallace:
@zvonimirfras anything else needed before this can be merged?

src/breadcrumb/breadcrumb.stories.ts
```diff
@@ -0,0 +1,65 @@
...
+		}
+	}))
+	.add("Model", () => ({
+		template: `
+		<ibm-breadcrumb
+			[noTrailingSlash]="noTrailingSlash"
+			[overflowThreshold]="overflowThreshold"
+		  [breadcrumbItems]="breadcrumbItems(itemCount)">
```
zvonimirfras: Mixed indentation.

src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+		[breadcrumbItems]="_breadcrumbItems">
+	</ibm-breadcrumb>`,
+	entryComponents: [BreadcrumbComponent]
+})
+class TestBreadcrumbModelComponent {
+	_noTrailingSlash = true;
+	_overflowThreshold = 4;
+	_breadcrumbItems = [];
+
+	set noTrailingSlash(val: boolean) {
```
zvonimirfras: Why setters for these?

src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+		const overflowMenuElement = testFixture.debugElement.query(By.directive(OverflowMenu));
+		expect(overflowMenuElement).toBeNull();
+	});
+
+	it("should create a breadcrumb with six items and an overflow menu in second position", () => {
+		const testFixture: ComponentFixture<TestBreadcrumbModelComponent> =
+			TestBed.createComponent(TestBreadcrumbModelComponent);
+		const testComponent = testFixture.componentInstance;
+		testComponent.overflowThreshold = 4;
+		testComponent.breadcrumItems = [
```
zvonimirfras: Test says `six items` but there are 5 here?

src/breadcrumb/breadcrumb-item.component.ts
```diff
@@ -0,0 +1,23 @@
...
+	<a class="bx--link"
+		href="{{href}}"
+		*ngIf="href; else content">
+		<ng-container *ngTemplateOutlet="content"></ng-container>
+	</a>
+	<ng-template #content>
+		<ng-content></ng-content>
+	</ng-template>`
+})
+export class BreadcrumbItemComponent {
```
zvonimirfras: `Component` in the name here is fine since there is also a `BreadcrumbItem` interface which is more likely to be used more often than the component 👍

src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+		<ibm-breadcrumb-item href="#">
+			Breadcrumb 5
+		</ibm-breadcrumb-item>
+	</ibm-breadcrumb>`,
+	entryComponents: [BreadcrumbComponent]
+})
+class TestBreadcrumbComponent {
+	_noTrailingSlash = true;
+
+	set noTrailingSlash(val: boolean) {
```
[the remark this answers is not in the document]
markewallace: AFAIK this is the recommended pattern for testing `@Input()` properties, see https://medium.com/@AikoPath/testing-angular-components-with-input-3bd6c07cfaf6

src/breadcrumb/breadcrumb.component.spec.ts
markewallace: So we can test inputs on the breadcrumb components

markewallace:
@zvonimirfras can you take another look

src/breadcrumb/breadcrumb.component.spec.ts
cal-smith: I'd either expose the whole breadcrumb component or swap out for straight inputs ... as far as I can see there really isn't a requirement to have setters like that ... I'm definitely still learning when it comes to Angular testing though 👍 so if there's something I'm missing let me know

markewallace: @cal-smith 
AFAIK this is the recommended pattern for testing @Input() properties, see https://medium.com/@AikoPath/testing-angular-components-with-input-3bd6c07cfaf6

cal-smith: It's certainly _a_ pattern for testing Inputs, but I don't see anything obvious in that blogpost that _requires_ setters ... In the post for example they just use "old school setters" (`setMethod`) and later abandon them in favor of directly accessing the component. 

If it was wrapping up some additional logic it would be 💯 but all it's doing is `this.prop = value`, so I don't really see the purpose.

src/breadcrumb/breadcrumb.component.spec.ts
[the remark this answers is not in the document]
zvonimirfras: I still see no benefit of this :\ There is already a component that passes its properties to breadcrumb's inputs so that should cover any angular change detection issues.

Do tests fail if you set it directly?

src/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,108 @@
...
+			this._threshold = MINIMUM_OVERFLOW_THRESHOLD;
+		}
+	}
+
+	get threshold(): number {
+		return this._threshold;
+	}
+
+	@Input()
+	set items(items: Array<BreadcrumbItem>) {
```
zvonimirfras: `items` `set`ter/`get`ter combo looks pretty straightforward and if it's all they do, probably don't have to be getters and setters.

src/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,108 @@
...
+				{{secondLast?.content}}
+			</ibm-breadcrumb-item>
+			<ibm-breadcrumb-item [href]="last?.href">
+				{{last?.content}}
+			</ibm-breadcrumb-item>
+		</ng-template>
+	</nav>`
+})
+export class Breadcrumb {
+	_threshold: number;
```
zvonimirfras: Should be
```suggestion
	protected _threshold: number;
```

unless there's a reason for it to be public.

markewallace:
Closing this due to issues resolving `package-lock.json` will create a new PR

PR #60, opened 2018-09-12
fix(toast): add toast to banner.service and enable the close event

### LINKED ISSUE #56: Can't close toast notifications

In https://github.com/IBM/carbon-components-angular/blob/master/src/banner/toast.component.ts
`bx--toast-notification__close-button` doesn't have a click listener like https://github.com/IBM/carbon-components-angular/blob/master/src/banner/banner.component.ts does.

bug

Toast notifications

cal-smith:
closes #56


src/banner/banner.service.ts
```diff
@@ -85,14 +76,15 @@ export class BannerService implements OnDestroy {
 	 * @param {any} [bannerComp=null] If provided, used to resolve component factory
 	 * @memberof BannerService
 	 */
-	showBanner(bannerObj: BannerContent | NotificationContent | ToastContent, bannerComp = null): Banner {
+	showBanner(bannerObj: BannerContent | NotificationContent | ToastContent, bannerComp = null) {
+		let componentFactory;
 		if (!bannerComp) {
```
zvonimirfras: This would read better as "if there's a banner component, use it, otherwise use something else" instead of the negated version we have here.

And "useful" words ;)

src/banner/banner.stories.ts
```diff
@@ -32,11 +31,35 @@ class BannnerStory {
...
+@Component({
+	selector: "app-toast-story",
+	template: `
+		<button class="bx--btn bx--btn--primary" (click)="showToast()">Show info toast</button>
+		<div class="banner-container"></div>
+	`,
+	providers: [BannerService]
+})
+class ToastStory {
+
```
zvonimirfras: Empty line.

src/banner/banner.service.ts
cal-smith: I realized we could actually just pass a default value of `Banner` and skip the ifs ... so that's coming up.

src/banner/banner.stories.ts
cal-smith: I'm changing it, but I'm not sure I like the constructors hugging the class declaration like that

PR #76, opened 2018-09-14
fix(overflow-menu): fix overflow menu not positioning correctly in all cases

### LINKED ISSUE #66: Overflow menu is positioned incorrectly when added to a table column
## Detailed description

- Take the sample from this PR https://github.com/IBM/carbon-components-angular/pull/67
- Run the storybook and open the `Table with overflow menu` story
- Click in last column to display the overflow menu
- Expect the overflow menu to display in last column but displays in the first column

## Additional information
overflow_in_table

cal-smith:
closes #66


src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -9,16 +10,27 @@ import { Dialog } from "../dialog.component";
...
+	onDialogInit() {
+		/**
+		 *  -20 shifts the menu up to compensate for the
+		 *  extra offset generated by the wrapper component.
+		 *
+		 *  60 shifts the menu right to align the arrow.
+		 * (position service trys it's best to center everything,
+		 * so we need to add some compensation)
+		 */
+		this.addGap["bottom"] = pos => position.addOffset(pos, -20, 60);
```
asudoh: The offset thing tends to be whack-a-mole. To avoid dealing with it for every application’s use case, we make it customizable: https://github.com/IBM/carbon-components/blob/v9.14.0/src/components/overflow-menu/overflow-menu.js#L191 - Just a food for thoughts.

src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -9,16 +10,27 @@ import { Dialog } from "../dialog.component";
...
-				class="bx--overflow-menu-options bx--overflow-menu-options--open"
-				tabindex="-1">
-				<ng-template
-					[ngTemplateOutlet]="dialogConfig.content"
-					[ngTemplateOutletContext]="{overflowMenu: this}">
-				</ng-template>
-			</ul>
-		</div>
-		`
+		<ul
```
zvonimirfras: 👍

src/dialog/overflow-menu/overflow-menu-pane.component.ts
zvonimirfras: Good point. This seems to be what's happening at the dialog component level with `addGap` and `dialogConfig`. Overflow menu just extends dialog so it makes sense that it provides offsets for it and removes a line that devs need to write when using it.

PR #84, opened 2018-09-18
Add styling for pill component

teresamonahan:
Closes https://github.com/IBM/carbon-components-angular/issues/83

{{short description}}
There is no style guide in the carbon design system for a pill component so this PR proposes using the Tag styling as suggested by @zvonimirfras.

#### Changelog

**Changed**
* Tag is now a top-level component. 
* The template for the Pill component is now a div and svg which is how it was originally used in the [ComboBox Component](https://github.com/IBM/carbon-components-angular/blob/a931e13ebfaa9cac9b9da233fdd2ac5a7067f9ee/src/combobox/combobox.component.ts). 
The styles we need for the pill component are available in the [Multi-Select Dropdown](http://www.carbondesignsystem.com/components/dropdown/style) section of the design system.
* The SVG for the pill is updated to use the carbon icon.
* Unit tests and stories added for Pill and Tag components.

With the changes in this PR, the pill component looks like this: 
image

And the tags render like this:
image

### DIFF
--- a/src/pill-input/pill.component.ts
+++ b/src/pill-input/pill.component.ts
@@ -2,6 +2,7 @@ import {
 	Component,
 	EventEmitter,
 	Output,
+	OnInit,
 	Input,
 	HostBinding
 } from "@angular/core";
@@ -13,19 +14,30 @@ import {
  * @class Pill
  */
 @Component({
-	selector: "ibm-pill",
+	selector: "ibm-pill, ibm-tag",
 	template: `
 		<span><ng-content></ng-content></span>
-		<button class="pill_close" (click)="doRemove($event)" type="button">
+		<button *ngIf="showCloseIcon" class="pill_close" (click)="doRemove($event)" type="button">
 			<ibm-static-icon icon="x" size="sm" classList="close_icon"></ibm-static-icon>
 		</button>`
 })
-export class Pill {
+export class Pill implements OnInit {
 	/** ListItem to render */
 	@Input() item;
+	/** type of the tag determines the styling */
+	@Input()
+	type: "beta" | "community" | "custom" | "dedicated" | "experimental" | "ibm" | "local" | "private" | "third-party" = "ibm";
+	/** show or hide close icon */
+	@Input()
+	showCloseIcon = true;
 	/** emits an empty event when the close button is clicked */
 	@Output() remove = new EventEmitter();
-	@HostBinding("attr.class") attrClass = "pill";
+	@HostBinding("attr.class") attrClass;
+
+	ngOnInit() {
+		this.attrClass = `pill bx--tag bx--tag--${this.type}`;
+	}
+
 	/** close button handler */
 	public doRemove(ev) {
 		this.item.selected = false;
--- a/src/pill-input/pill.component.ts
+++ b/src/pill-input/pill.component.ts
@@ -2,7 +2,6 @@ import {
 	Component,
 	EventEmitter,
 	Output,
-	OnInit,
 	Input,
 	HostBinding
 } from "@angular/core";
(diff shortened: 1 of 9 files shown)


teresamonahan:
@zvonimirfras Can you please review this first attempt at supporting styling for the pill component as discussed in slack?
I have a couple of concerns about this approach and it'd be great to get your feedback on them:
1. The close icon on the pill needs some additional styling to look right (transparent background, smaller icon size). I am not sure where this styling can be provided as I don't see any other custom styles in this repo and the carbon design system doesn't support a Pill.
2. There is probably a better way to hide the close icon for the tag component rather than always passing `[showCloseIcon]=false`. Is there a way to detect the selector in a component? Then we could only show the close icon if the selector is ibm-pill.
3. Is reusing the pill component for tags the right approach? If there are style changes for the tag, they will have an impact on the pill, so this would need to be monitored.

src/pill-input/pill.component.ts
```diff
@@ -13,19 +14,30 @@ import {
  * @class Pill
  */
 @Component({
-	selector: "ibm-pill",
+	selector: "ibm-pill, ibm-tag",
 	template: `
 		<span><ng-content></ng-content></span>
-		<button class="pill_close" (click)="doRemove($event)" type="button">
+		<button *ngIf="showCloseIcon" class="pill_close" (click)="doRemove($event)" type="button">
 			<ibm-static-icon icon="x" size="sm" classList="close_icon"></ibm-static-icon>
```
teresamonahan: @zvonimirfras ibm-static-icon uses the peretz icons by default. Are there plans to update it to use carbon icons?

zvonimirfras: Yes, @cal-smith is working on the alternative.
In the mean time, replacing `svg`s in static-icon.component would do the trick. Cal can say is it worth doing at this time or not.

zvonimirfras:
It looks great, thank you for this :)

1. Using a different element (like a `span`) instead of the button would go a long way to improving the look of the close button.
2. I couldn't find a clean way for it when I was looking to detect used selector a little while ago.
An OK solution is to have one class extend the other.
Given that these templates are really small, we could ditch the input completely, have the pill extend tag, and add the close/remove inputs and outputs. What do you think?
3. As long as they share style, it make sense.

zvonimirfras:
@teresamonahan That's great! I'm on it! 😃

teresamonahan:
@zvonimirfras @cal-smith Any more thoughts on these changes?

zvonimirfras:
@teresamonahan This brought up a discussion with our design team who want this formalized so we don't have to redo it. I'll get back to you when I have more info (soon).

teresamonahan:
@zvonimirfras I'm finishing up tomorrow for maternity leave so I won't be able to contribute any more work on this. It is a feature we still require for our product though so please help us to get it resolved as soon as you can. Should I close this or leave it open for reference?

zvonimirfras:
Understood. Good luck, @teresamonahan! :)

Leave it open, we'll close it when design is resolved.

nicoleroppel:
Hi! After chatting with Virginia and Anna from design, this is the spec we'd like for the default version of the pill.

Thanks!

screen shot 2018-12-07 at 2 51 38 pm

panpan-lin:
Is there any plan for the Tag / Pill to come in? Need the Beta tag.

zvonimirfras:
@panpan-lin Design was blocking this. With designs speced out, this looks ready to get in order.
There are some conflicts in the PR at the moment and the design in the screenshot doesn't match the design.
That's something @teresamonahan might get on when she finds time.

markewallace:
@zvonimirfras Teresa is on leave for the next 10 months so this PR can be closed until we can find someone to take it over

cal-smith:
We can leave it for now ... but I'll stick a help wanted label on it, when someone picks it up they can just pull from Teresa's branch and we can close this

cal-smith:
I'm going to close this since we have a set of tag components now

PR #98, opened 2018-09-20
Add grid system

imandyie:
Closes https://github.com/IBM/carbon-components-angular/issues/97

Add support for grid from Carbon Design System to be able to create a responsive design.

You can use it like:
```
<div ibmGrid>
  <div ibmRow>
    <div ibmCol columnNumbers='{"lg":3}'></div>
    <div ibmCol columnNumbers='{"lg":3}'></div>
    <div ibmCol columnNumbers='{"lg":3}'></div>
    <div ibmCol columnNumbers='{"lg":3}'></div>
  </div>
</div>
```


src/grid/grid.directive.ts
```diff
@@ -0,0 +1,86 @@
...
+
+@Directive({
+  selector: '[ibmGrid]',
+})
+export class GridDirective implements OnInit {
+  @Input()
+  class = '';
+
+  @HostBinding('class')
+  gridClasses = 'bx--grid';
```
cal-smith: I think `@HostBinding("class.bx--grid") = true` might give you the functionality you expect here, without needing to manually build/check the class input.

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,86 @@
...
+  @HostBinding('class')
+  columnClasses: string;
+
+  constructor() {}
+
+  ngOnInit() {
+    const classNames = [];
+    try {
+      const columns = JSON.parse(this.columnNumbers);
+      const offsets = JSON.parse(this.offsets);
```
cal-smith: This does work ... but I'm wondering if asking for the columns/offsets as `[columns]="{}"` `[offsets]="{}` would be better? Then you could pass down a column/offset config as an object, without having to `JSON.stringify` it first

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,86 @@
...
+  @Input()
+  columnNumbers = '{}';
+
+  @Input()
+  offsets = '{}';
+
+  @HostBinding('class')
+  columnClasses: string;
+
+  constructor() {}
```
cal-smith: Empty constructors can be removed 😀

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,86 @@
+import { Directive, HostBinding, Input, OnInit } from '@angular/core';
```
cal-smith: Hm, if we're doing everything in one file I'd be tempted to call it `grid.directives.ts` ... unless that would add confusion?

Generally things have been one component/directive per file, but I'm not opposed to a multi-directive file if it makes sense

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,86 @@
...
+
+@Directive({
+  selector: '[ibmRow]',
+})
+export class RowDirective implements OnInit {
+  @Input()
+  class = '';
+
+  @HostBinding('class')
+  rowClasses = 'bx--row';
```
cal-smith: Same comment as above RE. the `HostBinding`

cal-smith:
So for some reason all my comments were marked as "resolved" when they're all still applicable.

cal-smith:
Let me know what you're thinking @imandyie so we can get this in 👍

cal-smith:
Hi! Do you think you'll be able to get back to this? If not no worries! We're just trying to clear out some of the backlog

imandyie:
@cal-smith changes addressed

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	@Input()
+	offsets = {};
+
+	@HostBinding("class")
+	columnClasses: string;
+
+	ngOnInit() {
+		const classNames = [];
+		try {
```
cal-smith: Just want to double check why this is in a try/catch ...

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+export class RowDirective {
+	@HostBinding("class.bx--row") baseClass = true;
+}
+
+@Directive({
+	selector: "[ibmCol]"
+})
+export class ColumnDirective implements OnInit {
+	@Input()
+	class = "";
```
cal-smith:
```suggestion
	@Input() class = "";
```

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+})
+export class RowDirective {
+	@HostBinding("class.bx--row") baseClass = true;
+}
+
+@Directive({
+	selector: "[ibmCol]"
+})
+export class ColumnDirective implements OnInit {
+	@Input()
```
cal-smith:
```suggestion
```

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+	class = "";
+
+	@Input()
+	columnNumbers = {};
```
cal-smith: Generally we have the decorator on the same line ... I've done up github suggestions for these, if that helps 👍 
```suggestion
	@Input() columnNumbers = {};
```

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+@Directive({
+	selector: "[ibmCol]"
+})
+export class ColumnDirective implements OnInit {
+	@Input()
+	class = "";
+
+	@Input()
```
cal-smith:
```suggestion
```

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+	columnNumbers = {};
+
+	@Input()
+	offsets = {};
```
cal-smith:
```suggestion
	@Input() offsets = {};
```

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+	selector: "[ibmCol]"
+})
+export class ColumnDirective implements OnInit {
+	@Input()
+	class = "";
+
+	@Input()
+	columnNumbers = {};
+
+	@Input()
```
cal-smith:
```suggestion
```

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+	offsets = {};
+
+	@HostBinding("class")
+	columnClasses: string;
```
cal-smith:
```suggestion
	@Input() columnClasses: string;
```

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+	@Input()
+	class = "";
+
+	@Input()
+	columnNumbers = {};
+
+	@Input()
+	offsets = {};
+
+	@HostBinding("class")
```
cal-smith:
```suggestion
```

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+				} else {
+					classNames.push(`bx--col-${key}-${this.columnNumbers[key]}`);
+				}
+			});
+
+			Object.keys(this.offsets).forEach(key => {
+				classNames.push(`bx--offset-${key}-${this.offsets[key]}`);
+			});
+		} catch (err) {
+			console.log(err);
```
cal-smith: I'd go with `error` in this case

```suggestion
			console.error(err);
```

src/grid/grid.directive.ts
imandyie: Just in case `columnNumbers` or `offsets` are not in the right format I was thinking it was good to catch the error and log it.

cal-smith: Yeah that makes sense :+1: probably want to add a clarifying message to the log then ... I'll do up a suggestion

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,53 @@
...
+				} else {
+					classNames.push(`bx--col-${key}-${this.columnNumbers[key]}`);
+				}
+			});
+
+			Object.keys(this.offsets).forEach(key => {
+				classNames.push(`bx--offset-${key}-${this.offsets[key]}`);
+			});
+		} catch (err) {
+			console.error(err);
```
cal-smith: Maybe something like:

```suggestion
			console.error(`Malformed \`offsets\` or \`columnNumbers\`: ${err}`);
```

imandyie:
@cal-smith changes done

src/grid/grid-test.component.ts
```diff
@@ -0,0 +1,11 @@
+/**
```
zvonimirfras: This should probably be in the spec file with other tests, added after imports but before tests start. That's the way other tests in the codebase approach this.

src/grid/grid.directive.spec.ts
```diff
@@ -0,0 +1,77 @@
+import { async, TestBed } from "@angular/core/testing";
+import { By } from "@angular/platform-browser";
+import { TestGridComponent } from "./grid-test.component";
+import { ColumnDirective, GridDirective, RowDirective } from "./grid.directive";
+
+describe("Grid Directive", () => {
```
zvonimirfras:
```suggestion
describe("GridDirective", () => {
```

src/grid/grid.directive.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+			expect(directiveEl).not.toBeNull();
+		});
+	}));
+
+	it("should render a column", async(() => {
+		TestBed.overrideComponent(TestGridComponent, {
+			set: {
+				template:
+				"<div ibmCol [offsets]='{\"md\": 2}' [columnNumbers]='{\"lg\":3, \"md\": \"nobreak\"}' class='test-directive'></div>"
```
zvonimirfras: Template string would make for a cleaner looking template.

src/grid/grid.directive.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+				RowDirective,
+				ColumnDirective
+			]
+		});
+	});
+
+	it("should render a grid", async(() => {
+		TestBed.overrideComponent(TestGridComponent, {
+			set: {
+				template: "<div ibmGrid class='test-directive'></div>"
```
zvonimirfras: What is the class used for here?

src/grid/grid.directive.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+			fixture.detectChanges();
+
+			expect(directiveEl).not.toBeNull();
+		});
+	}));
+
+	it("should render a row", async(() => {
+		TestBed.overrideComponent(TestGridComponent, {
+			set: {
+				template: "<div ibmRow class='test-directive'></div>"
```
zvonimirfras: What is the class used for here?

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,53 @@
...
+})
+export class RowDirective {
+	@HostBinding("class.bx--row") baseClass = true;
+}
+
+@Directive({
+	selector: "[ibmCol]"
+})
+export class ColumnDirective implements OnInit {
+	@Input() class = "";
```
zvonimirfras: How does this work with `columnClasses` host binding?
My impression would be that all classes would be accessible there and this would just be an input that has name collision with html keyword.

src/grid/grid.directive.ts
```diff
@@ -0,0 +1,53 @@
...
+	@Input() offsets = {};
+
+	@HostBinding("class") columnClasses: string;
+
+	ngOnInit() {
+		const classNames = [];
+		try {
+			Object.keys(this.columnNumbers).forEach(key => {
+				if (this.columnNumbers[key] === "nobreak") {
+					classNames.push(`bx--col-${key}`);
```
zvonimirfras: What if input values change? User for some reason wants to re-flow the layout.

Since classes are only set once on init, end user would have to refresh the page.

`get`ter, `set`ter combo for `columnClasses` would do the trick.

src/grid/grid-test.component.ts
imandyie: Done

src/grid/grid.directive.spec.ts
imandyie: Done

src/grid/grid.directive.spec.ts
imandyie: Is used to test that you can add a custom class

src/grid/grid.directive.spec.ts
imandyie: Done

src/grid/grid.directive.spec.ts
imandyie: Removed

src/grid/grid.directive.ts
imandyie: dynamically added classes don't seem to be  available in host binding for directives

src/grid/grid.directive.ts
imandyie: Done

imandyie:
Rebased master

PR #101, opened 2018-09-21
Add loading component

imandyie:
Closes https://github.com/IBM/carbon-components-angular/issues/82

### DIFF
--- a/src/loading/loading.component.ts
+++ b/src/loading/loading.component.ts
@@ -1,0 +1,35 @@
+import { Component, Input } from "@angular/core";
+
+@Component({
+	selector: "ibm-loading",
+	template: `
+		<div class="bx--loading-overlay">
+			<div data-loading class="bx--loading">
+				<svg class="bx--loading__svg" viewBox="-75 -75 150 150">
+					<title>{{title}}</title>
+					<circle cx="0" cy="0" r="37.5" />
+				</svg>
+			</div>
+		</div>`,
+	styles: []
+})
+export class LoadingComponent {
+	@Input()
+	title = "Loading";
+}
+
+@Component({
+	selector: "ibm-small-loading",
+	template: `
+		<div data-loading class="bx--loading bx--loading--small">
+			<svg class="bx--loading__svg" viewBox="-75 -75 150 150">
+				<title>{{title}}</title>
+				<circle cx="0" cy="0" r="37.5" />
+			</svg>
+		</div>`,
+	styles: []
+})
+export class SmallLoadingComponent {
+	@Input()
+	title = "Loading";
+}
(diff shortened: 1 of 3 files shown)


src/loading/loading.component.ts
```diff
@@ -0,0 +1,35 @@
...
+					<title>{{title}}</title>
+					<circle cx="0" cy="0" r="37.5" />
+				</svg>
+			</div>
+		</div>`,
+	styles: []
+})
+export class LoadingComponent {
+	@Input()
+	title = "Loading";
```
cal-smith: Generally for property declarations, we have decorators on the same line as the property, just a style consistency note

src/loading/loading.component.ts
```diff
@@ -0,0 +1,35 @@
...
+	template: `
+		<div class="bx--loading-overlay">
+			<div data-loading class="bx--loading">
+				<svg class="bx--loading__svg" viewBox="-75 -75 150 150">
+					<title>{{title}}</title>
+					<circle cx="0" cy="0" r="37.5" />
+				</svg>
+			</div>
+		</div>`,
+	styles: []
```
cal-smith: Empty `styles` block can be dropped 👍

src/loading/loading.component.ts
```diff
@@ -0,0 +1,35 @@
...
+	template: `
+		<div data-loading class="bx--loading bx--loading--small">
+			<svg class="bx--loading__svg" viewBox="-75 -75 150 150">
+				<title>{{title}}</title>
+				<circle cx="0" cy="0" r="37.5" />
+			</svg>
+		</div>`,
+	styles: []
+})
+export class SmallLoadingComponent {
```
cal-smith: Not a huge benefit, but you _could_ have `SmallLoadingComponent` extend `LoadingComponent` and just change the templates ... Also, generally we go without the `Component` suffix ... so `SmallLoading` instead

src/loading/loading.component.ts
```diff
@@ -0,0 +1,35 @@
...
+})
+export class LoadingComponent {
+	@Input()
+	title = "Loading";
+}
+
+@Component({
+	selector: "ibm-small-loading",
+	template: `
+		<div data-loading class="bx--loading bx--loading--small">
```
asudoh: FWIW `data-loading` is for vanilla component instantiation

zvonimirfras: If styling works the same without it, drop it :)

Container `div`s should normally be dropped and attributes assigned to the component element with `HostBinding`. It creates fewer elements in the DOM and makes apps with better performance.

src/loading/loading.component.ts
```diff
@@ -0,0 +1,35 @@
+import { Component, Input } from "@angular/core";
+
+@Component({
+	selector: "ibm-loading",
+	template: `
+		<div class="bx--loading-overlay">
```
asudoh: My apologies - My "overlay" comment seems to have introduced a confusion. Overlay is an _optional_ element, that comes only with non-small spinner.

cal-smith:
Hey! We got another PR (#200) that also addresses this! Since they cover the same ground (but this one is pretty out of date) I'm going to close this one.

PR #103, opened 2018-09-21
put focus on primary focus element in modal if specified

teresamonahan:
Closes IBM/carbon-components-angular#102

Allows the developer to specify what element within a modal initial focus should go to when the modal is opened.

#### Changelog

**Changed**

* selectorPrimaryFocus property added to the Modal which determines which element to focus.
* ngOnInit updated to find the element with the selectorPrimaryFocus selector and focus is called on it if available

Behavior after this change when the `data-modal-primary-focus` selector is placed on the primary button in the modal footer - note the focus styling on the close button in the modal footer, instead of the previous blue border on the entire modal:

modal-focus

### DIFF
--- a/src/modal/modal.component.ts
+++ b/src/modal/modal.component.ts
@@ -1,11 +1,12 @@
 import { ModalService } from "./modal.service";
 import {
+	AfterViewInit,
 	Component,
 	EventEmitter,
 	HostListener,
 	Input,
-	OnInit,
 	OnDestroy,
+	OnInit,
 	Output,
 	ElementRef,
 	ViewChild
@@ -110,7 +111,7 @@ import { cycleTabs } from "./../common/tab.service";
 		])
 	]
 })
-export class ModalComponent implements OnInit, OnDestroy {
+export class ModalComponent implements AfterViewInit, OnInit, OnDestroy {
 	/**
 	 * Size of the modal to display.
 	 * (size `"default"` is being deprecated as of neutrino v1.2.0, please use `"md"` instead)
@@ -149,6 +150,12 @@ export class ModalComponent implements OnInit, OnDestroy {
 	modalState = "out";
 
 	/**
+	 * An element should have 'data-modal-primary-focus' as an attribute to receive initial focus within the `Modal` component.
+	 * @memberof ModalComponent
+	 */
+	selectorPrimaryFocus = "[modal-primary-focus]";
+
+	/**
 	 * Creates an instance of `ModalComponent`.
 	 * @param {ModalService} modalService
 	 * @memberof ModalComponent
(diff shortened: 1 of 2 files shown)


src/modal/modal.component.ts
```diff
@@ -148,6 +148,12 @@ export class ModalComponent implements OnInit, OnDestroy {
 	 */
 	modalState = "out";
 
+	/**
+	 * An element should have 'data-modal-primary-focus' as an attribute to receive initial focus within the `Modal` component.
+	 * @memberof ModalComponent
+	 */
+	selectorPrimaryFocus = "[data-modal-primary-focus]";
```
cal-smith: since this all passes through the angular template parser we should be okay with just `[modal-primary-focus]` ... thoughts?

src/modal/modal.component.ts
```diff
@@ -161,6 +167,11 @@ export class ModalComponent implements OnInit, OnDestroy {
 	 */
 	ngOnInit() {
 		this.modalState = "in";
+		const primaryFocusElement  = this.modal.nativeElement.querySelector(this.selectorPrimaryFocus);
+		if (primaryFocusElement && primaryFocusElement.focus) {
+			primaryFocusElement.focus();
+			return;
+		}
```
cal-smith: this should be run in an `ngAfterViewInit` since the view isn't guaranteed to exist until then.

src/modal/modal.component.ts
asudoh: Weighing in - I guess it's an attempt to align to our vanilla code, though such alignment is not a hard requirement.

teresamonahan: Yes I see this selector is used in both vanilla and react code so thought it best to use the same here to make it easier for people transitioning between the different implementations.

teresamonahan:
@cal-smith @asudoh Comments addressed. Can you review again please?

src/modal/modal.component.ts
zvonimirfras: We don't use the `data-` prefix anywhere else unless required by the carbon css.

src/modal/modal.component.ts
```diff
@@ -156,11 +163,23 @@ export class ModalComponent implements OnInit, OnDestroy {
...
 	ngOnInit() {
 		this.modalState = "in";
+	}
+
+	/**
+	 * Set document focus to be on the modal component after it is initialized.
+	 * @memberof ModalComponent
+	 */
+	ngAfterViewInit() {
+		const primaryFocusElement  = this.modal.nativeElement.querySelector(this.selectorPrimaryFocus);
```
zvonimirfras: There's an extra space before `=`

src/modal/modal.component.ts
```diff
@@ -156,11 +163,23 @@ export class ModalComponent implements OnInit, OnDestroy {
...
+		if (primaryFocusElement && primaryFocusElement.focus) {
+			primaryFocusElement.focus();
+			return;
+		}
 		this.modal.nativeElement.focus();
```
zvonimirfras: While we're on this topic, when modal is opened, focus goes to first focusable element for accessibility. That code would go here instead focusing the modal element.

teresamonahan:
@zvonimirfras I updated the selector to remove 'data-' and removed the extra space.
I created a new issue focusing the first focusable element by default as I don't have time to work on that right now - https://github.com/IBM/carbon-components-angular/issues/124

PR #131, opened 2018-09-26
 Fix overflow menu styling and update translation

### LINKED ISSUE #92: OverflowMenuOption.type="danger" doesn't turn the item red on hover
## Detailed description

Hovered menu item's background should turn red but it's not.
Same for "disabled".
I think probably styles are not loaded.

youda97:
closes #92

### DIFF
--- a/src/dialog/overflow-menu/overflow-menu-option.component.ts
+++ b/src/dialog/overflow-menu/overflow-menu-option.component.ts
@@ -24,7 +24,6 @@ import {
 			class="bx--overflow-menu-options__btn"
 			role="menuitem"
 			[ngClass]="{
-				'bx--overflow-menu-options__option--danger': type === 'danger',
 				'bx--overflow-menu-options__option--disabled': disabled
 			}"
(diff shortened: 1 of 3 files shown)


src/dialog/overflow-menu/overflow-menu-option.component.ts
```diff
@@ -24,7 +24,6 @@ import {
 			class="bx--overflow-menu-options__btn"
 			role="menuitem"
 			[ngClass]="{
-				'bx--overflow-menu-options__option--danger': type === 'danger',
 				'bx--overflow-menu-options__option--disabled': disabled
```
zvonimirfras: Considering this class is now applied on host, should it be removed from here too?
If yes, this `ngClass` should be removed too.

cal-smith: So since you have `tabindex="-1"` and  `disabled="true"` (it's missing now, so that's be good to add) you can use `${buttonClass}:not([disabled])` instead and drop the disabled class from here

Issue #122, opened 2018-09-25
Table Header: More sorting options


stanislavgeorgiev:
## Detailed description

> Describe in detail the issue you're having.

- [x] have a `sortable` or `sortingEnabled` property to allow sorting for certain columns only, for ex. column 1, 2 and 3 can be sorted but column 4 ("Actions") cannot be sorted thus should not show the sorting arrow at all
- [x] add a `class` property
- [x] add a generic property for user defined data to attach, for ex. `additionalData` property
- [x] all public properties should be initializeable through the constructor (`sorted`, `ascending`, `descending` are unavailable at the moment). Doing that would allow to set an initial default sorting direction

> Is this a feature request (new component, new icon), a bug, or a general issue?

Feature request

> Is this issue related to a specific component?

Table

stanislavgeorgiev:
@cal-smith @zvonimirfras Can you assign this one to me please? I have half of it done already

PR #134, opened 2018-09-27
Table Header: More sorting options #122

stanislavgeorgiev:
Closes IBM/carbon-components-angular#122


src/table/table.stories.ts
```diff
@@ -25,6 +27,50 @@ import {
...
+				data: "hwer",
+				style: {"width": "auto"},
+				sortable: this.sortable
+			})
+		];
+	}
+
+	ngOnChanges(changes: SimpleChanges): void {
+		if (changes.sortable) {
+			for (let column of this.model.header) {
```
zvonimirfras: Ideally, making the whole table sortable/non-sortable would be done by providing/not providing sort function.

Can be checked for with `this.sort.observers.length > 0`

This also requires changes in template where `columns.sortable` is checked.

src/table/table-header-item.class.ts
```diff
@@ -19,6 +19,14 @@ export class TableHeaderItem {
 	 */
 	sorted = false;
 
+	/**
+	 * Enables sorting on click by default.
+	 * If false then this column won't show a sorting arrow at all.
+	 *
+	 * @memberof TableHeaderItem
+	 */
+	sortable = false;
```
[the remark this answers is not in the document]
stanislavgeorgiev: why should it sortable by default?

src/table/table-header-item.class.ts
```diff
@@ -69,6 +84,15 @@ export class TableHeaderItem {
...
 	data: any;
 
+	/**
+	 * Data for the header item for general usage in the controller.
+	 * For example, which `field` is this column related to.
+	 *
+	 * @type {*}
+	 * @memberof TableHeaderItem
+	 */
+	additionalData: any;
```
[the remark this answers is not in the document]
stanislavgeorgiev: This is for non-presentational data that would be used by the controller later on, for example `dbField: "name"` to show which database field that column should be sorting by. The `data` property cannot hold that `dbField` because that would require for me to supply a template for that cell which I would really like not to do as it takes time and it's not relevant to the view at all.

src/table/table.stories.ts
stanislavgeorgiev: You're right that we need a global flag as to whether the table can be sorted at all or not. The idea of the `TableHeaderItem.sortable` is to disable sorting for some of the columns while leaving it for others.

zvonimirfras: We don't need to supply a special global flag, per se, presence of the sort observer (whether or not user said `(sort)="mySortingFunction()"` is a clear indicator that user wants to sort, so we can use that instead as `this.sort.observers.length > 0`

src/table/table-header-item.class.ts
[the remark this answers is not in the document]
stanislavgeorgiev: OK, I changed it. I think at a certain point I set it to `false` because I wanted to test it and then probably never set it back to `true`.

src/table/table-header-item.class.ts
```diff
@@ -28,6 +36,13 @@ export class TableHeaderItem {
 	 */
 	filterCount = 0;
 
+	/**
+	 * Attach a class to the column, applied to the header column cell only.
+	 *
+	 * @memberof TableHeaderItem
+	 */
+	className = "";
```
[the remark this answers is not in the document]
stanislavgeorgiev: Yeap :)

src/table/table.stories.ts
stanislavgeorgiev: I added changes to use `this.sort.observers.length > 0` as a global sorting flag and then tie it further down with the `column.sortable`.
I couldn't make the story nicer because I don't know how to set a conditional `(sort)` and that's why I left the `column.sortable` there.

stanislavgeorgiev:
I'm not sure what to do for that merge conflict. I just pulled master into my master and the my master to the Table branch here, no conflicts, no merges. Any ideas?

cal-smith:
Fixed it in the web ui ... think that worked?

stanislavgeorgiev:
There was a line with spaces in the `table.stories.ts`.
I noticed that sometimes I would pull from master and my branch will start having lint errors.

stanislavgeorgiev:
That's ridiculous, Travis has been stuck in _Waiting for status to be reported_ for 7 hours now...

stanislavgeorgiev:
Any chance of merging this PR soon?

src/table/table-header-item.class.ts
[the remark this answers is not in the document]
zvonimirfras: Make it `className: string;` unless empty string is required default value. This implies things will break if you don't give it an actual value.

src/table/table-header-item.class.ts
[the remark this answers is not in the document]
zvonimirfras: It makes sense to have this in those cases. Name it `metadata`, it's more descriptive what it should be used for.

src/table/table.stories.ts
```diff
@@ -29,6 +31,49 @@ import { clone } from "../utils/utils";
...
+class TableStory implements OnInit, OnChanges {
+	@Input() model = new TableModel();
+	@Input() size = "md";
+	@Input() showSelectionColumn = true;
+	@Input() striped = true;
+	@Input() sortable = true;
+	@Input() simpleSort;
+
+	ngOnInit() {
+
```
zvonimirfras: Unnecessary empty line.

src/table/table.stories.ts
```diff
@@ -29,6 +31,49 @@ import { clone } from "../utils/utils";
...
+		</ibm-table>
+	`
+})
+class TableStory implements OnInit, OnChanges {
+	@Input() model = new TableModel();
+	@Input() size = "md";
+	@Input() showSelectionColumn = true;
+	@Input() striped = true;
+	@Input() sortable = true;
+	@Input() simpleSort;
```
zvonimirfras: Since it doesn't seem feasible to conditionally provide an observer in template, does it still make sense to provide the sorting function as an input?

stanislavgeorgiev: I think it still makes sense because the user still needs to see the sorting in action (which is enabled by default in the story)

src/table/table.stories.ts
stanislavgeorgiev: Updated.
I like the rename to `metadata`. I didn't like the name `additionalData` but couldn't come up with a better name.

src/table/table.stories.ts
zvonimirfras: Sure, but it can be implemented in a TableStory instead provided to it from the outside.
That's the way devs would do it anyway.

Let's change it to that way for the people who copy code from the demos.

_The code from before didn't take a function as an input, it was added as an observer to `(sort)`, which we can no longer simply do because `ibm-table` is no longer directly exposed, so `simpleSort()` needs to move from global to the `TableStory` component (and doesn't need to be a prop)._

PR #135, opened 2018-09-27
Fix overflow menu visibility for table

### LINKED ISSUE #86: Table Overflow menu should stay active while hovering the opened menu
## Detailed description

The overflow menu icon disappears in the [Table storybook](https://angular.carbondesignsystem.com/?knob-size=md&knob-striped=true&knob-showSelectionColumn=true&selectedKind=Table&selectedStory=with%20overflow%20menu&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) as soon as the mouse moves to select an option.

For comparison, (Carbon's Data Table)[http://www.carbondesignsystem.com/components/data-table/code] keeps the overflow icon visible while browsing the menu.

Bug

Table

youda97:
closes #86

### DIFF
--- a/src/dialog/overflow-menu/overflow-menu.directive.ts
+++ b/src/dialog/overflow-menu/overflow-menu.directive.ts
@@ -51,6 +51,25 @@ export class OverflowMenuDirective extends DialogDirective {
 
 	onDialogInit() {
 		this.dialogConfig.content = this.ibmOverflowMenu;
+		this.dialogService.isClosed.subscribe(() => this.updateOpacity());
+	}
+
+	open() {
+		super.open();
+		this.updateOpacity();
+	}
+
+	toggle() {
+		super.toggle();
+		this.updateOpacity();
+	}
+
+	updateOpacity() {
+		if (this.elementRef.nativeElement.getAttribute("aria-expanded") === "true") {
+			this.elementRef.nativeElement.style.opacity = 1;
+		} else {
+			this.elementRef.nativeElement.style.opacity = null;
+		}
 	}
 
 	@HostListener("keydown", ["$event"])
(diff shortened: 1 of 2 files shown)


src/dialog/overflow-menu/overflow-menu.directive.ts
```diff
@@ -51,6 +51,25 @@ export class OverflowMenuDirective extends DialogDirective {
...
+	}
+
+	toggle() {
+		super.toggle();
+		this.updateOpacity();
+	}
+
+	updateOpacity() {
+		if (this.elementRef.nativeElement.getAttribute("aria-expanded") === "true") {
+			this.elementRef.nativeElement.style.opacity = 1;
```
asudoh: Weighing in - Wondering if team wants CSS with attribute selector?

youda97: Well this should be something that needs to be set in the `"bx--overflow-menu--open"` class, which is currently not doing anything

cal-smith: So I'm not sure why this needs to be applied in the first place? Is the overflow menu getting `opacity: 0` in the table? if so why?

youda97:
@cal-smith So currently it does not work in vanilla, as they are just applying a focus to the overflow menu item. Once it is unfocused the menu item will disappear

src/dialog/overflow-menu/overflow-menu.directive.ts
youda97: Well the design is to have the menu item only appear when you hover on it and to only stay visible when the menu is open

asudoh:
Another weigh-in - Would you guys want to open an issue against our vanilla repo if you think the trigger button should be visible while it's open, regardless of the focus state? Meanwhile, do you guys have a way to have your own CSS to override it (instead of using in-line style)?

cal-smith:
For sure ... I think it makes sense to have it explicitly styled ... we should be able to use an inline `styles:[]` property, if not on the directive, then on the wrapper component.

PR #143, opened 2018-10-02
chore(lint): introduce lint-staged and Prettier

asudoh:
(Please use Diff settings - Hide whitespace changes for easier review)

This change introduces `lint-staged` and Prettier. Prettier is not applied to TypeScript code yet, given current version of Prettier adds newlines to all decorators, and it doesn't seem to be what the team here wants.

Let me know if there are any changes/configs that drift from what the team here desires - Thanks!

#### Changelog

**New**

* `lint-staged` running:
    * TSLint for TypeScript files
    * StyleLint and Prettier for SCSS files
    * Prettier for JavaScript/CSS/Markdown/YAML files

**Changed**

* Several format change (esp. YAML files) upon Prettier

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+dist
+node_modules
--- a/.stylelintrc.yml
+++ b/.stylelintrc.yml
@@ -1,195 +1,200 @@
+plugins: stylelint-prettier
+extends: stylelint-config-prettier
 rules:
-  # Colors
-  color-named: never
-  color-no-hex: true
+    # Prettier
+    prettier/prettier: true
 
-  # Fonts
-  font-family-name-quotes: always-where-recommended
-  font-family-no-duplicate-names: true
-  font-weight-notation: numeric
+    # Colors
+    color-named: never
+    color-no-hex: true
 
-  # Functions
-  function-blacklist:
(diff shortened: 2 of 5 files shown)


asudoh:
@cal-smith Your assumption is correct - I deliberately excluded `*.ts` for now because current version of Prettier adds newlines to all decorators, and it doesn't seem to be what the team here wants. Issue number 4924 in https://github.com/prettier/prettier/issues discusses this more in detail.

cal-smith:
Ah cool! I'll check that issue then and we can enable `*.ts` once that gets resolved (we may have to concede some stylistic preference and add/never add newlines to all decorators, but hopefully that's configurable)

asudoh:
Thanks @cal-smith for reviewing! Good to see if other reviewers are interested in this change, too.

cal-smith:
Looks like prettier v1.15 will handle decorators better: https://github.com/prettier/prettier/pull/5188

So let's wait for that, then we can bump the version and enable it for everything!

cal-smith:
Okay cool, looks like prettier is updated! I don't know if you want to update this one @asudoh or create another. FYI @zvonimirfras

cal-smith:
This has been open for a while, and AFAIK perttier still mangles our templates ... so I'm going to close this. We can re-evaluate at some point later.

PR #148, opened 2018-10-02
Feat(Pagination): Add pagination

### LINKED ISSUE #147: Add pagination component

zvonimirfras:
Closes #147


src/i18n/en.json
```diff
@@ -65,12 +65,19 @@
 			"RESET_SEARCH": "Reset search"
 		}
 	},
-	"MODAL" : {
+	"MODAL": {
 		"CLOSE": "Close modal"
 	},
 	"OVERFLOW_MENU" : {
```
magedhennawy: space here

zvonimirfras: Good catch! Looks like there's another one of those in there from way back when.

PR #150, opened 2018-10-02
fix(modal): Fix modal.service:show() `buttons` to be an array.

stanislavgeorgiev:
fix(modal): Added more documentation.
feat(modal): modal.service:show() can now supports HTML in `modalContent`.


src/modal/alert-modal.component.ts
```diff
@@ -64,7 +64,7 @@ import Modal from "./modal.decorator";
       			<p class="bx--modal-header__heading bx--type-beta">{{modalTitle}}</p>
 			</ibm-modal-header>
 			<div class="bx--modal-content">
-				<p>{{modalContent}}</p>
+				<p [innerHTML]="modalContent"></p>
```
cal-smith: I think this is probably okay (I can certainly see some use cases), but a `ng-template` output should probably be added as the default way to inject complex html content.

That of course could be added as a feature later, what do you think?

src/modal/modal.service.ts
```diff
@@ -82,7 +86,7 @@ export class ModalService {
 	 * @returns {ComponentRef<any>}
 	 * @memberof ModalService
 	 */
-	show(data: {modalType?: string, modalLabel?: string, modalTitle: string, modalContent: string, buttons?: null}) {
+	show(data: {modalType?: string, modalLabel?: string, modalTitle?: string, modalContent?: string, buttons?: []}) {
```
cal-smith: nice!

if we wanted to get really fancy we could declare a `interface ModalButton` and type `buttons` to `buttons?: ModalButton[]`

stanislavgeorgiev: yeap, would definitely add that in the future!
I'm still an Angular NOOB but also really pressed for our project's release in 3 weeks

src/modal/alert-modal.component.ts
cal-smith: Yep makes sense, and I think we already support that for banners/notifications (which have a similar usage model)

src/modal/modal.service.ts
cal-smith: Cool, I'll make an issue to note that 👍

PR #156, opened 2018-10-05
Remove ngx-translate and switch to framework agnostic inputs

### LINKED ISSUE #63: Why i18n and @ngx-translate in this library ?
I'm wondering why UI component library at this level handling its own i18n texts and moreover dictating which i18n library to use ? Aren't i18n texts and which i18n library to use are completely for the top level application to provide and decide instead, and this library itself should stay agnostic ? 

References and comparisons : 
- https://github.com/angular-ui/bootstrap
- https://github.com/valor-software/ngx-bootstrap
 
Thank You

cal-smith:
closes #63


src/i18n/en.json
```diff
@@ -65,22 +63,21 @@
 			"RESET_SEARCH": "Reset search"
 		}
 	},
-	"MODAL": {
+	"MODAL" : {
```
zvonimirfras:
```suggestion
	"MODAL": {
```

src/i18n/en.json
```diff
@@ -65,22 +63,21 @@
...
 		"CLOSE": "Close modal"
 	},
-	"OVERFLOW_MENU": {
+	"OVERFLOW_MENU" : {
```
zvonimirfras:
```suggestion
	"OVERFLOW_MENU": {
```

src/pagination/pagination.component.ts
```diff
@@ -149,6 +147,8 @@ export class Pagination {
 	 */
 	@Input() model: PaginationModel;
 
+	@Input() translations = EN.PAGINATION;
```
zvonimirfras: Missing service?
```suggestion
	@Input() translations = this.i18n.get().PAGINATION;
```

src/dialog/overflow-menu/overflow-menu.component.ts
```diff
@@ -43,7 +44,10 @@ import { Component, ElementRef } from "@angular/core";
 	`]
 })
 export class OverflowMenu {
-	constructor(private elementRef: ElementRef) {}
+
```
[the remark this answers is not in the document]
cal-smith: I think it's just your diff ... I don't see that locally

src/pagination/pagination.component.ts
cal-smith: Oops, good catch

cal-smith:
I'm not sure if we want to tackle the calendar stuff now given that we're going to be refactoring it anyway.

I was thinking about some of the potential weirdness with the i18n service ... I think the best fix would be to return an Observable that updates with the correct value ... thinking about it I've got a couple ideas that could work pretty well with that ... and thanks to the `async` pipe it's pretty easy to handle in templates

src/banner/banner.component.ts
```diff
@@ -38,7 +40,7 @@ import { NotificationContent } from "./banner-content.interface";
 			type="button">
 			<svg
 				class="bx--inline-notification__close-icon"
-				aria-label="close"
+				[attr.aria-label]="bannerObj.closeLabel"
```
youda97: this label should be on the button to fix DAP violation: `Form controls must have associated labels`. See #194 for more clarification

src/banner/toast.component.ts
```diff
@@ -28,7 +28,7 @@ import { Banner } from "./banner.component";
 			(click)="onClose()">
 			<svg
 				class="bx--toast-notification-icon"
-				aria-label="close"
+				[attr.aria-label]="bannerObj.closeLabel"
```
youda97: this label should be on the button to fix DAP violation: `Form controls must have associated labels`. See #194 for more clarification

src/banner/banner.component.ts
cal-smith: if it's in that PR I don't need to fix it here as well.

PR #160, opened 2018-10-08
Create an Angular Directive for Carbon link

panpan-lin:
screen shot 2018-10-08 at 18 20 58

 - Create a directive in the carbon-component-angular project to encapsulate the carbon link behaviour, see: http://www.carbondesignsystem.com/components/link/code
 - Support disabling the link per Carbon design
 - Create unit tests and story for the storybook

#### Changelog

**New**

* add new link module with basic link component, with support for disabling the link


src/link/link.directive.ts
```diff
@@ -0,0 +1,29 @@
...
+	selector: "[ibmLink]"
+})
+
+export class Link {
+	@HostBinding("class.bx--link") baseClass = true;
+	/**
+	 * Set to `true` to disable link.
+	 * @memberof Link
+	 */
+	@HostBinding("attr.aria-disabled") @Input() disabled = false;
```
zvonimirfras: Per carbon docs, `disabled` should also make the link not tabable, setting `tabindex="-1"`

This can be done by making this a [`get`ter `set`ter combo](https://www.typescriptlang.org/docs/handbook/classes.html#accessors), where `set`ter will set a private `_disabled` field and set tabindex to -1 if value is true and to null if value is false.
To be able to set tabindex, you may need `@HostBinding("attr.tabindex") tabindex;`

src/link/link.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { TestBed, ComponentFixture } from "@angular/core/testing";
+import { Component, Input, DebugElement } from "@angular/core";
+import { By } from "@angular/platform-browser";
+
+import { Link } from "./link.directive";
+
+describe("Link", () => {
+	it("should create a Link", () => {
+		TestBed.configureTestingModule({
```
zvonimirfras: Use `beforeEach` for setting up testing environment by putting common/repeating things into them. See other components' tests for reference.

src/link/link.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+			declarations: [TestDisabledLinkComponent, Link]
+		});
+
+		let fixture: ComponentFixture<TestDisabledLinkComponent> = TestBed.createComponent(TestDisabledLinkComponent);
+		let component: TestDisabledLinkComponent = fixture.componentInstance;
+		fixture.detectChanges();
+
+		const directiveEl = fixture.debugElement.query(By.directive(Link));
+		expect(directiveEl.attributes["disabled"]).toBe("true");
```
zvonimirfras: `disabled` attribute doesn't really _have to_ be `true`, per carbon specs, it's "accidentally" true because the input name matches it.
`tabindex` should be `-1` par carbon spec, though

src/link/link.spec.ts
panpan-lin: The testbed configuration for the two tests are different. One is using `TestLinkComponent`, the other is using `TestDisabledLinkComponent`.

panpan-lin:
@zvonimirfras ready for another round of review.

src/link/link.directive.ts
```diff
@@ -0,0 +1,47 @@
...
+ * ```hmtl
+ * <a href="#" ibmLink>A link</a>
+ * ```
+ *
+ * See the [vanilla carbon docs](http://www.carbondesignsystem.com/components/link/code) for more detail.
+ */
+@Directive({
+	selector: "[ibmLink]"
+})
+
```
zvonimirfras: Remove the empty line.
```suggestion
```

src/link/link.directive.ts
```diff
@@ -0,0 +1,47 @@
...
+	set disabled(disabled: boolean) {
+		this._disabled = disabled;
+		this.tabindex = this.disabled ? -1 : null;
+	}
+
+	get disabled(): boolean {
+		return this._disabled;
+	}
+
+	private _disabled;
```
zvonimirfras: Let's default it to not disabled.
```suggestion
	private _disabled = false;
```

src/link/link.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+		});
+
+		let fixture: ComponentFixture<TestLinkComponent> = TestBed.createComponent(TestLinkComponent);
+		let component: TestLinkComponent = fixture.componentInstance;
+		fixture.detectChanges();
+
+		expect(component).toBeTruthy();
+		const directiveEl = fixture.debugElement.query(By.directive(Link));
+		expect(directiveEl).not.toBeNull();
+		expect(directiveEl.attributes["aria-disabled"]).toBe(null);
```
zvonimirfras: It looks like this test is failing. `aria-disabled` doesn't really have to be `null`, it just has to be `falsy`, no?
```suggestion
		expect(directiveEl.attributes["aria-disabled"]).toBeFalsy();
```

zvonimirfras: On a second look, setting `attr.aria-disabled` might have to require some additional logic to mask the value.
Test actually seems to compare a string `'false'` and not the boolean value, and a non-empty string is _truthy_.
It seems to be working fine in other places though, so the check may have to be:
```suggestion
		expect(directiveEl.attributes["aria-disabled"]).toBe("false");
```

panpan-lin:
@zvonimirfras Please see my comment above. Thanks!

src/link/link.spec.ts
cal-smith: I'd expect `.toBe(null)` and `.toBeFalsy()` to be equivalent in this case ... falsy attributes are generally expected to be `null`ed out of the DOM anyway.

panpan-lin: @cal-smith Please see question above. Thanks :-)

Issue #106, opened 2018-09-24
Fix DAP issues for Accordion


youda97:
Level | Description
-- | --
V | Elements that use 'role' must reference a valid WAI-ARIA role. | Details
V | All content must reside within a WAI-ARIA landmark or labelled region role.

aradhikaIbm:
Working on possible fix

cal-smith:
closed with #168

PR #165, opened 2018-10-11
Accordion DAP fixes

aradhikaIbm:
Closes IBM/carbon-components-angular #106

Fix DAP issues for Accordion
Set the role to heading and use aria-level per guidelines.
Defaulting level to 3 since must pages have heading level 1 already.
User can override.

#### Changelog

**Changed**

* Set the role to heading and use aria-level per guidelines.

### DIFF
--- a/src/accordion/accordion-item.component.ts
+++ b/src/accordion/accordion-item.component.ts
@@ -36,7 +36,9 @@ export class AccordionItem {
 
 	@HostBinding("class.bx--accordion__item") itemClass = true;
 	@HostBinding("class.bx--accordion__item--active") @Input() expanded = false;
-	@HostBinding("style.display") @HostBinding("attr.role") itemType = "list-item";
+	@HostBinding("style.display") itemType = "list-item";
+	@HostBinding("attr.role") roleType = "heading";
+	@HostBinding("attr.aria-level") @Input() ariaLevel = 3;
 
 	constructor() {
 		AccordionItem.accordionItemCount++;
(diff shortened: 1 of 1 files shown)


aradhikaIbm:
Does stacking the @HostBindings for the display and role not work?  I was trying not impact the css the widget was using. I believe heading is not a valid value for the display css. 

Level 3 was something on I saw the example that they provided in the design. I think it also represents the heading within the page.  So most pages use 1 as the page header.

src/accordion/accordion-item.component.ts
```diff
@@ -36,7 +36,9 @@ export class AccordionItem {
 
 	@HostBinding("class.bx--accordion__item") itemClass = true;
 	@HostBinding("class.bx--accordion__item--active") @Input() expanded = false;
-	@HostBinding("style.display") @HostBinding("attr.role") itemType = "list-item";
+	@HostBinding("style.display") itemType = "list-item";
+	@HostBinding("attr.role") roleType = "heading";
```
zvonimirfras: `itemType` makes sense as a variable name there, `roleType` should be `role`

aradhikaIbm: I'll make that change shortly

zvonimirfras:
It looks like you've overridden all the changes of the previous merge with yours. Please undo that.

Your diff also grew from 3 lines to whole files... Not sure what happened there :\

aradhikaIbm:
Will redo with a new branch. Its not clear to me the state of my branch.

PR #171, opened 2018-10-15
fix for alert modal not having useful types

cal-smith:


src/modal/modal.stories.ts
```diff
@@ -60,20 +61,20 @@ class ModalStory {
...
 	@Input() modalTitle: string;
 	@Input() modalContent: string;
-	@Input() buttons: any;
+	@Input() buttons: Array<ModalButton>;
 
 	constructor(private modalService: ModalService) { }
 
 	openModal() {
 		this.modalService.show({
 			modalType: this.modalType,
```
zvonimirfras: Did you miss `modalType` -> `type` or there's a reason for that? I see you check for it in `show()`.

cal-smith:
I change it to 2016 so typescript would stop complaining about `[].includes`

src/modal/modal.stories.ts
cal-smith: left it for now to show the error you can expect ... I can change it to all be updated

PR #180, opened 2018-10-17
fix(dialog): stop using aria-expanded since it isn't valid for tooltips

### LINKED ISSUE #119: Fix DAP issues for Tooltip
Level | Description
-- | --
V | The WAI-ARIA role(s) and/or attribute(s) aria-expanded are not valid for the element span. See rule help for guidance. | Details

cal-smith:
closes #119


zvonimirfras:
This doesn't feel right. [Tooltips _do_ use aria-expanded as a way to communicate to screen readers that an action happened.](https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/expanded/)
The problem DAP's reporting seems to be that `span` is used to trigger a tooltip.

cal-smith:
Well that's a useful site ...

cal-smith:
So `aria-expanded` isn't valid for anything but buttons and anchors ... in the case of dialog there really isn't anything we can do to prevent users from putting the element on anything and triggering a DAP failure

cal-smith:
The best we can do is have `overflow-menu` implement tooltip a11y correctly, and just leave it up to other teams to deal with the (potentially false) DAP errors when they apply a dialog to an unsupported element

cal-smith:
On further investigation it turns out that `aria-expanded` is only valid for *`role="button"`*. This means we can simply apply the correct roles and attributes to _any_ element and have a fully accessible tooltip! With zero extra effort on the part of the consumers!

src/dialog/tooltip/tooltip.directive.ts
```diff
@@ -64,6 +65,8 @@ export class TooltipDirective extends DialogDirective {
 	// tslint:disable-next-line:no-input-rename
 	@Input("tooltip-type") tooltipType: "warning" | "error" | "" = "";
 
+	@HostBinding("attr.aria-describedby") descriptorID = "";
```
zvonimirfras:
```suggestion
	@HostBinding("attr.aria-describedby") descriptorId: string;
```

`""` implies that it's a special value, or that it needs a value, so if it's not and it doesn't we can leave it undefined.
`Id` is what we most often use in the codebase.

src/dialog/tooltip/tooltip.directive.ts
```diff
@@ -89,6 +92,6 @@ export class TooltipDirective extends DialogDirective {
 		this.dialogConfig.compID = "tooltip-" + TooltipDirective.tooltipCounter;
 		this.dialogConfig.content = this.ibmTooltip;
 		this.dialogConfig.type = this.tooltipType !== undefined ? this.tooltipType : this.type;
-		this.elementRef.nativeElement.setAttribute("aria-describedby", this.dialogConfig.compID);
+		this.descriptorID = this.dialogConfig.compID;
```
zvonimirfras:
```suggestion
		this.descriptorId = this.dialogConfig.compID;
```
to go along with the name change.

zvonimirfras:
This is a really slick solution.

PR #196, opened 2018-10-22
fix(dropdown): add value input for ngModel

### LINKED ISSUE #188: Add value binding to dropdowns/selects/comboboxes
## Detailed description

> Currently selects/dropdowns/comboboxes only provide `ListItem` values. In almost all cases, our app would rather bind these controls to arbitrary objects, and specify both a `label` and a `value` property

Feature request

Selects/dropdowns/comboboxes/multi-select

I would expect to be able to specify the value I would like the dropdown to provide, rather than needing to extend listitem with the property I want, then parse any value returned for this property.

Any

1.8.x

IBM Clinical Development

## Additional information
Almost all other component libraries support this pattern.
https://material.angular.io/components/select/overview
https://www.primefaces.org/primeng/#/multiselect

cal-smith:
closes #188 

example:

```html
<ibm-dropdown
	placeholder="Select"
	[disabled]="disabled"
	[(ngModel)]="model"
	value="content">
	<ibm-dropdown-list [items]="items"></ibm-dropdown-list>
</ibm-dropdown>
<span>{{model | json}}</span>
```

Where in this case the `value` pulls the `content` field from the default `ListItem`. If you supplied `value="id"` and `ListItem`s that looked like `{content: "foo", id: "123"}` the model would be filled with the `id` property.

ListItem has also been updated to allow arbitrary additional properties to be supplied.


cal-smith:
well that was just super weird ... fixed!

PR #197, opened 2018-10-23
feat(switch): Add small version of swich component

esuau:
Closes IBM/carbon-components-angular#68

The need was to implement a small version of the already existing switch component.
I used the `size` to apply the corresponding styles as described in the [official documentation](http://www.carbondesignsystem.com/components/toggle/code) of Carbon Design System.

#### Changelog

**Changed**

* Implement the small version of switch component using `size` input
* Added unit tests


src/switch/switch.component.ts
```diff
@@ -134,5 +141,9 @@ export class SwitchComponent extends CheckboxComponent implements OnInit {
 		this.renderer.addClass(labelEl, labelClass);
 		this.renderer.addClass(buttonEl, buttonClass);
 		*/
+		if (this.size === "sm") {
+			const inputEl = this.elementRef.nativeElement.querySelector("input");
+			this.renderer.addClass(inputEl, "bx--toggle--small");
+		}
```
cal-smith: Could you instead remove the entire `ngOnInit` block and use an `[ngClass]` on the input?

src/switch/switch.stories.ts
```diff
@@ -10,7 +10,10 @@ storiesOf("Switch", module).addDecorator(
 )
 	.addDecorator(withKnobs)
 	.add("Basic", () => ({
-		template: `<ibm-switch [disabled]="disabled"></ibm-switch>`,
+		template: `
+		<ibm-switch [disabled]="disabled"></ibm-switch>
+		<ibm-switch [size]="'sm'"></ibm-switch>
```
cal-smith: Could you add the `[disabled]="disabled"` attribute as well? that way both will respect the "disabled" knob

src/switch/switch.component.ts
esuau: Sure. :)

src/switch/switch.stories.ts
esuau: Done.

src/switch/switch.component.ts
```diff
@@ -120,19 +121,5 @@ export class SwitchComponent extends CheckboxComponent implements OnInit {
...
-		// Get elements
-		const labelEl = this.elementRef.nativeElement.querySelector("label");
-		const buttonEl = this.elementRef.nativeElement.querySelector("button");
-
-		// Add classes to elements
-		this.renderer.addClass(labelEl, labelClass);
-		this.renderer.addClass(buttonEl, buttonClass);
-		*/
-	}
+	ngOnInit() { }
```
zvonimirfras: If we're removing this it makes sense we remove it completely. `implements` part too.
```suggestion
```

src/switch/switch.stories.ts
```diff
@@ -10,7 +10,10 @@ storiesOf("Switch", module).addDecorator(
 )
 	.addDecorator(withKnobs)
 	.add("Basic", () => ({
-		template: `<ibm-switch [disabled]="disabled"></ibm-switch>`,
+		template: `
+		<ibm-switch [disabled]="disabled"></ibm-switch>
```
zvonimirfras: Indent template code by a tab.
```suggestion
			<ibm-switch [disabled]="disabled"></ibm-switch>
```

src/switch/switch.stories.ts
```diff
@@ -10,7 +10,10 @@ storiesOf("Switch", module).addDecorator(
...
+		<ibm-switch [disabled]="disabled" [size]="'sm'"></ibm-switch>
```
zvonimirfras: This can be further simplified as:
```suggestion
			<ibm-switch [disabled]="disabled" size="sm"></ibm-switch>
```

src/switch/switch.component.ts
```diff
@@ -60,17 +52,27 @@ export class SwitchChange {
 	selector: "ibm-switch",
 	template: `
 		<input
-			class="bx--toggle"
+			[ngClass]="{
+				'bx--toggle': true,
```
zvonimirfras: We normally leave class names that are always true in `class`, so that would bring back the deleted line and remove this one.

PR #198, opened 2018-10-23
Add inline select component

### LINKED ISSUE #152: Add Inline Select component
HTML and style: http://carbondesignsystem.com/components/select/code
Usage: http://carbondesignsystem.com/components/select/usage
Styling: http://carbondesignsystem.com/components/select/style

youda97:
closes #152


src/select/select.component.ts
```diff
@@ -4,7 +4,9 @@ import { Component, Input } from "@angular/core";
 	selector: "ibm-select",
 	template: `
 		<div class="bx--form-item">
-			<div class="bx--select">
+			<div
+				[ngClass]="{ 'bx--select--inline': inlineSelect }"
```
zvonimirfras: All other places in the codebase do
```suggestion
				[ngClass]="{'bx--select--inline': inlineSelect}"
```

src/select/select.component.ts
```diff
@@ -18,6 +20,7 @@ import { Component, Input } from "@angular/core";
 })
 export class Select {
 	static selectCount = 0;
+	@Input() inlineSelect = false;
```
cal-smith: I'd go with `inline = true` ... we know it's a select 😉

cal-smith: Or even `type = "inline"` ... but I'm not so sure about that one

zvonimirfras: Will we have other types of select?

cal-smith: Probably not ... dropdown uses `type` for multi/single ... but even that is kinda weird (the actual `select` element uses a boolean `multiple` for multi/single)

src/select/select.stories.ts
```diff
@@ -25,4 +25,20 @@ storiesOf("Select", module).addDecorator(
           </optgroup>
 		</ibm-select>
 	`
+	}))
+	.add("Inline", () => ({
+		template: `
+		<ibm-select inline="true">
```
panpan-lin: if you use `inline` here, the HTML Dom element for the Select component itself will have an extra attribute `inline` there. If you use `[inline]` instead, then the style will be applied at the right place without having the additional attribute added to the Dom note.

src/select/select.component.ts
zvonimirfras: @cal-smith Based on our Friday discussion, would this be
```suggestion
	@Input() display: "inline";
```
or something along those lines?

cal-smith:
Also there's some merge conflicts ...

youda97:
@cal-smith Why not just fo `@Input() display: "inline";`. Since its inline or its not. We never compare to `default`

cal-smith:
Yeah we could do that ... Personally I prefer having an explicit default state (so there's no questions) but that works too

zvonimirfras:
So are we agreeing that `"default"` is always an option if there's only one extra choice?

So here, we'd have `"default"` for explicit reasons, but for i.e. `"inline" | "single" | "multi"` we don't have a 4th `"default"` because (let's say) `"single"` is the default?

Do a 👍 or 👎

cal-smith:
Yep I can get behind that, keeps things obvious

PR #207, opened 2018-10-25
fix(deploy): deploy versioned documentation

### LINKED ISSUE #199: Versioned Documentation

cal-smith:
should fix #199

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -17,5 +17,8 @@ if [[ $TRAVIS_BRANCH == "master" ]]; then
 	mkdir pages
 	cp -R dist/docs/documentation/ pages/documentation
 	cp -R dist/docs/storybook/* pages
+	version=$(node -e 'const package = require("./package.json"); console.log(package.version);')
+	cp -R dist/docs/documentation/ pages/$version/documentation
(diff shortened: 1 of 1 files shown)


scripts/deploy.sh
```diff
@@ -17,5 +17,8 @@ if [[ $TRAVIS_BRANCH == "master" ]]; then
 	mkdir pages
 	cp -R dist/docs/documentation/ pages/documentation
 	cp -R dist/docs/storybook/* pages
+	version=$(node -e 'const package = require("./package.json"); console.log(package.version);')
```
zvonimirfras: Isn't `package.json` version kinda stuck at 0.1.0?

cal-smith: nope, since this runs after semantic release `package.json` will have the correct value.

scripts/deploy.sh
```diff
@@ -17,5 +17,8 @@ if [[ $TRAVIS_BRANCH == "master" ]]; then
...
+	cp -R dist/docs/documentation/ pages/$version/documentation
+	cp -R dist/docs/storybook/* pages/$version
```
zvonimirfras: I like these paths :D

PR #210, opened 2018-10-25
fix(dialog): fix dialog in IE11

### LINKED ISSUE #159: Overflow menu positioning in IE 11
Hi 

I am using overflow menu, it cannot work correctly in IE11, and when I run storybook(https://angular.carbondesignsystem.com) in IE11, I see a blank page. Does carbon-components-angular support IE?

cal-smith:
closes #159


.storybook/addons.js
```diff
@@ -1,3 +1,21 @@
+import "core-js/es6/symbol";
```
asudoh: Weighing in a bit - Carbon vanilla/React variants maintain separate lists of "polyfills modules" so consumers know what polyfill they need to run the library on IE11. Great to have a similar thing, if the topic has not been documented.

src/dialog/dialog.component.ts
```diff
@@ -137,7 +137,11 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
 	 */
 	ngAfterViewInit() {
 		const dialogElement = this.dialog.nativeElement;
-		dialogElement.classList = `${dialogElement.classList} ${this.dialogConfig.wrapperClass}`;
+		if (this.dialogConfig.wrapperClass) {
```
zvonimirfras: Maybe add a comment to blame IE11 for the complication?

src/dialog/dialog.directive.ts
```diff
@@ -61,7 +61,7 @@ export class DialogDirective implements OnInit, OnDestroy, OnChanges {
 	 * @type {string}
 	 * @memberof DialogDirective
 	 */
-	@Input() wrapperClass: string;
+	@Input() wrapperClass = "";
```
zvonimirfras: It can't be undefined or empty string signifies something?

.storybook/addons.js
```diff
@@ -1,3 +1,21 @@
...
+import "core-js/es6/array";
+import "core-js/es6/regexp";
+import "core-js/es6/map";
+import "core-js/es6/set";
+import "core-js/es6/reflect";
+
+import "core-js/es7/reflect";
+import "zone.js/dist/zone";
+
 import '@storybook/addon-actions/register';
```
zvonimirfras: Single quotes 👀

.storybook/addons.js
cal-smith: I'll have to double check what Angular CLI ships/recommends ... this should mirror that, or at least we should be following what they recommend.

It's probably a good idea to review our actual polyfill dependency list though, the less we have to rely on the better.

src/dialog/dialog.directive.ts
cal-smith: It could probably go back to undefined ... I tend to prefer initializing to useful (non undefined) values where possible though.

.storybook/addons.js
cal-smith: that's what happens when the linter doesn't lint all the things 😂

src/dialog/dialog.directive.ts
zvonimirfras: That implies that a value is needed when it's not.

cal-smith: Not really? It just implies that it's set to an empty string

src/dialog/dialog.component.ts
zvonimirfras: Is github not showing the comment or you think it's no longer necessary?

cal-smith: I actually totally forgot to add a comment 😅

src/dialog/dialog.component.ts
```diff
@@ -137,7 +137,14 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
 	 */
 	ngAfterViewInit() {
 		const dialogElement = this.dialog.nativeElement;
-		dialogElement.classList = `${dialogElement.classList} ${this.dialogConfig.wrapperClass}`;
+		// split the wrapper class list and apply seperatly to avoid IE from
+		// 1. throwing an error due to asigning a readonly property (classList)
```
zvonimirfras:
```suggestion
		// 1. throwing an error due to assigning a readonly property (classList)
```

src/dialog/dialog.component.ts
```diff
@@ -137,7 +137,14 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
 	 */
 	ngAfterViewInit() {
 		const dialogElement = this.dialog.nativeElement;
-		dialogElement.classList = `${dialogElement.classList} ${this.dialogConfig.wrapperClass}`;
+		// split the wrapper class list and apply seperatly to avoid IE from
```
zvonimirfras:
```suggestion
		// split the wrapper class list and apply separately to avoid IE from
```

src/dialog/dialog.component.ts
```diff
@@ -137,7 +137,14 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
...
+		// 1. throwing an error due to asigning a readonly property (classList)
+		// 2. throwing an SyntaxError due to passing an empty string to .add
```
zvonimirfras:
```suggestion
		// 2. throwing a SyntaxError due to passing an empty string to `add`
```

src/dialog/tooltip/tooltip.stories.ts
```diff
@@ -35,10 +35,12 @@ storiesOf("Tooltip", module)
 			</ng-template>
 			<span
 				[ibmTooltip]="template"
+				appendToBody="true"
```
zvonimirfras: No longer needed 🎉 
```suggestion
```

src/dialog/tooltip/tooltip.stories.ts
```diff
@@ -35,10 +35,12 @@ storiesOf("Tooltip", module)
...
 				trigger="click"
 				placement="bottom"
 				style="cursor: pointer;">
 				Click for tooltip
 			</span>
+			<ibm-dialog-placeholder></ibm-dialog-placeholder>
```
zvonimirfras: Deprecated 🙌 
```suggestion
			<ibm-placeholder></ibm-placeholder>
```

Issue #163, opened 2018-10-09
Add a light style to the Carbon components that support it


stanislavgeorgiev:
## Detailed description

> Describe in detail the issue you're having.

Provide an option to choose between the default style and the light style for the components that are supporting it in Carbon:

- [ ] Dropdown
- [x] Input
- [ ] Select
- [x] Textarea

That new setting should have the same name for all components. For example, `style: "" | "light" = "";` or `light = false`.

> Is this a feature request (new component, new icon), a bug, or a general issue?

Yes

> Is this issue related to a specific component?

Select

zvonimirfras:
To whoever picks this up, we decided that `@Input` for controlling this should be named `theme`.

stanislavgeorgiev:
https://github.com/IBM/carbon-components-angular/pull/215 is fixing the Input and Textarea

PR #212, opened 2018-10-26
Add styles for Select component

panpan-lin:

add styles for Select component, for issue #163 

#### Changelog

**Changed**

* `light` and `inline` styles as per Carbon design:
http://www.carbondesignsystem.com/components/select/code

Inline:
screen shot 2018-10-26 at 17 31 26

Light:
screen shot 2018-10-26 at 17 31 14

### DIFF
--- a/src/select/select.stories.ts
+++ b/src/select/select.stories.ts
@@ -44,7 +44,7 @@ storiesOf("Select", module).addDecorator(
 	}))
 	.add("Light", () => ({
 		template: `
-		<ibm-select [theme]="'light'">
+		<ibm-select theme="light">
 			<option value="" disabled selected hidden>Choose an option</option>
           <option value="solong">A much longer option that is worth having around to check how text flows</option>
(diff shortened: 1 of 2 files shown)


youda97:
Only light styles is necessary as there is already an existing PR for inline styles

panpan-lin:
@youda97 OK. Thanks for the update. I'll update this PR once the other one is in then, instead of changing now and then having to resolve conflict later.

panpan-lin:
@cal-smith ready for another round of review :-)

src/select/select.stories.ts
```diff
@@ -42,6 +42,22 @@ storiesOf("Select", module).addDecorator(
 		</ibm-select>
 	`
 	}))
+	.add("Light", () => ({
+		template: `
+		<ibm-select [theme]="'light'">
```
cal-smith: Just a small change:
```suggestion
		<ibm-select theme="light">
```

PR #213, opened 2018-10-26
 Fix DAP issues for overflow

youda97:


src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -86,10 +85,26 @@ export class OverflowMenuPane extends Dialog {
 				event.preventDefault();
 				listItems[listItems.length - 1].focus();
 				break;
+
+			case "Esc": // IE specific value
+			case "Escape": {
```
zvonimirfras: Lose the `{}`, other cases here don't have it and they are not actually needed.

src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -86,10 +85,26 @@ export class OverflowMenuPane extends Dialog {
...
+				event.stopImmediatePropagation();
+				this.doClose();
+				break;
+			}
+		}
+	}
+
+	clickClose(event) {
+		if (this.dialogConfig.parentRef.nativeElement.firstChild.contains(event.target) ||
+			this.listItems().some(button => button === (event.relatedTarget)) ||
```
zvonimirfras: What are the brackets for?
```suggestion
			this.listItems().some(button => button === event.relatedTarget) ||
```

src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -86,10 +85,26 @@ export class OverflowMenuPane extends Dialog {
...
+			(event.type === "focusout" && event.relatedTarget === this.dialogConfig.parentRef.nativeElement)) {
```
zvonimirfras: What are the brackets for?
```suggestion
			event.type === "focusout" && event.relatedTarget === this.dialogConfig.parentRef.nativeElement) {
```

src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -86,10 +85,26 @@ export class OverflowMenuPane extends Dialog {
...
+			case "Escape": {
+				event.stopImmediatePropagation();
+				this.doClose();
+				break;
+			}
+		}
+	}
+
+	clickClose(event) {
+		if (this.dialogConfig.parentRef.nativeElement.firstChild.contains(event.target) ||
```
zvonimirfras: Add a comment that in simple words explains in which cases to we not actually close.

src/dialog/overflow-menu/overflow-menu.stories.ts
```diff
@@ -18,7 +18,7 @@ storiesOf("Overflow Menu", module)
 				</ibm-overflow-menu-option>
 				<ibm-overflow-menu-option>Option 2</ibm-overflow-menu-option>
 				<li class="bx--overflow-menu-options__option">
-					<button class="bx--overflow-menu-options__btn">A fully custom option</button>
+					<button class="bx--overflow-menu-options__btn" tabindex="-1">A fully custom option</button>
```
zvonimirfras: Do users always have to set this in their options?

src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -46,12 +46,20 @@ export class OverflowMenuPane extends Dialog {
 			this.dialogConfig.menuLabel = this.i18n.get().OVERFLOW_MENU.OVERFLOW;
 		}
 
-		setTimeout(() => this.listItems()[0].focus());
+		setTimeout(() => {
+			let options = Array.from<any>(this.dialog.nativeElement.querySelectorAll("*"));
```
cal-smith: This should probably be 

```suggestion
			let options = Array.from<any>(this.dialog.nativeElement.querySelectorAll(".bx--overflow-menu-options__btn"));
```

Since the options need that class to look like anything useful

src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -46,12 +46,20 @@ export class OverflowMenuPane extends Dialog {
...
+
+			options.forEach(element => {
+				if (element.tabIndex === 0) {
```
cal-smith: Maybe something like this? Per @zvonimirfras's comment
```suggestion
				if (element.tabIndex && element.tabIndex === 0) {
```

src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -46,18 +48,20 @@ export class OverflowMenuPane extends Dialog {
...
-		}
+		};
 
 		if (!this.dialogConfig.menuLabel) {
 			this.dialogConfig.menuLabel = this.i18n.get().OVERFLOW_MENU.OVERFLOW;
 		}
 
-		setTimeout(() => this.listItems()[0].focus());
+		setTimeout(() => {
+			getFocusElementList(this.elementRef.nativeElement).every(button => button.tabIndex = -1);
```
zvonimirfras: It's not checking if `tabIndex` is already set so it will override user set values.
```suggestion
			getFocusElementList(this.elementRef.nativeElement).every(button => {
				if (button.tabIndex === undefined) {
					button.tabIndex = -1;
				}
			});
```
should do the trick but test it.

PR #214, opened 2018-10-26
Add light style for Dropdown component

panpan-lin:

issue #163 points out dropdown should select a light style.
In Carbon Design, if you choose `field-02` and you will see the "light" style in use.
http://www.carbondesignsystem.com/components/dropdown/code
screen shot 2018-10-26 at 18 17 48

#### Changelog

**Changed**

* add @Input light to control if the user wants to apply a light style

screen shot 2018-10-26 at 18 16 32


src/dropdown/dropdown.component.ts
```diff
@@ -96,7 +98,11 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy {
 	 * @memberof Dropdown
 	 */
 	@Input() type: "single" | "multi" = "single";
-
+	/**
+	 * Set to `true` to apply the light style to the dropdown.
+	 * @memberof Dropdown
+	 */
+	@Input() light = false;
```
zvonimirfras: We decided to standardize this and go instead with 
```suggestion
	@Input() theme: "light" | "dark" = "dark";
```

This will become a pattern we use across the library. Theme `enum` will be the type.

src/dropdown/dropdown.component.ts
```diff
@@ -34,7 +34,9 @@ import { I18n } from "./../i18n/i18n.module";
 @Component({
 	selector: "ibm-dropdown",
 	template: `
-	<div class="bx--list-box">
+	<div
+		class="bx--list-box"
+		[ngClass]="{'bx--dropdown--light': light}">
```
zvonimirfras: With the change this would become
```suggestion
		[ngClass]="{'bx--dropdown--light': theme === 'light'}">
```

src/dropdown/dropdown.stories.ts
```diff
@@ -40,6 +40,29 @@ storiesOf("Dropdown", module)
 			onClose: action("Dropdown closed")
 		}
 	}))
+	.add("Light", () => ({
+		template: `
+		<ibm-dropdown
+			[light]="true"
```
zvonimirfras:
```suggestion
			theme="light"
```

panpan-lin:
@zvonimirfras This is ready for another round of review.

src/dropdown/dropdown.stories.ts
```diff
@@ -41,6 +41,29 @@ storiesOf("Dropdown", module)
 			onClose: action("Dropdown closed")
 		}
 	}))
+	.add("Light", () => ({
+		template: `
+		<ibm-dropdown
+			[theme]="'light'"
```
cal-smith: This can just be:

```suggestion
			theme="light"
```

panpan-lin: But if you do it `theme="light"`, the theme will be an extra attribute appearing in the HTML Dom node. I use `[theme]="'light'"` because I think it is cleaner. However, if you guys prefer to have that as an attribute in the dom I will gladly apply your suggested changesets :-)

cal-smith: Since it's just a static attr in this case I think it makes more sense to bind it as an attr ... even though it clutters the DOM a little bit ...

panpan-lin: Sure, just applied your change :-)

src/dropdown/dropdown.component.ts
```diff
@@ -28,7 +28,9 @@ import { I18n } from "./../i18n/i18n.module";
 @Component({
 	selector: "ibm-dropdown",
 	template: `
-	<div class="bx--list-box">
+	<div
+		class="bx--list-box"
+		[ngClass]="{'bx--dropdown--light': theme === 'light'}">
```
youda97: `@hostbinding` should be used instead to set the classes

zvonimirfras: This class is not set on the host so it can't be set using host binding.

zvonimirfras:
There's something happening with the tests.

cal-smith:
it's the same `detachFromAppRef` thing ... I'm taking a look at it, but it shouldn't block this PR

PR #221, opened 2018-10-30
Add toggle component and deprecate switch

### LINKED ISSUE #129: Naming inconsistency: Toggle (Carbon) v. Switch (Carbon Angular)
For synchronicity with Carbon, can Switch be changed to Toggle? It will help our designers and developers stay on the same page as designers work with Carbon design assets and developers work with Carbon Angular components.

youda97:
closes #129


src/switch/switch.stories.ts
```diff
@@ -3,6 +3,11 @@ import { withKnobs, boolean } from "@storybook/addon-knobs/angular";
 
 import { SwitchModule } from "../";
 
+/**
+ * Deprecated in favour of `Toggle` (to be removed in v3.0).
+ *
+ * @deprecated
+ */
 storiesOf("Switch", module).addDecorator(
```
cal-smith: We should remove the switch stories to reduce confusion

zvonimirfras: Yes.

src/toggle/toggle.component.ts
```diff
@@ -0,0 +1,118 @@
...
+	 * Variable used for creating unique ids for toggle components.
+	 * @type {number}
+	 * @static
+	 * @memberof ToggleComponent
+	 */
+	static toggleCount = 0;
+
+	/**
+	 * Size of the toggle component.
+	 * (size `"default"` is being deprecated as of neutrino v1.2.0, please use `"md"` instead)
```
zvonimirfras: Let's remove "default"

PR #233, opened 2018-11-01
 Remove mention of neutrino and replace it with carbon-angular

magedhennawy:


gulpfile.js
```diff
@@ -29,7 +29,7 @@ const dirs = {
 
 const licenseTemplate = `/*!
  *
- * Neutrino v@PACKAGE_VERSION@ | @FILE_NAME@
+ * Carbon-Angular v@PACKAGE_VERSION@ | @FILE_NAME@
```
cal-smith: Just to make sure the incense header and package name line up ...

```suggestion
 * carbon-components-angular v@PACKAGE_VERSION@ | @FILE_NAME@
```

zvonimirfras: Actually, why aren't we using full name everywhere to minimize confusion?

cal-smith: For the stuff in webpack I think `carbon-angular` is fine - `carbon-components-angular` is kinda long (since it's going to be in the filenames and how people access it. Everywhere else though yeah, it could be the full name

webpack.build.js
```diff
@@ -22,12 +22,12 @@ function rxjsExternal(context, request, cb) {
 module.exports = [{
 	devtool: "source-map",
 	entry: {
-		neutrino: "./src/index.ts",
+		"carbon-angular": "./src/index.ts",
```
cal-smith: This should actually be 

```suggestion
		carbonAngular: "./src/index.ts",
```

Because I think it's used to generate some names people can use ... same with the library definition below

webpack.build.js
```diff
@@ -22,12 +22,12 @@ function rxjsExternal(context, request, cb) {
...
 	},
 	output: {
 		path: __dirname + "/dist/bundle",
-		filename: "neutrino.umd.js",
-		library: "neutrino",
+		filename: "carbon-angular.umd.js",
+		library: "carbon-angular",
```
cal-smith: The filename is fine, but the library name is what gets set to the window if it's loaded directly.

```suggestion
		library: "carbonAngular",
```

PR #235, opened 2018-11-02
fix(dialog): add singleton click listener for safari fix

### LINKED ISSUE #224: Improve tooltip performance
## Detailed description

Tooltip in a table column with 50+ rows was causing performance issue. Tooltip was taking close to 5+ seconds to render.

From further investigation 
```document.body.firstElementChild.addEventListener("click", () => null, true);``` in `Dialog Directive` was causing the issue. 

Performance / Scalability issue. 

Dialog 

Chrome

"carbon-components": "9.23.0",
"carbon-components-angular": "1.9.2",

### PR DESCRIPTION
**Pull request by zvonimirfras**
Improves performance in extreme cases.

closes #224

### DIFF
--- a/src/dialog/dialog.service.ts
+++ b/src/dialog/dialog.service.ts
@@ -21,7 +21,7 @@ export class DialogService {
 	/**
 	 * Used in `singletonClickListen`, don't count on its existence and values.
 	 */
-	static listeningForBodyClicks = false;
+	private static listeningForBodyClicks = false;
 
 	/**
 	 * Reflects the open or closed state of the `Dialog`.
@@ -174,7 +174,9 @@ export class DialogService {
 	 * and changes to it won't be considered a breaking change. Use at your own risk.
 	 */
 	singletonClickListen() {
+		console.log("singleton");
 		if (!DialogService.listeningForBodyClicks) {
+			console.log("singleton click");
 			document.body.firstElementChild.addEventListener("click", () => null, true);
 			DialogService.listeningForBodyClicks = true;
 		}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/dialog/dialog.service.ts [hunk 1]
```diff
@@ -18,6 +18,11 @@ import { PlaceholderService } from "./../placeholder/placeholder.module";
  */
 @Injectable()
 export class DialogService {
+	/**
+	 * Used in `singletonClickListen`, don't count on its existence and values.
+	 */
+	static listeningForBodyClicks = false;
```
cal-smith: Does `private static` or `static private` work in addition to the warning?

src/dialog/dialog.service.ts
```diff
@@ -157,4 +162,21 @@ export class DialogService {
...
+	 * Fix for safari hijacking clicks.
+	 *
+	 * Runs on `ngOnInit` of every dialog. Ensures we don't have multiple listeners
+	 * because having many of them could degrade performance in certain cases (and is
+	 * not necessary for our use case)
+	 *
+	 * This is an internally used function, can change at any point (even get removed)
+	 * and changes to it won't be considered a breaking change. Use at your own risk.
+	 */
+	singletonClickListen() {
```
cal-smith: `singletonClickListen` works ... I can't think of anything else right now that makes more sense 👍

src/dialog/dialog.service.ts
zvonimirfras: Yep, I didn't think that was a thing. Today I learned...

PR #236, opened 2018-11-03
fix(tabs): Redefine tab selection area

esuau:
Closes IBM/carbon-components-angular#201

Tab selection was only triggered by clicking on the tab title, not the entire heading area.

#### Changelog

**Changed**

* Move `(click)="selectTab(tabref, tab, i)"` from `<a>` to `<li>` tag.


src/tabs/tab-headers.component.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+		});
+
+		fixture = TestBed.createComponent(TabHeaders);
+		component = fixture.componentInstance;
+	});
+
+	it("should work", () => {
+		expect(component instanceof TabHeaders).toBe(true);
+	});
+
```
zvonimirfras:
```suggestion
```

PR #238, opened 2018-11-05
fix(notification): Allow falsy `notificationObj`

### LINKED ISSUE #237: Notification shouldn't throw errors when `notificationObj` is set to `null` or not set.

zvonimirfras:
closes #237

### DIFF
--- a/src/notification/notification.component.ts
+++ b/src/notification/notification.component.ts
@@ -5,7 +5,6 @@ import {
 	EventEmitter,
 	ComponentRef,
 	ViewChild,
-	OnInit,
 	HostBinding
 } from "@angular/core";
 
@@ -48,7 +47,7 @@ import { I18n } from "./../i18n/i18n.module";
 	`,
 	providers: [NotificationService]
 })
-export class Notification implements OnInit {
+export class Notification {
 	/**
 	 * Can have `type`, `title`, and `message` members.
 	 *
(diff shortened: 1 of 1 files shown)


zvonimirfras:
Because subclassed internal components use it. Also, defensive programming.

cal-smith:
So the subclasses are subclassing wrong ... Dialog has a ton of subclasses and we haven't run into issues like this.

zvonimirfras:
It's not the subclass itself but the demo using the subclass.
Consider a static notification banner with content loading from backend. It will start as undefined and populate when data becomes available.

It's also [defensive programming](https://en.wikipedia.org/wiki/Defensive_programming). The component should just fall apart and break the page if an input is not provided. Especially if there's a reasonable way around it.

cal-smith:
I understand what defensive programming is thanks.

So really the issue at hand is that, for some cases, someone may want to pass a non-useful object to a banner?

Three solutions:

- Provide sensible defaults in the component. Because yes, the component _should_ just fall over, but it also shouldn't need a million checks and `ngIf`ing out the entire template.
- Throw an error if non-usefil objects are passed. Just because something is loading, doesn't mean it's a good candidate to pass nonsensical inputs to. It's better to fail fast than to hide errors.
- Add a skeleton state. We don't have this yet, and I haven't given it a ton of thought as to what the ideal implementation is, but carbon does provide skeleton states for most components.
  - In this case I don't believe there is one for notifications, but that's something we could add.

_Any_ of those three solutions in isolation is better, and better for our users. Together they would be a pretty great enhancement.

zvonimirfras:
After internal discussion, we decided to deprecate problematic internal behavior and come up with a more elegant solution.

PR #239, opened 2018-11-05
feat(inline-loading): Add inline loading component

### LINKED ISSUE #71: Add Inline loading component/helper
HTML and styles: http://www.carbondesignsystem.com/components/inline-loading/code
Usage: http://www.carbondesignsystem.com/components/inline-loading/usage

youda97:
closes #71


src/inline-loading/inline-loading.component.ts
```diff
@@ -0,0 +1,87 @@
...
+	/**
+	 * Specify the text description for the success state.
+	 *
+	 * @memberof InlineLoading
+	 */
+	@Input() successText;
+
+	/**
+	 * Returns value `true` if the component is in the success state.
+	 * @readonly
```
cal-smith:
```suggestion
```

Since it's not readonly

src/inline-loading/inline-loading.component.ts
```diff
@@ -0,0 +1,87 @@
...
+	 */
+	@Input() get success() {
+		return this._success;
+	}
+	/**
+	 * Set the component's state to match the parameter and emits onSuccess if it exits.
+	 */
+	set success (success: boolean) {
+		this._success = success;
+		if (this._success) {
```
cal-smith: I see what the goal is, but this block will only emit true _after_ the value is set to true. The carbon usage docs say it should be set to true and emitted after 1.5s only if no callback is specified.

That said I don't think that pattern is useful in the angular world, so I'd just go with a straight `Input` and drop the EventEmitter and related code

zvonimirfras: Well spotted!
My understanding is that the pattern drives user experience and is not there because of framework limitation.

If we're keeping this, the `if` statement should check against `success` and `this._success` should only be set in the `setTimeout` after delay has elapsed.

src/inline-loading/inline-loading.component.ts
```diff
@@ -0,0 +1,87 @@
...
+			setTimeout(() => {
+				if (this.onSuccess.observers.length > 0) {
```
cal-smith: So we can drop this if block, cause we can just emit anyway

src/inline-loading/inline-loading.component.ts
```diff
@@ -0,0 +1,87 @@
...
+	@Output() onSuccess: EventEmitter<any> = new EventEmitter();
+
+	@HostBinding("class.bx--inline-loading") loadingClass = true;
+
+	/**
+	 * Provide a delay for the `setTimeout` for success.
+	 *
+	 * @memberof InlineLoading
+	 */
+	successDelay = 1500;
```
cal-smith: This should be an Input

PR #244, opened 2018-11-06
fix(dialog): Remove console.log("singleton") statements.

stanislavgeorgiev:


stanislavgeorgiev:
OK, I obviously didn't run my own lint check but I don't know why Git allowed me to push the changes at all. I see a couple of inconsistency problems with deprecation messages - some components are using `console.log` and others `console.warn`. Perhaps the `log()` calls need to be converted to `warn()` and then allow only `warn()` and `error()`?

cal-smith:
Yep that sounds good

stanislavgeorgiev:
That "Deploy preview for c4ng ready!" you've added is fantastic!!!

PR #260, opened 2018-11-10
fix(table): Keep header when data is emptied

esuau:
Closes IBM/carbon-components-angular#240

The table header was lost when data was flushed. The goal was to preserve it when only the data was emptied.

#### Changelog

**Changed**

* Changed condition of `set data(…)` in `TableModel` to keep the header when data is set to empty.


esuau:
It was actually an issue caused by the use of `destroy()`  in the test case `should emit change when notification is closed programmatically` of `Notification` component. I just ignored it, but I will try to rewrite it properly.

cal-smith:
Thanks for looking into it! I'll leave it up to you if you want to tackle it or not 👍

esuau:
I'll certainly submit it in a next PR.

esuau:
Fixed.

PR #266, opened 2018-11-12
Fix(overflow menu): Click doesn't work in firefox and IE

### LINKED ISSUE #259: ibm-overflow-menu click doesn't work in firefox or IE
## Detailed description

In firefox or IE, when you click on the kebab or meatball icon, the list of options open up but when you click on it the onclick event doesn't fire. Rather it closes the overflow menu and that's it

Bug

ibm-overflow-menu / Dialog

onClick of the option, I expect  it to trigger the function. 

Firefox and IE, works fine on chrome

2.0.6

youda97:
closes #259

### DIFF
--- a/src/dialog/overflow-menu/overflow-menu-pane.component.ts
+++ b/src/dialog/overflow-menu/overflow-menu-pane.component.ts
@@ -19,7 +19,8 @@ import { I18n } from "./../../i18n/i18n.module";
 			#dialog
 			class="bx--overflow-menu-options bx--overflow-menu-options--open"
 			role="menu"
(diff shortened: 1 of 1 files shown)


johnharan:
@youda97 @cal-smith 

Hi guys, I'm still seeing the following error when clicking an option in IE:
"ERROR TypeError: Object doesn't support property or method 'focus'"

However, if I tab to the overflow menu, open it with enter key, and press enter key on an option, the above error does not show.

I'm reproducing this in the storybook as well as in our team's implementation (where it results in the page navigation/loading breaking).

Interestingly I never experienced the issue in firefox, even before this update.

PR #267, opened 2018-11-12
feat(search): Add search component

esuau:
Closes IBM/carbon-components-angular#72

A search component was needed.

#### Changelog

**New**

* Added search component (`ibm-search`).


src/search/search.component.ts
```diff
@@ -0,0 +1,236 @@
...
+	onChange(event) {
+		event.stopPropagation();
+	}
+
+	/**
+	 * Handles click events on the `Search` and emits changes to other classes.
+	 * @param {any} event
+	 * @memberof Search
+	 */
+	onClick(event) {}
```
cal-smith: If this isn't being used we can just drop it

src/search/search.component.ts
```diff
@@ -0,0 +1,236 @@
...
+	public registerOnTouched(fn: any) {
+		this.onTouched = fn;
+	}
+
+	/**
+	 * Executes on the event of a change within `Search` to block propagation.
+	 * @param {any} event
+	 * @memberof Search
+	 */
+	onChange(event) {
```
cal-smith: I'm curious if this is something we need? It's not immediately clear to me why we would need to stop propagation for `change`.

src/search/search.component.ts
```diff
@@ -0,0 +1,236 @@
...
+	 * @memberof Search
+	 */
+	clearSearch(): void {
+		this.value = "";
+	}
+
+	/**
+	 * Creates a class of `RadioChange` to emit the change in the `RadioGroup`.
+	 */
+	emitChangeEvent() {
```
cal-smith: this should probably call `propagateChange` with `this.value`

src/search/search.component.ts
```diff
@@ -0,0 +1,236 @@
...
+		this.value = search;
+		this.emitChangeEvent();
+	}
+
+	/**
+	 * Called when clear button is clicked.
+	 * @memberof Search
+	 */
+	clearSearch(): void {
+		this.value = "";
```
cal-smith: this should also call `propagateChange` with an empty string

src/search/search.stories.ts
```diff
@@ -0,0 +1,25 @@
...
+storiesOf("Search", module).addDecorator(
+	moduleMetadata({
+		imports: [SearchModule]
+	})
+)
+	.addDecorator(withKnobs)
+	.add("Basic", () => ({
+		template: `
+		<div style="width: 250px;">
+			<ibm-search [size]="'lg'"></ibm-search>
```
cal-smith: `size` can just be set directly

```suggestion
			<ibm-search size="lg"></ibm-search>
```

src/search/search.stories.ts
```diff
@@ -0,0 +1,25 @@
...
+		</div>
+		`
+	}))
+	.add("Small", () => ({
+		template: `
+		<div style="width: 250px;">
+			<ibm-search [size]="'sm'"></ibm-search>
```
cal-smith: same here

```suggestion
			<ibm-search size="sm"></ibm-search>
```

src/search/search.component.ts
esuau: Looking at other components (checkbox), I think the idea was to stop the native event propagation and to override it by emitting our own `[ComponentName]Change` object. However, I don't get the point of this behaviour and I agree to stick to the standard.

src/search/search.component.ts
```diff
@@ -0,0 +1,220 @@
...
+	@Input() value = "";
+	/**
+	 * Sets the placeholder attribute on the  `input` element.
+	 */
+	@Input() placeholder = "Search";
+	/**
+	 * Used to set the `aria-label` attribute on the input element.
+	 */
+	// tslint:disable-next-line:no-input-rename
+	@Input("aria-label") ariaLabel = "Clear search input";
```
zvonimirfras: This (and all other hard-coded default English strings) should be set from a field in `src/i18n/en.json` using `I18n` service.
See `src/loading/loading.component.ts` `title` for example.

src/search/search.component.ts
```diff
@@ -0,0 +1,220 @@
...
+		<div class="bx--form-item">
+			<div
+				class="bx--search"
+				[ngClass]="{
+					'bx--search--sm': size === 'sm',
+					'bx--search--lg': size === 'lg',
+					'bx--search--light': theme === 'light'
+				}"
+				role="search">
+				<label class="bx--label" [for]="id">Search</label>
```
zvonimirfras: `Search` should be translatable.

src/search/search.component.ts
```diff
@@ -0,0 +1,220 @@
...
+				<input
+					class="bx--search-input"
+					type="text"
+					role="search"
+					[id]="id"
+					[value]="value"
+					[placeholder]="placeholder"
+					[disabled]="disabled"
+					[required]="required"
+					[attr.aria-labelledby]="ariaLabelledby"
```
zvonimirfras: Isn't this input labeled by the "Search" label?
If so, label should get an id and we don't need `ariaLabelledby`, it doesn't have to be an input. If provided, it clashes with label's `for` anyway.

src/search/search.component.ts
```diff
@@ -0,0 +1,220 @@
...
+					<path
+						d="M6.5 12a5.5 5.5 0 1 0 0-11 5.5 5.5 0 0 0 0 11zm4.936-1.27l4.563 4.557-.707.708-4.563-4.558a6.5 6.5 0 1 1 .707-.707z"
+						fill-rule="nonzero"/>
+				</svg>
+				<button
+					class="bx--search-close"
+					[ngClass]="{
+						'bx--search-close--hidden': !value || value === 0
+					}"
+					title="Clear search input"
```
zvonimirfras: `title` should be translatable.
For multiple translation strings example see how `src/table/table.component.ts` `translations` input does things.

src/search/search.component.ts
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * @export
+ * @class Search
+ * @implements {ControlValueAccessor}
+ */
+@Component({
+	selector: "ibm-search",
+	template: `
+		<div class="bx--form-item">
```
zvonimirfras: It looks like this div is just a container div. It's usually possible to remove it and let component tag be the container. In that case class is assigned with a `HostBinding`.

See `src/accordion/accordion-item.component.ts` as an example.

src/search/search.module.ts
```diff
@@ -0,0 +1,23 @@
...
+@NgModule({
+	declarations: [
+		Search
+	],
+	exports: [
+		Search
+	],
+	imports: [
+		FormsModule,
+		CommonModule
```
zvonimirfras: You need `I18nModule` here for build to pass.
```suggestion
		CommonModule,
		I18nModule
```

src/search/search.module.ts
```diff
@@ -0,0 +1,23 @@
+// modules
+import { NgModule } from "@angular/core";
+import { FormsModule } from "@angular/forms";
+import { CommonModule } from "@angular/common";
+
+// imports
```
zvonimirfras:
```suggestion
// imports
import { I18nModule } from "../i18n/i18n.module";
```

esuau:
Finally solved all the import issues 😓

src/search/search.component.ts
```diff
@@ -0,0 +1,223 @@
...
+				height="16"
+				viewBox="0 0 16 16">
+				<path
+					d="M6.5 12a5.5 5.5 0 1 0 0-11 5.5 5.5 0 0 0 0 11zm4.936-1.27l4.563 4.557-.707.708-4.563-4.558a6.5 6.5 0 1 1 .707-.707z"
+					fill-rule="nonzero"/>
+			</svg>
+			<button
+				class="bx--search-close"
+				[ngClass]="{
+					'bx--search-close--hidden': !value || value === 0
```
zvonimirfras: `!value` should cover `value === 0` case. Is there a special reason why you need both? If not, do
```suggestion
					'bx--search-close--hidden': !value
```

esuau: I actually meant `value.length === 0`.

Issue #73, opened 2018-09-14
Add Slider component


zvonimirfras:
HTML and styles: http://www.carbondesignsystem.com/components/slider/code
Usage: http://www.carbondesignsystem.com/components/slider/usage

PR #291, opened 2018-11-25
feat(number): Add number input component

esuau:
Closes IBM/carbon-components-angular#125

The library needed a number input component.

#### Changelog

**New**

* Added number input component

**Notes** 

This component might be useful for the development of a slider component (#73)

### DIFF
--- a/src/number-input/number.module.ts
+++ b/src/number-input/number.module.ts
@@ -1,0 +1,25 @@
+// modules
+import { NgModule } from "@angular/core";
+import { FormsModule } from "@angular/forms";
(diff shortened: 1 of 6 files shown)


src/number-input/number.module.ts
```diff
@@ -0,0 +1,25 @@
...
+	declarations: [
+		Number
+	],
+	exports: [
+		Number
+	],
+	imports: [
+		FormsModule,
+		CommonModule,
+		I18nModule
```
cal-smith: We can drop this if we're not providing any translations in this module 👍 

```suggestion
```

PR #302, opened 2018-12-01
fix(i18n): set now does a deep merge

### LINKED ISSUE #301: i18n: Support default language strings
## Detailed description

We're using `i18n.set()` to provide translated strings which works great. The issue is that if we don't have the Carbon string keys in our files then the Carbon components won't show any text.

For example, providing English strings as the following:
```	
	"PAGINATION": {
		"ITEMS_PER_PAGE": "Items per page:",
		"OPEN_LIST_OF_OPTIONS": "Open list of options",
		"BACKWARD": "Backward",
		"FORWARD": "Forward",
		"TOTAL_ITEMS": "{{start}}-{{end}} of {{total}} items",
		"TOTAL_PAGES": "{{current}} of {{last}} pages"
	},
```
would end rendering correctly as:
image

Then switching the language to Chinese which hasn't been fully translated yet (pending PII) would end up looking like this:

image

Feature request

i18n

If a string key hasn't been found then fall back to the default English strings that come along with Carbon. Another option to make it even more flexible is allow users to set the default strings (which initially are set to the Carbon ones)

Any

2.4.0

ADI

> Reproduce

This should show the problem right away as the new keys for TOTAL_ITEMS and TOTAL_PAGES are missing:

```
i18n.set({
	"PAGINATION": {
		"ITEMS_PER_PAGE": "Eléments par page :",
		"BACKWARD": "Vers l'arrière",
		"FORWARD": "Vers l'avant"
	},
});
```

cal-smith:
closes #301 

`Object.assign` only does a shallow merge, which of course means we'd lose any untranslated keys (potentially for whole components)

### DIFF
--- a/src/i18n/i18n.service.ts
+++ b/src/i18n/i18n.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from "@angular/core";
 import { BehaviorSubject } from "rxjs";
 import { map } from "rxjs/operators";
+import { merge } from "./../utils/object";
 
 const EN = require("./en.json");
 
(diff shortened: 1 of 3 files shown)


src/i18n/i18n.service.ts
```diff
@@ -34,8 +34,30 @@ export const replace = (subject, variables) => subject.pipe(
 	})
 );
 
+// custom deep object merge
+const merge = (target, ...objects) => {
```
zvonimirfras: Maybe add this to `src/utils/utils.ts` or `src/common/utils.ts` instead?

cal-smith: 👍

PR #314, opened 2018-12-12
chore(docs): update tsconfig to generate docs correctly

cal-smith:
Seems like something changed in the way compodoc looks for files, and the includes/excludes in tsconfig are more important now. Updated to have sane and useful values for docs/build.


package.json
```diff
@@ -6,8 +6,9 @@
   "scripts": {
     "build": "bash scripts/build.sh",
     "storybook": "start-storybook -s .storybook/public -p 6006",
-    "docs:build": "compodoc src -p tsconfig.json",
-    "docs:server": "compodoc src -p tsconfig.json -s -w",
+    "docs:build": "compodoc -p tsconfig.json",
+	"docs:server": "compodoc -p tsconfig.json -s -w",
```
zvonimirfras: Tabs here, spaces elsewhere.
```suggestion
    "docs:server": "compodoc -p tsconfig.json -s -w",
```

package.json
```diff
@@ -6,8 +6,9 @@
...
+	"docs": "compodoc -h",
```
zvonimirfras:
```suggestion
    "docs": "compodoc -h",
```

PR #318, opened 2018-12-12
Added missing overlay block

stanislavgeorgiev:
Closes IBM/carbon-components-angular#317

{{short description}}


cal-smith:
From what I can see this is already available in the `overlay` input? (https://github.com/IBM/carbon-components-angular/pull/318/files#diff-4109bcd3cabf1297b9205880ce21ab7fR38)

PR #321, opened 2018-12-14
fix(tooltip): DAP violation

### LINKED ISSUE #262: Tooltip DAP Violation (AVT 1)
The 'aria-describedby' attribute which is binded to the host element of the TooltipDirective (selector 'ibmTooltip') references a DOM element which only exists on hover of the host element.
So when running a DAP scan, the following DAP violation shows if the host element is not being hovered:

> The id identifier specified for WAI-ARIA property identifier on element identifier is not valid. See rule help for remediation guidance. 

>The id tooltip-1 specified for WAI-ARIA property aria-describedby on element span is not valid. See rule help for remediation guidance.

https://aat.w3ibm.mybluemix.net/token/8e33e901-068f-4fe0-941a-7a5aa0af44ca/6f511ffe-45fc-48e3-ac7f-1d0d8a6975c7/latest///doc/w3/help/en-US/idhi_accessibility_check_g1077.html

bug

chrome Version 70.0.3538.77

"carbon-components": "9.48.0",
"carbon-components-angular": "2.0.6",

IWR

Screenshot
example

youda97:
closes #262

### DIFF
--- a/src/dialog/tooltip/tooltip.directive.ts
+++ b/src/dialog/tooltip/tooltip.directive.ts
@@ -52,9 +52,7 @@ export class TooltipDirective extends DialogDirective {
 	@Input("tooltip-type") tooltipType: "warning" | "error" | "" = "";
 
 	@HostBinding("attr.aria-describedby") get descriptorId(): string {
-		if (this.expanded) {
-			return this.dialogConfig.compID;
-		}
+		return this.expanded ? this.dialogConfig.compID : null;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/dialog/tooltip/tooltip.directive.ts
```diff
@@ -51,7 +51,11 @@ export class TooltipDirective extends DialogDirective {
 	// tslint:disable-next-line:no-input-rename
 	@Input("tooltip-type") tooltipType: "warning" | "error" | "" = "";
 
-	@HostBinding("attr.aria-describedby") descriptorId: string;
+	@HostBinding("attr.aria-describedby") get descriptorId(): string {
+		if (this.expanded) {
```
zvonimirfras: `else`? Technically, yes, it's going to be `undefined` but we can be more deliberate about it. Looks a bit unfinished this way. (or a `return null` after the `if`)

Whole getter body could be

```typescript
	return this.expanded ? this.dialogConfig.compID : null;
```

PR #322, opened 2018-12-17
fix(tooltip): keyboard accessibility

### LINKED ISSUE #263: Tooltip keyboard accessibility (AVT 2)
The tooltip does not show on keyboard focus, only on mouse hover.

> Guideline 2.1 Keyboard Accessible: Make all functionality available from a keyboard.
https://www.ibm.com/able/guidelines/ci162/accessibility_checklist.html

bug

chrome Version 70.0.3538.77

"carbon-components": "9.48.0",
"carbon-components-angular": "2.0.6",

IWR

youda97:
closes #263


src/dialog/tooltip/tooltip.directive.ts
```diff
@@ -73,6 +80,20 @@ export class TooltipDirective extends DialogDirective {
 		this.dialogConfig.compID = "tooltip-" + TooltipDirective.tooltipCounter;
 		this.dialogConfig.content = this.ibmTooltip;
 		this.dialogConfig.type = this.tooltipType;
-		this.descriptorId = this.dialogConfig.compID;
+	}
+
+	close() {
+		this.dialogService.close(this.viewContainerRef, false);
```
cal-smith: Why do we not want to focus the triggering element?

src/dialog/tooltip/tooltip.directive.ts
```diff
@@ -73,6 +80,20 @@ export class TooltipDirective extends DialogDirective {
...
+		this.expanded = false;
+	}
+
+	@HostListener("keydown", ["$event"])
+	onKeydown(event: KeyboardEvent) {
```
cal-smith: This is already handled in `dialog.directive`

src/dialog/tooltip/tooltip.component.ts
```diff
@@ -43,4 +43,9 @@ export class Tooltip extends Dialog {
 	onDialogInit() {
 		this.hasContentTemplate = this.dialogConfig.content instanceof TemplateRef;
 	}
+
+	@HostListener("keydown", ["$event"])
+	onBlur(event: KeyboardEvent) {
```
cal-smith: Tooltips can have interactive elements, so we can't just close the tooltip onKeydown

src/dialog/dialog.service.ts
```diff
@@ -155,7 +155,9 @@ export class DialogService {
 			}
 			this.dialogRef = null;
 			this.isOpen = false;
-			elementToFocus.focus();
+			if (focusPreviouslyFocusedElement) {
```
cal-smith: So if we don't return focus to the triggering element, focus ends up _somewhere_ on the page ... probably not what we want to do

src/dialog/dialog.component.ts
```diff
@@ -146,7 +146,6 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
 			}
 		}
 		this.placeDialog();
-		dialogElement.focus();
```
cal-smith: Why do we not want to focus the dialog?

youda97: https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip

In the note it says `Focus stays on the triggering element while the tooltip is displayed.`

src/dialog/dialog.service.ts
youda97: So since we don't focus to the dialog, the previous focus is the trigger element. So when you focus on the trigger the focus gets stuck there.

src/dialog/tooltip/tooltip.component.ts
youda97: Yup I meant to make that onBlur

src/dialog/tooltip/tooltip.directive.ts
youda97: `dialog.directive` only handles click events

cal-smith: Which also handles enter and space (for buttons) ... if we want to make sure we handle keydowns, add the appropriate bindings in `dialog.directive`

src/dialog/dialog.component.ts
cal-smith: That doesn't apply here since we can have interactive elements in standard tooltips. For icon or definition tooltips it would ... We don't have icon or definition tooltips right now, but when we add them we'll need to consider some mechanism for tidying up the focus

src/dialog/dialog.service.ts
cal-smith: But we need to focus the dialog, so we can access potential interactive elements

youda97: I mean you can still interact with the elements inside without having the dialog focusable

src/dialog/tooltip/tooltip.directive.ts
youda97: So I have to add keydowns because enter/space was not opening the tooltip.. So ill just add it to `dialog.directive`

src/dialog/dialog.service.ts
cal-smith: Not with a keyboard AFAIK?

cal-smith: If it's not focused we can't really tab into it right?

youda97: Like for the tooltip with the template, if you are focused on the trigger element and press enter, you can tab to focus on the close button.. so it seems to be working

cal-smith: Hmm, it looks like it is, but when I tested it closer it was just cause there were no other tabbable elements on the page ... if I add a button (for example) after the tooltip span, it tabs there _then_ into the tooltip

src/dialog/dialog.component.ts
```diff
@@ -146,7 +146,9 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
...
+		if (getFocusElementList(this.dialog.nativeElement).length > 0) {
+			dialogElement.focus();
```
cal-smith: Yep, something exactly like this 👍 nice

src/dialog/dialog.directive.ts
```diff
@@ -114,6 +114,19 @@ export class DialogDirective implements OnInit, OnDestroy, OnChanges {
 		this.toggle();
 	}
 
+	@HostListener("keydown", ["$event"])
+	onKeydown(event: KeyboardEvent) {
```
cal-smith: There's already a block of observables dealing with this in `ngOnInit` ... It'd be great if we could co-locate all this event handling code ... either merging this into the `ngOnInit` block, or moving that functionality to this style.

PR #330, opened 2019-01-04
 feat(slider): add slider component 

### LINKED ISSUE #73: Add Slider component

cal-smith:
closes #73


src/slider/slider.component.ts
```diff
@@ -0,0 +1,328 @@
...
+
+	/** Receives a value from the model */
+	writeValue(v: any) {
+		this.value = v;
+	}
+
+	/** Returns the amount of "completeness" as a fraction of the total track width */
+	getPercentComplete() {
+		const trackWidth = this.track.nativeElement.getBoundingClientRect().width;
+		return this.slidAmount / trackWidth;
```
zvonimirfras: If I'm reading this right, this might have to be multiplied by a 100, or function should change the name.
```suggestion
		return this.slidAmount / trackWidth * 100;
```

src/slider/slider.component.ts
```diff
@@ -0,0 +1,328 @@
...
+	onChange(event) {
+		this.value = event.target.value;
+	}
+
+	/** Handles clicks on the range track, and setting the value to it's "real" equivalent */
+	onClick(event) {
+		if (this.disabled) { return; }
+		const trackLeft = this.track.nativeElement.getBoundingClientRect().left;
+		this.value = this.convertToValue(event.clientX - trackLeft);
+		console.log(event);
```
zvonimirfras: Leftover?
```suggestion
```

src/slider/slider.component.ts
```diff
@@ -0,0 +1,328 @@
...
+
+	/** Focus handler for the optional input */
+	onFocus({target}) {
+		target.select();
+	}
+
+	/** Mouse move handler. Responsible for updating the value and visual selection based on mouse movement */
+	onMouseMove(event) {
+		if (this.disabled) { return; }
+		if (this.isMouseDown) {
```
zvonimirfras: This could also be a guard (like for disabled), indicating that function does nothing if mouse isn't down.
It could be combined withing the same if statement with disabled.

src/slider/slider.component.ts
```diff
@@ -0,0 +1,328 @@
...
+			const trackWidth = this.track.nativeElement.getBoundingClientRect().width;
```
zvonimirfras: We can get away with calling `getBoundingClientRect()` just once without loosing clarity by setting
`const track = this.track.nativeElement.getBoundingClientRect();`
and then using `track.width` and `track.left`

src/slider/slider.component.ts
cal-smith: Percent, fraction, it's all the same right? 😉 

(fixed)

PR #337, opened 2019-01-08
feat(accordion): Add skeleton state

youda97:


src/accordion/accordion.component.ts
```diff
@@ -8,6 +15,12 @@ import { Component } from "@angular/core";
...
 	`
 })
-export class Accordion {
+export class Accordion implements AfterContentInit {
+	@ContentChildren(AccordionItem) childs: QueryList<AccordionItem>;
 
+	@Input() skeleton = false;
+
+	ngAfterContentInit() {
+		this.childs.toArray().forEach(child => child.skeleton = this.skeleton);
```
cal-smith: Same issue as breadcrumbs, we'll need something else to update the child state when the input changes. Also `children` not `childs`

src/accordion/accordion-item.component.ts
```diff
@@ -21,17 +21,29 @@ import {
...
+				class="bx--accordion__title"
+				[ngClass]="{
+					'bx--skeleton__text': skeleton
+				}">
+				{{!skeleton ? title : null}}
+			</p>
 		</button>
 		<div [id]="id" class="bx--accordion__content">
-			<ng-content></ng-content>
+			<ng-content *ngIf="!skeleton; else skeletonTemp"></ng-content>
```
zvonimirfras:
```suggestion
			<ng-content *ngIf="!skeleton; else skeletonTemplate"></ng-content>
```

src/accordion/accordion-item.component.ts
```diff
@@ -21,17 +21,29 @@ import {
...
+			<ng-template #skeletonTemp>
```
zvonimirfras:
```suggestion
			<ng-template #skeletonTemplate>
```

src/accordion/accordion-item.component.ts
```diff
@@ -21,17 +21,29 @@ import {
 				viewBox="0 0 7 12">
           		<path fill-rule="nonzero" d="M5.569 5.994L0 .726.687 0l6.336 5.994-6.335 6.002L0 11.27z"/>
 			</svg>
-			 <p class="bx--accordion__title">{{title}}</p>
+			<p
+				class="bx--accordion__title"
+				[ngClass]="{
+					'bx--skeleton__text': skeleton
+				}">
+				{{!skeleton ? title : null}}
```
zvonimirfras: Is it bad if we show the title regardless if it's available?

youda97: If its available it will show in the loading state.. which we don't want happening

PR #338, opened 2019-01-08
 feat(breadcrumb): Add skeleton state

youda97:


src/breadcrumb/breadcrumb.component.ts
```diff
@@ -71,6 +80,10 @@ export class Breadcrumb {
 		return this._threshold;
 	}
 
+	ngAfterContentInit() {
```
cal-smith: this won't let us un-set the skeleton state ... we'll need something else to update the child skeleton state when the input changes. Also `children` instead of `childs`

PR #342, opened 2019-01-08
feat(dropdown): Add skeleton state

youda97:


src/dropdown/dropdown.component.ts
```diff
@@ -198,33 +213,37 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy {
...
-		this.view.select.subscribe(event => {
-			if (this.type === "multi") {
-				this.propagateChange(this.view.getSelected());
-			} else {
-				this.closeMenu();
-				this.dropdownButton.nativeElement.focus();
-				if (event.item && event.item.selected) {
-					if (this.value) {
-						this.propagateChange(event.item[this.value]);
+		if (!this.skeleton) {
```
cal-smith: this may introduce some bugs when transitioning from a skeleton to non-skeleton state

src/dropdown/dropdown.component.ts
```diff
@@ -198,33 +213,37 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy {
...
-		this.view.select.subscribe(event => {
-			if (this.type === "multi") {
-				this.propagateChange(this.view.getSelected());
-			} else {
-				this.closeMenu();
-				this.dropdownButton.nativeElement.focus();
-				if (event.item && event.item.selected) {
-					if (this.value) {
-						this.propagateChange(event.item[this.value]);
+		if (this.view) {
```
zvonimirfras: It's more readable to put a guard than wrapping the whole content of the function.
 ```suggestion
		if (!this.view) {
			return;
		}
```
and then have the rest of the function unchanged/as it was before this PR.

File: src/dropdown/dropdown.component.ts [hunk 3]
```diff
@@ -323,17 +342,19 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy {
 	 * Returns the display value if there is no selection, otherwise the selection will be returned.
 	 */
 	getDisplayValue(): Observable<string> {
-		let selected = this.view.getSelected();
-		if (selected && !this.displayValue) {
-			if (this.type === "multi") {
-				return of(`${selected.length} ${this.selectedLabel}`);
-			} else {
-				return of(selected[0].content);
+		if (this.view) {
```
zvonimirfras: Looks like another candidate for a guard.

src/dropdown/dropdown.component.ts
```diff
@@ -93,6 +100,14 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy {
...
 	 */
 	@Input() disabled = false;
+	/**
+	 * Set to `true` for a loading dropdown.
+	 */
+	@Input() skeleton = false;
+	/**
+	 * Set to `true` for an inline dropdown.
+	 */
+	@Input() inline = false;
```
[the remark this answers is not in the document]
youda97: yup

PR #347, opened 2019-01-10
feat(search): Add skeleton state

youda97:


src/search/search.stories.ts
```diff
@@ -21,4 +21,12 @@ storiesOf("Search", module).addDecorator(
 			disabled: boolean("disabled", false),
 			placeholder: text("placeholder", "Search")
 		}
+	}))
+	.add("Skeleton", () => ({
+		template: `
+		<div style="width: 200px;">
+			<ibm-search skeleton="true"></ibm-search>&nbsp;
```
zvonimirfras: Just because that's the way you do it in 3 other PRs
```suggestion
			<ibm-search skeleton="true"></ibm-search>
			&nbsp;
```

PR #358, opened 2019-01-14
Add temporary experimental service

cal-smith:
So consumers on the 2.x branch can see what's coming. Will be removed with 3.x.

### DIFF
--- a/.storybook/experimental.component.ts
+++ b/.storybook/experimental.component.ts
@@ -1,0 +1,34 @@
+import { Component, AfterViewInit, OnDestroy } from "@angular/core";
+import { ExperimentalService } from "../src/experimental.module";
+
+@Component({
+	selector: "app-experimental-component",
+	template: ``
+})
+export class ExperimentalComponenent implements AfterViewInit, OnDestroy {
+	constructor(protected experimental: ExperimentalService) {
+		experimental.isExperimental = false;
+	}
+
+	ngAfterViewInit() {
+		this.toggleExperimental();
+		document.querySelectorAll(".toggle-btn").forEach(button => {
+			button.addEventListener("click", this.toggleExperimental);
+		});
+	}
+
+	ngOnDestroy() {
+		document.querySelectorAll(".toggle-btn").forEach(button => {
+			button.removeEventListener("click", this.toggleExperimental);
+		});
+	}
+
+	// arrow function to capture the correct `this`
+	toggleExperimental = () => {
+		if (document.body.classList.contains("experimental")) {
+			this.experimental.isExperimental = true;
+		} else {
+			this.experimental.isExperimental = false;
+		}
+	}
+}
--- a/.storybook/preview.scss
+++ b/.storybook/preview.scss
@@ -1,1 +1,59 @@
-@import "~carbon-components/src/globals/scss/styles";
+$feature-flags: (
+	components-x: false,
+	ui-shell: true,
+	css--body: false, // we're providing our own body styles
+	css--reset: false // prevent thousands of resets being included...
+);
+
+.carbon {
+	@import '~carbon-components/scss/globals/scss/styles';
+
+	// carbon body reset and styles
+	// @include reset;
+	@include font-family;
+	color: $text-01;
+	background-color: $ui-02;
+	line-height: 1;
+
+	@include typography;
+
+	// padding/margin body reset
+	padding: 0;
+	margin: 0;
+
+	// for some reason this isn't included when tabs are built. So lets just manually include it here.
+	.bx--tabs__nav-item + .bx--tabs__nav-item {
+		margin-left: 3rem;
+	}
+
+	// for some reason this isn't included when progress indicator is built. So lets just manually include it here.
+	.bx--progress-step:first-child .bx--progress-line {
+		display: none;
+	}
+
+	// for some reason all this isn't included when structured list is built. So lets just manually include it here.
+	.bx--structured-list--selection .bx--structured-list-row:hover:not(.bx--structured-list-row--header-row) {
+		background-color: $hover-row;
+		cursor: pointer;
+	}
+
+	.bx--structured-list-row:focus:not(.bx--structured-list-row--header-row) {
+		@include focus-outline('border');
+	}
+
+	.bx--structured-list-svg {
+		display: inline-block;
+		fill: transparent;
+		vertical-align: middle;
+		transition: $transition--base $carbon--standard-easing;
+	}
+
+	.bx--structured-list-row:hover .bx--structured-list-svg {
+		fill: $hover-row;
+	}
+
+	.bx--structured-list-input:checked + .bx--structured-list-row .bx--structured-list-svg,
+	.bx--structured-list-input:checked + .bx--structured-list-td .bx--structured-list-svg {
+		fill: $brand-02;
+	}
+}
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "@types/node": "8.5.2",
     "angular2-template-loader": "0.6.2",
     "babel-loader": "8.0.4",
-    "carbon-components": "9.56.2",
+    "carbon-components": "9.68.16",
     "codelyzer": "4.5.0",
     "commitizen": "2.10.1",
     "core-js": "2.5.5",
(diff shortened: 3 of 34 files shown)


package.json
```diff
@@ -97,7 +97,7 @@
     "@types/node": "8.5.2",
     "angular2-template-loader": "0.6.2",
     "babel-loader": "8.0.4",
-    "carbon-components": "9.56.2",
+    "carbon-components": "^9.66.3",
```
youda97:
```suggestion
    "carbon-components": "9.66.3",
```

src/experimental.module.ts
```diff
@@ -0,0 +1,27 @@
+import { NgModule, SkipSelf, Optional } from "@angular/core";
+import { ExperimentalService } from "./experimental.service";
+
+// either provides a new instance of ModalPlaceholderService, or returns the parent
```
zvonimirfras:
```suggestion
// either provides a new instance of ExperimentalService, or returns the parent
```

src/experimental.module.ts
```diff
@@ -0,0 +1,27 @@
...
+export function EXPERIMENTAL_SERVICE_PROVIDER_FACTORY(parentService: ExperimentalService) {
+	return parentService || new ExperimentalService();
+}
+
+// placholder service *must* be a singleton to ensure the placeholder viewref is accessible globally
```
zvonimirfras: Doesn't look relevant.
```suggestion
```

src/modal/modal-header.component.ts
```diff
@@ -63,15 +76,18 @@ export class ModalHeader {
 
 	/**
 	 * To emit the event of clicking on the close icon within the modal.
-	 * @memberof ModalHeader
 	 */
 	@Output() closeSelect = new EventEmitter();
 
-	constructor(protected i18n: I18n) {}
+	get isExperimental() {
+		console.log(this.experimental.isExperimental);
```
zvonimirfras:
```suggestion
```

src/pagination/pagination.component.ts
```diff
@@ -41,60 +42,85 @@ import { BehaviorSubject } from "rxjs";
...
+					[for]="itemsPerPageSelectId"
+					class="bx--label bx--visually-hidden">
+					{{itemsPerPageText | async}}
+				</label>
+				<select
+					[id]="itemsPerPageSelectId"
+					[(ngModel)]="itemsPerPage"
+					class="bx--select-input"
+					aria-describedby="false">
+					<option class="bx--select-option" value="10">{{itemsPerPageText | async}} 10</option>
```
zvonimirfras: Carbon components experimental demo has the number first, and no colon. We can probably ignore colon for now but open an issue to change it before X.
```suggestion
					<option class="bx--select-option" value="10">10 {{itemsPerPageText | async}}</option>
```

zvonimirfras: Also, pagination story doesn't look right in standard any more.

.storybook/experimental.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component, OnInit, AfterViewInit } from "@angular/core";
+import { ExperimentalService } from "../src/experimental.module";
+
+@Component({
+	selector: "app-experimental-component",
+	template: ``
+})
+export class ExperimentalComponenent implements AfterViewInit {
+	constructor(protected experimental: ExperimentalService) {
+		experimental.isExperimental = false;
```
zvonimirfras: I can't find any place in the PR where experimental is set to true, what am I missing?

cal-smith: We could drop this, the main benefit is it gives us a single place to handle storybooks experimental state. For actual development we need to toggle the `preview.scss` flag and this flag.

zvonimirfras: The functionality part and templates still change based on `ExperimentalService`'s `isExperimental`, no?

cal-smith: > templates still change based on ExperimentalService's isExperimental

Let me poke it again to make sure, but it wasn't updating dynamically.

cal-smith: 🤦‍♂️ yeah I totally messed up the first time I tried implementing ... works way better now

.storybook/preview.scss
```diff
@@ -1,9 +1,18 @@
...
 
 body {
-	padding: 20px;
 	// for tooltips/popovers ... can be on any chld element before the popover
 	// but body is most conveient
 	position: relative;
+	display: flex;
+}
+
+#root {
```
youda97: Not sure what this does.. but it removes the center alignment styles that is on the body tag

src/dialog/tooltip/tooltip.stories.ts
```diff
@@ -16,30 +16,34 @@ storiesOf("Tooltip", module)
...
 	.add("Basic", () => ({
 		template: `
-			<span
-				ibmTooltip="tooltip text"
-				trigger="hover"
-				placement="bottom"
-				style="cursor: pointer;">
-				Hover for tooltip
-			</span>
+			<div style="padding: 80px;">
```
youda97: Not needed since it is already center aligned from my last PR

src/dialog/tooltip/tooltip.stories.ts
```diff
@@ -16,30 +16,34 @@ storiesOf("Tooltip", module)
...
-			</ng-template>
-			<span
-				[ibmTooltip]="template"
-				trigger="click"
-				placement="bottom"
-				style="cursor: pointer;">
-				Click for tooltip
-			</span>
-			<ibm-placeholder></ibm-placeholder>
+			<div style="padding: 80px;">
```
youda97: Also not needed cz story is already center aligned

youda97:
Just some things I noticed by looking at the preview:

- It seems that the font style is always on experimental (seen on the welcome page and in other stories).
- The close button on experimental notifications seems like it is still not aligned correctly
- For our standard components it looks like some have been altered. From what I can see the tabs, clickable tiles, progress indicator and structured list (selection) are different.
- Also, on line 578 in `table.stories` you forgot to change selectV2 to select

.storybook/preview.scss
cal-smith: That was fixing things before everything was centered

cal-smith:
For reasons I cannot explain, carbon seems to rely on owning the document body. So applying classes nested from the body (i.e. `.carbon { @import "carbon"; }` just doesn't work correctly - a whole bunch of stuff doesn't get applied.

Of course I switched it to nested imports because `.experimental { @import "carbon"; }` _also_ doesn't work correctly (at least not well enough for actual dev purposes) ... I think the correct solution is dynamically changing the prefix, plus some hacky css tricks to make the typography behave.

.storybook/preview.scss
```diff
@@ -1 +1,59 @@
...
+	}
+
+	// for some reason all this isn't included when structured list is built. So lets just manually include it here.
+	.bx--structured-list--selection .bx--structured-list-row:hover:not(.bx--structured-list-row--header-row) {
+		background-color: $hover-row;
+		cursor: pointer;
+	}
+
+	.bx--structured-list-row:focus:not(.bx--structured-list-row--header-row) {
+      @include focus-outline('border');
```
youda97: indenting

.storybook/preview.scss
```diff
@@ -1 +1,59 @@
...
+
+	// for some reason all this isn't included when structured list is built. So lets just manually include it here.
+	.bx--structured-list--selection .bx--structured-list-row:hover:not(.bx--structured-list-row--header-row) {
+		background-color: $hover-row;
+		cursor: pointer;
+	}
+
+	.bx--structured-list-row:focus:not(.bx--structured-list-row--header-row) {
+		@include focus-outline('border');
+    }
```
youda97: Forgot to indent the bracket :P

cal-smith: oh nooooooo

youda97:
Conflict with table

src/index.ts
```diff
@@ -34,3 +34,5 @@ export * from "./switch/switch.module";
 export * from "./table/table.module";
 export * from "./tabs/tabs.module";
 export * from "./tiles/tiles.module";
+// experimental ... to be removed
```
zvonimirfras: Let's keep it :)
Could be useful for future updates. Experimental features can then be always available to our users.
```suggestion
```

cal-smith: Agreed!

PR #377, opened 2019-01-25
 feat(table): Add table toolbar

youda97:


youda97:
Not really... the toolbar is just a parent class that you can put whatever in

PR #379, opened 2019-01-28
feat(search): Add toolbar search

youda97:


src/search/search.component.ts
```diff
@@ -59,15 +63,25 @@ export class SearchChange {
...
-					height="16"
-					viewBox="0 0 16 16">
-					<path
-						d="M6.5 12a5.5 5.5 0 1 0 0-11 5.5 5.5 0 0 0 0 11zm4.936-1.27l4.563 4.557-.707.708-4.563-4.558a6.5 6.5 0 1 1 .707-.707z"
-						fill-rule="nonzero"/>
-				</svg>
+				<button
+					*ngIf="toolbar; else svg"
+					class="bx--toolbar-search__btn"
+					aria-label="Toolbar search"
```
zvonimirfras: Should be translatable.

src/search/search.component.ts
```diff
@@ -156,12 +174,14 @@ export class Search implements ControlValueAccessor {
 	 */
 	@Output() change = new EventEmitter<SearchChange>();
 
+	active = false;
```
zvonimirfras: Let's make this an input so it can be programmatically controlled from outside.

PR #386, opened 2019-02-05
chore(side-nav): make the host work as <aside>

asudoh:
This change removes wrapper element in side nav by putting `<aside>`'s semantic role to the host element, so we don't need a custom style rule here. Not sure if it makes sense to the team, just a food for thoughts.

#### Changelog

**Changed**

* Merged `<aside>` with the host element of `<ibm-sidenav>`.

### DIFF
--- a/src/ui-shell/sidenav.component.ts
+++ b/src/ui-shell/sidenav.component.ts
@@ -61,7 +61,7 @@ import { I18n } from "./../i18n/i18n.module";
 export class SideNav {
 	@HostBinding("attr.role") role = "complementary";
 	@HostBinding("class.bx--side-nav") hostClass = true;
-	@HostBinding("class.bx--side-nav--expanded") expanded = false;
+	@HostBinding("class.bx--side-nav--expanded") @Input() expanded = false;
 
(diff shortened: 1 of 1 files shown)


src/ui-shell/sidenav.component.ts
```diff
@@ -9,70 +9,59 @@ import { I18n } from "./../i18n/i18n.module";
...
-	],
 	encapsulation: ViewEncapsulation.None
 })
 export class SideNav {
-	@Input() expanded = false;
-
-	@HostBinding("class.bx--side-nav--wrapper") hostClass = true;
+	@HostBinding("attr.role") role = "complementary";
+	@HostBinding("class.bx--side-nav") hostClass = true;
+	@HostBinding("class.bx--side-nav--expanded") expanded = false;
```
cal-smith: Just one small suggestion, but otherwise makes sense to me 👍 

```suggestion
	@HostBinding("class.bx--side-nav--expanded") @Input() expanded = false;
```

asudoh: Thanks @cal-smith for catching this!

PR #390, opened 2019-02-05
 fix(dropdown): don't focus the button on load 

### LINKED ISSUE #387: Dropdown focusess itself on init
Repro:

- add an input to the basic dropdown story
- watch as focus immediately pops onto the dropdown

cal-smith:
closes #387 

The dropdown was applying focus to the button in the `view.select` subscription. Two things to fix that:
1. move the focus to `closeMenu`
2. return early from `closeMenu` if the menu is already closed (should avoid another entire class of errors too as a bonus)


src/dropdown/dropdown.stories.ts
```diff
@@ -17,6 +17,7 @@ storiesOf("Dropdown", module)
 	.add("Basic", () => ({
 		template: `
 		<div style="width: 300px">
+			<input type="text"/>
```
zvonimirfras: Looks like something used for testing, do we want to keep it here?

cal-smith: Oops, yep that definitely shouldn't be there

PR #391, opened 2019-02-05
 feat(table): keyboard accessibility

youda97:


cal-smith:
Tests seem to be failing

src/table/table.stories.ts
```diff
@@ -164,7 +168,8 @@ class DynamicTableStory implements OnInit {
 			[size]="size"
 			[showSelectionColumn]="showSelectionColumn"
 			[striped]="striped"
-			(sort)="customSort($event)">
+			[isDataGrid]="isDataGrid"
+>
```
cal-smith: trailing `>`

youda97:
(1) We need `columnIndex` because if we do it the way you suggested then, for example, when you go into the expanded row from column 2, the index will go to 0. So we need a way to save the previous value before going into the expanded row.

(2) If for example, you had a dropdown then when you press `up/down` in that cell you would be stuck there.

cal-smith:
How does navigation into and out of an expanded row work? Ex. would it be more or less jarring to start at the first column again? Especially considering expanded rows can contain anything (including more tables)

src/table/table.component.ts
```diff
@@ -843,4 +916,86 @@ export class Table {
...
+			return this.model.header.length + 2;
+		} else if (this.model.hasExpandableRows() || this.showSelectionColumn) {
+			return this.model.header.length + 1;
+		} else {
+			return this.model.header.length;
+		}
+	}
+
+	// This is handled in the component so that the columnIndex is saved as you go into expandable rows
+	handleInteractions(event: KeyboardEvent) {
```
zvonimirfras: Directive can know (based on classList) if it's extended row. `columnIndex` can use double binding to update values. This won't be necessary.

src/utils/a11y.ts
```diff
@@ -95,3 +97,14 @@ export function focusPrevElem(elem, parentRef = null) {
 		}
 	}
 }
+
+export function setTabIndex(element: HTMLElement, index: -1 | 0 | 1) {
```
zvonimirfras: Why limit to just those?
```suggestion
export function setTabIndex(element: HTMLElement, index: number) {
```

cal-smith: That was my suggestion ... since tabindex `-1 | 0` are the only really useful options (we shouldn't be specifying positive indexes, since that will break DOM ordering). So, my bad on suggesting to include `1` in the range

zvonimirfras: If this is in utils, our users can also use it in their apps (why wouldn't they) so it makes sense to allow them to set ordering. On a closer look, this function looks table specific so it should be a static function on table or sth.

src/table/data-grid-focus.directive.ts
```diff
@@ -0,0 +1,139 @@
...
+					} else {
+						const firstBodyCell = rows[1].querySelectorAll("td")[0];
+						setTabIndex(firstBodyCell, 0);
+						this.focus(firstBodyCell);
+					}
+				} else if (element.parentElement.className !== "bx--expandable-row-v2") {
+					const firstRowCell = rows[rowIndex].querySelectorAll("th, td")[0];
+					setTabIndex(firstRowCell, 0);
+					this.focus(firstRowCell);
+				}
```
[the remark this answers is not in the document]
youda97: Nothing happens

src/table/data-grid-focus.directive.ts
```diff
@@ -0,0 +1,139 @@
...
+					this.focus(previousSibling);
+				}
+				break;
+			case "Down": // IE specific value
+			case "ArrowDown":
+				if (rowIndex < rows.length - 1) {
+					rowIndex++;
+					const row = rows[rowIndex].querySelectorAll("td");
+					setTabIndex(element, -1);
+					if (rows[rowIndex].className === "bx--expandable-row-v2") {
```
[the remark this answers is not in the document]
youda97: the expanded row and the row before it both have this class. So I have to use `className`

src/table/data-grid-focus.directive.ts
```diff
@@ -0,0 +1,139 @@
...
+				if (event.ctrlKey) {
+					this.columnIndex = lastRow.length - 1;
+					setTabIndex(lastRow[this.columnIndex], 0);
+					this.focus(lastRow[this.columnIndex]);
+				} else if (element.parentElement.className !== "bx--expandable-row-v2") {
+					this.columnIndex = lastRow.length - 1;
+					const lastRowCell = rows[rowIndex].querySelectorAll("th, td")[this.columnIndex];
+					setTabIndex(lastRowCell, 0);
+					this.focus(lastRowCell);
+				}
```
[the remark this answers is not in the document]
youda97: Nothing

src/utils/a11y.ts
cal-smith: Is there a table utils it could live in? Having it as a static method on table means that other classes need to import table to use it, which doesn't look immediately correct

zvonimirfras: Sure, that would work too. Using `Table.setTabIndex()` looks right, though, considering it's a table specific function for setting tab index. Either way makes sense, it's a stylistic preference, I'm good with either, they have the same spirit.

PR #392, opened 2019-02-06
feat(progress-indicator): adds 'vertical' variant

### LINKED ISSUE #384: Add progress indicator vertical
Once https://github.com/IBM/carbon-components/pull/1717 lands we should make sure we support it too.

werdnanoslen:
fixes #384

Adds a vertical variant to the progress-indicator widget, as defined in https://github.com/IBM/carbon-components/pull/1717

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "core-js": "^2.5.3",
     "rxjs": "^6.0.0",
     "zone.js": "^0.8.26",
-    "carbon-components": "^9.20.0"
+    "carbon-components": "^9.70.0"
   },
   "devDependencies": {
     "@angular-devkit/build-angular": "0.7.1",
@@ -98,7 +98,7 @@
     "@types/node": "8.5.2",
     "angular2-template-loader": "0.6.2",
     "babel-loader": "8.0.4",
-    "carbon-components": "9.68.16",
+    "carbon-components": "^9.70.0",
     "codelyzer": "4.5.0",
     "commitizen": "2.10.1",
     "core-js": "2.5.5",
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "@types/node": "8.5.2",
     "angular2-template-loader": "0.6.2",
     "babel-loader": "8.0.4",
-    "carbon-components": "^9.70.0",
+    "carbon-components": "9.70.0",
     "codelyzer": "4.5.0",
     "commitizen": "2.10.1",
     "core-js": "2.5.5",
--- a/src/progress-indicator/progress-indicator.component.ts
+++ b/src/progress-indicator/progress-indicator.component.ts
@@ -10,7 +10,7 @@ import { ExperimentalService } from "./../experimental.module";
 		class="bx--progress"
 		[ngClass]="{
 			'bx--skeleton': skeleton,
-			'bx--progress--vertical': vertical
+			'bx--progress--vertical': (orientation === 'vertical')
 		}">
 		<li
 		class="bx--progress-step bx--progress-step--{{step.state}}"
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -98,7 +98,7 @@
     "@types/node": "8.5.2",
     "angular2-template-loader": "0.6.2",
     "babel-loader": "8.0.4",
-    "carbon-components": "9.68.16",
+    "carbon-components": "^9.70.0",
```
cal-smith: We err on the side of caution and pin all our direct/dev deps

```suggestion
    "carbon-components": "9.70.0",
```

src/progress-indicator/progress-indicator.component.ts
```diff
@@ -59,7 +60,7 @@ export class ProgressIndicator {
 	}
 
 	@Input() steps = [];
-
+	@Input() vertical = false;
```
cal-smith: I'm not sure if `@Input() type: "horizontal" | "vertical" = "horizontal";` or `@Input() direction: "horizontal" | "vertical" = "horizontal";` would make more sense ... I'm good with `@Input() vertical = false` though so I don't think it _needs_ to change.

cal-smith:
Thanks for getting this done btw 🎉

src/progress-indicator/progress-indicator.component.ts
werdnanoslen: I just kept it 1:1 with the accepted PR's style shown here: https://github.com/IBM/carbon-components/pull/1717/files#diff-6e73227357afc271c1386a942097df26R61. If we were to depart from that here, the original component should be updated to use the same variable assignment style.

cal-smith: We do deviate from the vanilla implementation where it makes sense from an API perspective - ex. in dropdown `@Input() type: "single" | "multi" = "single"` ... That seems to be the exception rather then the rule though, since we have a bunch of components with modifiers like this now, so it's going to be more consistent this way

werdnanoslen: @cal-smith By being more consistent, do you mean we should continue the exception and use your suggested change, or avoid deviations and not commit your suggested change?

cal-smith: Sorry that was ambiguous, I mean avoid deviations.

I'll leave this open if anyone else has a strong opinion, but I'm good to give this an approval 👍

zvonimirfras: This _should_ be `type`. It goes very well with how we use `type` in combobox, overflow-menu-option, dropdown, dropdown-list...

werdnanoslen: @zvonimirfras but is that a frequent exception as @cal-smith said, or is that the design direction/rule across the board going forward? Totally open to it myself, just wanna be sure I'm committing based on where we're going, not where we are.

cal-smith: We took some time to think it over ... how about `orientation: "horizontal" | "vertical" = "horizontal"` ... we're trying to establish some primitives we can use across the codebase, and `orientation` makes a lot of sense :+1:

cal-smith: Sorry for the delay!

werdnanoslen: no prob, how's this? 1df8fce5c7f2b96f4ccbbefc60642dafc2ed33be

cal-smith: Perfect 👍

cal-smith:
Looks like `package-lock.json` and `package.json` have some conflicts ... once the ones in `package.json` are fix just regenerate `package-lock.json` and we'll be good to go 🎉

werdnanoslen:
@cal-smith resolved

PR #395, opened 2019-02-06
Use the correct href when not displaying a skeleton

markewallace:
Closes IBM/carbon-components-angular#394

Use the correct href when not displaying a skeleton

#### Changelog

**Changed**

* Use the correct href when not displaying a skeleton


cal-smith:
Good catch 😬

cal-smith:
Just going to close and reopen so travis will pick this up ...

cal-smith:
Travis is happy! Merge!

PR #398, opened 2019-02-06
Trigger `selectRow` or `deselectRow` when a row is clicked.

stanislavgeorgiev:
Closes IBM/carbon-components-angular#393


stanislavgeorgiev:
I added an `action()` logger but it doesn't get anything out... I know it gets executed because I tested it with a `console.log()`. I saw that you're using the action logger in other places and it does work fine there. Any suggestions?

cal-smith:
Just going to close and reopen so travis will pick this up ...

cal-smith:
I believe `action` returns a function and that would be passed to the `Output` ... the first parameter is just a name that it logs with ... it may work if you have `selectRow` just `return action("row selected");` (and so `(selectRow)="selectRow($event)"` would just become `(selectRow)="selectRow()"`), otherwise you'll need to move it into the story props and add an `Output` in the story component

stanislavgeorgiev:
OK, I tried the `return action()` but it didn't change anything.
Anyways, that's not a hold up for the delivery I hope.

stanislavgeorgiev:
What's the ETA on the second review and merging?

src/table/table.stories.ts
```diff
@@ -37,8 +38,10 @@ import { clone } from "../utils/utils";
 			[model]="model"
 			[size]="size"
 			[showSelectionColumn]="showSelectionColumn"
+			[enableSingleSelect]="enableSingleSelect"
```
zvonimirfras: This is actually not officially supported by current carbon design so we're not going to surface it in the story.

stanislavgeorgiev: Sure, didn't know that. It's all reverted now except the fix.

PR #400, opened 2019-02-07
fix(dialog): position wrong due to scroll x/y

### LINKED ISSUE #399: Overflow menu bug on the page with scroll
## Detailed description

We have a lot of tables in our application. There are overflow menu in each table row. If we scroll page overflow menu appears much lower than we expected. If we don't scroll page or set scroll at the top - Overflow menu appears in the right place.

a bug

OverflowMenu

I expect overflow menu appears near three vertical dots. Instead it appears much lower.

Chrome

carbon-components-angular:  2.29.0

https://codesandbox.io/s/y369zqxol1

cal-smith:
closes #399 

I can't actually find any specific justification for keeping this, and from what I can tell it's already handled elsewhere in dialog/position ... so away it goes!


cal-smith:
Also 400 issues/PRs 🎉

PR #403, opened 2019-02-12
Table: Expanded row doesn't filter.

stanislavgeorgiev:
Closes IBM/carbon-components-angular#378


stanislavgeorgiev:
Thanks guys, that was quick to say the least! :+1: :1st_place_medal:

PR #406, opened 2019-02-14
fix(dropdown): remove scroll addition to re-fix positioning

cal-smith:
Same as the fix for dialog ... seems like improvements to position.service create conflicts with adding the window scroll


thesupernvk:
We are still getting this issue. Versions of carbon used:
`"carbon-components": "^10.4.1",
"carbon-components-angular": "^3.14.2"`

Angular version is 8.1.x.

Dropdown menu is still positioned incorrectly when page is scrolled.

cal-smith:
see https://github.com/IBM/carbon-components-angular/pull/695

PR #415, opened 2019-02-19
Dropdown updates

cyrusfurtado:

Dropdown component in `multi select` has been given a clickable count button which resets all the checkboxes on click. Also the each list item has been given an additional secondary content for display.

#### Changelog

**Changed**

* src/dropdown/abstract-dropdown-view.class.ts
* src/dropdown/dropdown.component.ts
* src/dropdown/dropdown.stories.ts
* src/dropdown/list-item.interface.ts
* src/dropdown/list/dropdown-list.component.ts

drop_down_demo


src/dropdown/dropdown.component.ts
```diff
@@ -62,6 +73,52 @@ import { I18n } from "./../i18n/i18n.module";
 		</div>
 	</div>
 	`,
+	styles: [`.bx--list-box {
```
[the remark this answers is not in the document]
cyrusfurtado: So i would need to modify the vanilla `carbon-components` library for this.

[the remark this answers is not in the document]
cal-smith: Why? Looking at the demo the multi-select clear selection button is already supported.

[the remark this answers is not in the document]
cyrusfurtado: Okay. So could add `scss` files in this repo then. Cause i couldn't find any other component with `scss` files included.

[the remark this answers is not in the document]
cal-smith: Which repo? The clear selection button is already supported in carbon-components, there shouldn't be any more `scss` required.

src/dropdown/list/dropdown-list.component.ts
```diff
@@ -409,4 +411,10 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnChan
 		this.listElementList[index].classList.remove("bx--list-box__menu-item--highlighted");
 		this.listElementList[index].tabIndex = -1;
 	}
+
+	resetSelected() {
+		this.items = this.items.map((item) => {
```
cal-smith: Probably want to do 
```ts
const clearedItems = this.getListItems().map(item => Object.assign({}, item, {selected: false}));
this.updateList(clearedItems);
```

`this.items` could return an observable, so we have `getListItems` which will always return the resolved `ListItem[]`. `updateList` also handles a few housekeeping tasks we need to do when a list is updated, so it's best to route everything through there.

src/dropdown/dropdown.component.ts
```diff
@@ -35,24 +35,35 @@ import { I18n } from "./../i18n/i18n.module";
...
 			(blur)="onBlur()"
-			[disabled]="disabled">
-			<span class="bx--list-box__label">{{getDisplayValue() | async}}</span>
-			<div *ngIf="!skeleton" class="bx--list-box__menu-icon" [ngClass]="{'bx--list-box__menu-icon--open': !menuIsClosed }">
-				<svg fill-rule="evenodd" height="5" role="img" viewBox="0 0 10 5" width="10" [attr.aria-label]="menuButtonLabel">
-					<title>{{menuButtonLabel}}</title>
-					<path d="M0 0l5 4.998L10 0z"></path>
+			>
+			<span (click)="clearSelected()" *ngIf="type === 'multi' && view.getSelected() && view.getSelected().length"
+				class="bx--tag bx--tag--ibm">
```
cal-smith: This should be `bx--list-box__selection bx--list-box__selection--multi` ... dropdown already has specific styles for this in the carbon-components css :+1: be sure to check out the rendered output of https://v9.carbondesignsystem.com/components/dropdown/code (specifically the multiselect) to see the structure

src/dropdown/dropdown.component.ts
```diff
@@ -345,7 +357,7 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy {
 		let selected = this.view.getSelected();
 		if (selected && !this.displayValue) {
 			if (this.type === "multi") {
-				return of(`${selected.length} ${this.selectedLabel}`);
+				return of(`${this.selectedLabel}`);
```
cal-smith: this whole method (`getDisplayValue`) should be something like:

```ts
    /**
     * Returns the display value if there is a selection and displayValue is set,
     * if there is just a selection the ListItem content property will be returned,
     * otherwise the placeholder will be returned.
     */
    getDisplayValue(): Observable<string> {
        if (!this.view) {
            return;
        }
        let selected = this.view.getSelected();
        if (selected && !this.displayValue) {
            if (this.type === "multi") {
                return of(this.placeholder);
            } else {
                return of(selected[0].content);
            }
        } else if (selected) {
            return of(this.displayValue);
        }
        return of(this.placeholder);
    }
```

This always returns the placeholder value when we're multi-selecting (unless we have a displayValue, which always overrides)

src/dropdown/dropdown.component.ts
```diff
@@ -43,15 +43,28 @@ import { DropdownService } from "./dropdown.service";
...
 			[attr.aria-disabled]="disabled"
-			(click)="toggleMenu()"
 			(blur)="onBlur()"
-			[disabled]="disabled">
-			<span class="bx--list-box__label">{{getDisplayValue() | async}}</span>
-			<div *ngIf="!skeleton" class="bx--list-box__menu-icon" [ngClass]="{'bx--list-box__menu-icon--open': !menuIsClosed }">
-				<svg fill-rule="evenodd" height="5" role="img" viewBox="0 0 10 5" width="10" [attr.aria-label]="menuButtonLabel">
-					<title>{{menuButtonLabel}}</title>
-					<path d="M0 0l5 4.998L10 0z"></path>
+			[disabled]="disabled"
```
cal-smith: This is super nit picky but:
```suggestion
			[disabled]="disabled">
```

src/dropdown/dropdown.component.ts
```diff
@@ -43,15 +43,28 @@ import { DropdownService } from "./dropdown.service";
...
+		>
```
cal-smith: and we can do this
```suggestion
```

src/dropdown/dropdown.component.ts
cal-smith: to be clear, the big change here is updating `return of(`${this.selectedLabel}`);` to be `return of(this.placeholder);` and updating the comment to be more descriptive :+1:

cal-smith:
Changing the base from `master` to `v9`, since these changes are more applicable to the old codebase. We definitely want the changes to `abstract-dropdown-view` in master though 👍 

You'll want to `git pull upstream v9` instead of `git pull upstream master` when you're making changes :+1:

PR #416, opened 2019-02-20
Dropdown enhancements

cal-smith:
includes:

- fix(combobox): use blur event to focus input
  - lets us cleanly move focus between the list and the input
- fix(dropdown-list): element iterators should operate on the displayItems
  - `getNextElement` and related methods now operate on the visiable `disaplayItems` instead of the entire list. Also moves to a `ViewChildren` query instead of manually querying the DOM (!!)
- feat(AbstractDropdownView): add blur event to view spec
  - adds `blur` as a requirement for the view. Implementing classes should specify what blurs they'll emit (top, bottom, end, start, whatever makes sense for the type of view)

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -264,7 +264,7 @@ export class ComboBox implements OnChanges, OnInit, AfterViewInit, AfterContentI
 				this.updateSelected();
 			});
 
-			this.view.blur.pipe(filter(v => v === "top")).subscribe(() => {
+			this.view.blurIntent.pipe(filter(v => v === "top")).subscribe(() => {
 				this.elementRef.nativeElement.querySelector(".bx--text-input").focus();
 			});
 		}
--- a/src/dropdown/abstract-dropdown-view.class.ts
+++ b/src/dropdown/abstract-dropdown-view.class.ts
@@ -27,7 +27,7 @@ export class AbstractDropdownView {
 	 * ArrowUp -> emit event
 	 *
 	 * It's recommended that the implementing view include a specific type union of possible blurs
-	 * ex. `@Output() blur = new EventEmitter<"top" | "bottom">();`
+	 * ex. `@Output() blurIntent = new EventEmitter<"top" | "bottom">();`
 	 */
 	@Output() blurIntent: EventEmitter<any>;
 	/**
--- a/src/dropdown/abstract-dropdown-view.class.ts
+++ b/src/dropdown/abstract-dropdown-view.class.ts
@@ -29,7 +29,7 @@ export class AbstractDropdownView {
 	 * It's recommended that the implementing view include a specific type union of possible blurs
 	 * ex. `@Output() blur = new EventEmitter<"top" | "bottom">();`
 	 */
-	@Output() blur: EventEmitter<any>;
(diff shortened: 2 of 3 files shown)


src/combobox/combobox.component.ts
```diff
@@ -294,7 +299,7 @@ export class ComboBox implements OnChanges, OnInit, AfterViewInit, AfterContentI
 		} else if ((ev.key === "ArrowUp" || ev.key === "Up") // `"Up"` is IE specific value
 			&& this.dropdownMenu.nativeElement.contains(ev.target)
 			&& !this.view.hasPrevElement()) {
-			this.elementRef.nativeElement.querySelector(".bx--text-input").focus();
+			// this.elementRef.nativeElement.querySelector(".bx--text-input").focus();
```
magedhennawy: Do we want to completely remove this?

cal-smith: yep, I think that was the intention ... oops 😬

src/dropdown/abstract-dropdown-view.class.ts
```diff
@@ -19,6 +19,17 @@ export class AbstractDropdownView {
 	 * Emits selection events to other class.
 	 */
 	@Output() select: EventEmitter<Object>;
+	/**
+	 * Event to suggest a blur on the view.
+	 * Emits _after_ the first/last item has been focussed
```
zvonimirfras:
```suggestion
	 * Emits _after_ the first/last item has been focused.
```

src/dropdown/list/dropdown-list.component.ts
```diff
@@ -116,6 +120,14 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 	 * Event to emit selection of a list item within the `DropdownList`.
 	 */
 	@Output() select: EventEmitter<Object> = new EventEmitter<Object>();
+	/**
+	 * Event to suggest a blur on the view.
+	 * Emits _after_ the first/last item has been focussed
```
zvonimirfras:
```suggestion
	 * Emits _after_ the first/last item has been focused.
```

zvonimirfras: Maybe also add that they should catch the event and focus something else (to actually blur the thing), if they are using list in their own dropdown component implementation. That's what it's for, no?

I'm not the happiest with the name because it's the same as https://developer.mozilla.org/en-US/docs/Web/Events/blur and it doesn't behave the same. `blurIntent`? I know you've thought about it more so if this is really the best choice, let's go with it.

src/dropdown/abstract-dropdown-view.class.ts
```diff
@@ -19,6 +19,17 @@ export class AbstractDropdownView {
...
 	@Output() select: EventEmitter<Object>;
+	/**
+	 * Event to suggest a blur on the view.
+	 * Emits _after_ the first/last item has been focused.
+	 * ex.
+	 * ArrowUp -> focus first item
+	 * ArrowUp -> emit event
+	 *
+	 * It's recommended that the implementing view include a specific type union of possible blurs
+	 * ex. `@Output() blur = new EventEmitter<"top" | "bottom">();`
```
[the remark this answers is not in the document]
cal-smith: Yep!

PR #421, opened 2019-02-22
fix(dropdown): add dropdown service to enable placeholder use

cal-smith:


src/dropdown/dropdown.service.ts
```diff
@@ -0,0 +1,92 @@
+import { Injectable, ElementRef } from "@angular/core";
+import { PlaceholderService } from "./../placeholder/placeholder.module";
+import { fromEvent, Subscription } from "rxjs";
+import { throttleTime } from "rxjs/operators";
+import position from "./../utils/position";
+
+@Injectable()
+export class DropdownService {
+
```
zvonimirfras:
```suggestion
```

PR #430, opened 2019-03-01
fix(overflow-menu): stop updating items outside of the angular change detection

cal-smith:


src/dialog/overflow-menu/overflow-menu-option.component.ts
```diff
@@ -61,29 +62,18 @@ export class OverflowMenuOption {
 
 	public tabIndex = -1;
 
+	public title = null;
```
zvonimirfras: Maybe this can be a getter so it updates automatically like it used to?

cal-smith: Nope, that was also causing some change after check errors ... the component instance shouldn't be reused in this case anyway, and if it is I'd expect `AfterViewInit` to be called again

zvonimirfras: Cool. Do you mind adding that explanation as a comment?

cal-smith: Yep 👍

PR #438, opened 2019-03-11
feat(tag): add tag component

cal-smith:
closes the main part of #84


.storybook/experimental.component.ts
```diff
@@ -1,13 +1,28 @@
...
+	EventEmitter
+} from "@angular/core";
 import { ExperimentalService } from "../src/experimental.module";
 
 @Component({
 	selector: "app-experimental-component",
 	template: ``
 })
 export class ExperimentalComponenent implements AfterViewInit, OnDestroy {
+
```
zvonimirfras:
```suggestion
```

src/tag/tag.component.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Supported tags types for carbon v10
+ */
+export type TagTypeExperimental = "basic" | "red" | "magenta" | "purple" | "blue" | "cyan" | "teal" | "green" | "cool-gray" | "warm-gray";
+
+/**
+ * Component that represents a tag for labelling/categorizing using keywords
+ */
+@Component({
+	selector: "ibm-tag",
+	template: `<span><ng-content></ng-content></span>`
```
zvonimirfras: Does it have to be in a span?

src/tag/tag.component.ts
```diff
@@ -0,0 +1,36 @@
...
+export class Tag {
+	/**
+	 * type of the tag determines the styling
+	 *
+	 * Reference `TagType` for v9 applications, and `TagTypeExperimental` for v10/v9 experimental mode applications
+	 */
+	@Input()
+	type: TagType | TagTypeExperimental = "ibm";
+
+	@HostBinding("attr.class") get attrClass() {
```
zvonimirfras: Should we allow setting of additional classes?

src/tag/tag.component.ts
cal-smith: nope, since it's applying to the host

src/tag/tag.component.ts
cal-smith: Why not

PR #441, opened 2019-03-12
 feat(progress-indicator): get/set current progress

youda97:


src/progress-indicator/progress-indicator.stories.ts
```diff
@@ -45,75 +45,71 @@ storiesOf("ProgressIndicator", module)
...
+			],
+			current: number("Current progress", 1)
 		}
+	}))
+	.add("Skeleton", () => ({
+		template: `
+		<app-experimental-component></app-experimental-component>
+		<app-skeleton-progress-indicator></app-skeleton-progress-indicator>
+		&nbsp;&nbsp;
+		<app-skeleton-progress-indicator [orientation]="'vertical'"></app-skeleton-progress-indicator>
```
zvonimirfras: Would 
```suggestion
		<app-skeleton-progress-indicator orientation="vertical"></app-skeleton-progress-indicator>
```
work?

src/progress-indicator/progress-indicator.component.ts
```diff
@@ -63,6 +63,29 @@ export class ProgressIndicator {
...
+		if (current === undefined) {
+			for (let i = 0; i < this.steps.length; i++) {
+				this.steps[i].state[0] = "incomplete";
+			}
+			return;
+		}
+
+		if (current > this.steps.length - 1) {
+			current = this.steps.length - 1;
+		}
```
zvonimirfras: What if `current` is < 0?

We can't skip steps and complete them later? That was a requirement at some point.

PR #442, opened 2019-03-12
feat(carbon-icon): adding carbon icon component

DusanMilko:
Adding support for carbon icons.

Pulling in carbon-icons npm package to create carbon icons


cal-smith:
Thanks for putting in the effort to do this! However in the future, it would be great if you could open an issue before beginning work on a major feature like this - specifically in this case because teams should be looking to use the [`@carbon/icons-angular`](https://github.com/IBM/carbon-elements/tree/master/packages/icons-angular) package for icons going forward.

That package is part of the [carbon-elements](https://github.com/ibm/carbon-elements) work, which is an effort to codify the core IBM Design Language elements (iconography, color, type, grid, and more).

Feel free to reach out here or on slack if you have any questions.

PR #443, opened 2019-03-13
feat(tiles): Add selectable and expandable tiles

youda97:


src/tiles/expandable-tile.component.ts
```diff
@@ -0,0 +1,108 @@
...
+
+@Component({
+	selector: "ibm-below-the-fold",
+	template: `
+		<span class="bx--tile-content__below-the-fold">
+			<ng-content></ng-content>
+		</span>
+	`
+})
+export class BelowTheFold { }
```
cal-smith: Both AboveTheFold and BelowTheFold should be directives ... split out into a separate file (they can both live in the same file however)

src/tiles/tile-group.component.ts
```diff
@@ -0,0 +1,95 @@
...
+export class TileGroup implements AfterContentInit {
+	static tileGroupCount = 0;
+	/**
+	 * The tile group `name`
+	 */
+	@Input() name = `tile-group-${TileGroup.tileGroupCount}`;
+	/**
+	 * Set to `"multi"` to support multiple tile selection
+	 */
+	@Input() type: "single" | "multi" = "single";
```
cal-smith: I'd like to start moving everything to `multiple` instead of `type: "single" | "multi"` for multiple selections ... primarily to match existing html semantics

src/tiles/tile-group.component.ts
```diff
@@ -0,0 +1,95 @@
...
+	 * Emits an object that looks like:
+	 * ```javascript
+	 * {
+	 * 	value: "something",
+	 * 	selected: true,
+	 * 	name: "tile-group-1"
+	 * }
+	 * ```
+	 */
+	@Output() selected: EventEmitter<{value: string, selected: boolean, name: string}> = new EventEmitter();
```
cal-smith: Can we encode this type explicitly? Something like `TileSelection`?

src/tiles/expandable-tile.component.ts
```diff
@@ -0,0 +1,88 @@
...
+	 * {
+	 *		"EXPAND": "Expand",
+	 *		"COLLAPSE": "Collapse",
+	 * }
+	 * ```
+	 */
+	@Input()
+	set translations (value) {
+		if (value.EXPAND) {
+			this.expand = new BehaviorSubject(value.EXPAND);
```
cal-smith: I think this needs to be 
```suggestion
			this.expand.next(value.EXPAND);
```

src/tiles/expandable-tile.component.ts
```diff
@@ -0,0 +1,88 @@
...
+		}
+		if (value.COLLAPSE) {
+			this.collapse = new BehaviorSubject(value.COLLAPSE);
```
cal-smith: same here
```suggestion
			this.collapse.next(value.COLLAPSE);
```

otherwise i18n can't update the values

src/tiles/tile-group.component.ts
```diff
@@ -0,0 +1,95 @@
...
+	 * Emits an object that looks like:
+	 * ```javascript
+	 * {
+	 * 	value: "something",
+	 * 	selected: true,
+	 * 	name: "tile-group-1"
+	 * }
+	 * ```
+	 */
+	@Output() tileSelection: EventEmitter<{value: string, selected: boolean, name: string}> = new EventEmitter();
```
cal-smith: Same comment as before, can we encode `{value: string, selected: boolean, name: string}` as an explicit type?

src/tiles/expandable-tile.component.ts
```diff
@@ -0,0 +1,87 @@
...
+
+	ngAfterContentInit() {
+		this.setMaxHeight();
+	}
+
+	get expandedHeight() {
+		return this.tileMaxHeight + parseInt(getComputedStyle(this.element.querySelector(".bx--tile")).paddingBottom, 10);
+	}
+
+	setMaxHeight() {
```
zvonimirfras: Maybe
```suggestion
	updateMaxHeight() {
```
since you're not really manually setting it?

src/tiles/selection-tile.component.ts
```diff
@@ -0,0 +1,90 @@
...
+	@Input() value: string;
+	/**
+	 * Internal event used to notify the containing `TileGroup` of changes.
+	 */
+	@Output() change: EventEmitter<Event> = new EventEmitter();
+
+	/**
+	 * Set by the containing `TileGroup`. Used for the `name` property on the input.
+	 */
+	name = "tile";
```
zvonimirfras: Why initialize it if group sets it?
And since we're initializing, why not to a unique value?

PR #445, opened 2019-03-18
fix(table): make the checkbox label configurable

cal-smith:
Closes IBM/carbon-components-angular#310

- changes the en.json `CHECKBOX_ROW` to `"Select {{value}}"`
- adds `ROW` to the `TABLE` section of en.json (default value for the checkbox)
- adds `selectionLabelColumn` input to table (+ docs)
  - defaults to `Select row` (or the equivalent translation). if `selectionLabelColumn` is specified the table will use the data value from the specified column

fyi @stanislavgeorgiev


stanislavgeorgiev:
Awesome, thanks @cal-smith !

PR #447, opened 2019-03-19
Set overflow menu option title to "" instead of null

SaschaSchwarze0:
Overflow menu option component has a default title value of null which causes null to be shown as title in the browser. Changing this to empty string to get no title which is intended there.


SaschaSchwarze0:
> 
> 
> Thanks for this!
> 
> It may be better to remove the `title` attribute entirely if it doesn't exist ... I think `[attr.title]` instead of just `[title]` on the button will do this.

Done. @cal-smith

PR #461, opened 2019-04-03
feat(table): Add table header and update table toolbar

### LINKED ISSUE #531: Add translation for `Cancel` button in `TableToolbar`
`table-toolabar.component.ts` line 20

also for `count items selected`

### LINKED ISSUE #532: Update `TableToolbar` layout
Table toolbar layout changed and is now different from what we have. Update it according to the toolbar at https://www.carbondesignsystem.com/components/data-table/code
In particular when items are selected.

youda97:
closes #532 
closes #531


src/table/table.component.ts
```diff
@@ -168,244 +168,249 @@ import { I18n } from "./../i18n/i18n.module";
...
-							(dragleave)="columnDragLeave($event, 'right', i)"
-							(dragover)="columnDragover($event, 'right', i)"
-							(drop)="columnDrop($event, 'right', i)">
-							</div>
-							<div
-							*ngIf="columnDraggedHoverIndex == i && columnDraggedPosition == 'right'"
-							class="drop-indicator-right"></div>
-						</div>
+	<div class="bx--data-table-v2-container">
+		<h4 class="bx--data-table-v2-header">{{title}}</h4>
```
cal-smith: It's not clear to me that the title _needs_ to be baked into the table ... what does this get us?

youda97: The whole table should be wrapped with the class `bx--data-table-v2-container` and since the title and toolbar are part of the table I thought it made sense to make it part of the component. Not sure if that's what we want to do though

cal-smith: If none of the core table classes depend on being within `bx--data-table-v2-container` I think this would be better as another component ... that way consumers aren't locked into one way of using tables. We can even provide a `tableHeader` directive that applies the `bx--data-table-v2-header` class - that way consumers can totally customize their headers.

src/table/header/table-header.component.ts
```diff
@@ -0,0 +1,13 @@
+import { Component, HostBinding } from "@angular/core";
+
+@Component({
+	selector: "ibm-table-header",
+	template: `
+		<ng-content selector="ibmTableHeaderTitle"></ng-content>
+		<ng-content selector="ibmTableHeaderDescription"></ng-content>
```
cal-smith: This can just be one `ng-content` with no selector since we're not doing anything special with ordering 👍

src/table/table.stories.ts
```diff
@@ -34,17 +35,85 @@ import { clone } from "../utils/utils";
 	selector: "app-table",
 	template: `
 		<ibm-table
+			[title]="title"
```
cal-smith: Looks like you have some leftover `title` inputs

src/table/table.stories.ts
```diff
@@ -97,10 +162,15 @@ class TableStory implements OnInit, OnChanges {
 			[striped]="striped"
 			[isDataGrid]="isDataGrid"
 			(sort)="customSort($event)">
+			<ibm-table-toolbar [model]="model">
```
cal-smith: I think the toolbar needs to move outside of the table ... I'm not seeing it rendered in the story 🤔

youda97: Hmm, that's weird... it's working for me 🤔

src/table/table.stories.ts
```diff
@@ -80,8 +80,10 @@ class TableStory implements OnInit, OnChanges {
 @Component({
 	selector: "app-custom-table",
 	template: `
-		<button class="bx--btn bx--btn--sm bx--btn--primary" (click)="addRow()">Add row</button>
-		<button class="bx--btn bx--btn--sm bx--btn--primary" (click)="addColumn()">Add column</button>
+		<ibm-table-toolbar [model]="model">
+			<button class="bx--btn bx--btn--sm bx--btn--primary" (click)="addRow()">Add row</button>
```
zvonimirfras: Why not use `ibmButton` here?

src/table/table.stories.ts
```diff
@@ -472,20 +474,28 @@ storiesOf("Table", module).addDecorator(
...
-				[sortable]="sortable"
-				[isDataGrid]="isDataGrid">
-			</app-table>
-		</div>
+		<ibm-table-container>
+			<ibm-table-header>
+				<h4 ibmTableHeaderTitle>{{title}}</h4>
+				<p ibmTableHeaderDescription>{{description}}</p>
+			</ibm-table-header>
+			<div style="width: 650px">
```
zvonimirfras: Is the `div` necessary? Why not apply styles to table directly?

src/table/table.stories.ts
```diff
@@ -514,90 +530,100 @@ storiesOf("Table", module).addDecorator(
...
+					<button ibmButton="toolbar-action">
+						<svg class="bx--toolbar-action__icon" fill-rule="evenodd" height="16" name="download"
+							role="img" viewBox="0 0 14 16" width="14" aria-label="Download">
+							<title>Download</title>
+							<path d="M7.506 11.03l4.137-4.376.727.687-5.363 5.672-5.367-5.67.726-.687 4.14 4.374V0h1v11.03z"></path>
+							<path d="M13 15v-2h1v2a1 1 0 0 1-1 1H1a1 1 0 0 1-1-1v-2h1v2h12z"></path>
+						</svg>
+					</button>
+					<button ibmButton="toolbar-action">
+						<svg class="bx--toolbar-action__icon" fill-rule="evenodd" height="16" name="edit"
```
zvonimirfras: Is this replaceable with an icon component?

src/table/table.stories.ts
```diff
@@ -514,90 +530,100 @@ storiesOf("Table", module).addDecorator(
...
+							role="img" viewBox="0 0 16 16" width="16" aria-label="Edit">
+							<title>Edit</title>
+							<path d="M7.926 3.38L1.002 9.72V12h2.304l6.926-6.316L7.926 3.38zm.738-.675l2.308 2.304
+								1.451-1.324-2.308-2.309-1.451 1.329zM.002 9.28L9.439.639a1 1 0 0 1 1.383.03l2.309 2.309a1
+								1 0 0 1-.034 1.446L3.694 13H.002V9.28zM0 16.013v-1h16v1z"></path>
+						</svg>
+					</button>
+					<button ibmButton="toolbar-action">
+						<svg class="bx--toolbar-action__icon" fill-rule="evenodd" height="16" name="settings"
```
zvonimirfras: Is this replaceable with an icon component?

src/table/header/table-header.component.ts
```diff
@@ -0,0 +1,12 @@
...
+
+@Component({
+	selector: "ibm-table-header",
+	template: `
+		<ng-content></ng-content>
+	`
+})
+export class TableHeader {
+	@HostBinding("class.bx--data-table-header") headerClass = true;
+	@HostBinding("style.display") headerStyle = "block";
```
zvonimirfras:
```suggestion
	@HostBinding("style.display") displayStyle = "block";
```

src/table/table.stories.ts
youda97: Ya it doesn't work otherwise

zvonimirfras: Why doesn't it work? What does it do instead?

youda97: Well since its only two columns so the table shrinks down

youda97:
Yup seems like everything is working fine

src/table/toolbar/table-toolbar.component.ts
```diff
@@ -31,7 +32,11 @@ export class TableToolbar {
 	@Input() set ariaLabel (value) {
 		this.actionBarLabel.next(value.ACTION_BAR);
 	}
+	@Input() set cancel (value) {
+		this.cancelText.next(value.CANCEL);
+	}
 	actionBarLabel = this.i18n.get("TABLE_TOOLBAR.ACTION_BAR");
+	cancelText = this.i18n.get("TABLE_TOOLBAR.CANCEL");
```
cal-smith: This can be `_cancelText` and returned by a getter

src/table/toolbar/table-toolbar.component.ts
```diff
@@ -31,7 +32,11 @@ export class TableToolbar {
 	@Input() set ariaLabel (value) {
 		this.actionBarLabel.next(value.ACTION_BAR);
 	}
+	@Input() set cancel (value) {
```
cal-smith: This should be `cancelText` ... I think we're generally standardizing on `xText` for translations that appear in the UI and `xLabel` for accessible translations/text.

src/table/toolbar/table-toolbar.component.ts
```diff
@@ -42,7 +47,7 @@ export class TableToolbar {
 		return this.model.totalDataLength > 0 ? this.model.rowsSelected.some(item => item) : false;
 	}
 
-	cancel() {
+	onCancel() {
```
cal-smith: This is good though 🙌

src/table/table.stories.ts
zvonimirfras: What if you add width to the table? Might need to also add a display value with it.

cal-smith: Just confirming that the `width: 650px` is needed because of the centering styles in the storybook?

src/search/search.component.html
```diff
@@ -0,0 +1,41 @@
...
+	[ngClass]="{
+		'bx--search--sm': size === 'sm',
+		'bx--search--xl': size === 'xl',
+		'bx--search--light': theme === 'light',
+		'bx--skeleton': skeleton,
+		'bx--toolbar-search': toolbar,
+		'bx--toolbar-search--active': toolbar && active
+	}"
+	role="search">
+	<label class="bx--label" [for]="id">{{label}}</label>
```
zvonimirfras: Also support templates?

src/table/toolbar/table-toolbar-search.component.ts
```diff
@@ -1,9 +1,23 @@
...
 })
-export class TableToolbarSearch {
-	@HostBinding("class.bx--toolbar-search-container") class = true;
+export class TableToolbarSearch extends Search implements OnInit {
+	@HostBinding("class.bx--toolbar-search-container-expandable") @Input() expandable = false;
+
+	@HostBinding("class.bx--toolbar-search-container-persistent") get persistentClass() { return !this.expandable; }
+
+	@HostBinding("class.bx--toolbar-search-container-active") get activeClass() {
+		return this.active && this.value !== null || this.value !== "";
```
zvonimirfras: Did you mean
```suggestion
		return this.active && (this.value !== null || this.value !== "");
```
?

src/table/toolbar/table-toolbar-search.component.ts
```diff
@@ -1,9 +1,23 @@
-import { Component, HostBinding } from "@angular/core";
+import { Search } from "./../../search/search.component";
+import { Component, HostBinding, Input, OnInit } from "@angular/core";
```
zvonimirfras:
```suggestion
import {
	Component,
	HostBinding,
	Input,
	OnInit
} from "@angular/core";
```

src/search/search.component.html
youda97: Don't think this is an actual label? Looking at [`carbon-design system`](https://www.carbondesignsystem.com/components/search/code) I don't see any label

src/table/toolbar/table-toolbar-search.component.ts
```diff
@@ -1,9 +1,30 @@
...
-	@HostBinding("class.bx--toolbar-search-container") class = true;
+export class TableToolbarSearch extends Search implements OnInit {
+	tableSearch = true;
+
+	@HostBinding("class.bx--toolbar-search-container-expandable") @Input() expandable = false;
+
+	@HostBinding("class.bx--toolbar-search-container-persistent") get persistentClass() { return !this.expandable; }
+
+	@HostBinding("class.bx--toolbar-search-container-active") get activeClass() {
+		return this.active && (this.value !== null || this.value) !== "";
```
zvonimirfras:
```suggestion
		return this.active && (this.value !== null || this.value !== "");
```
?

PR #463, opened 2019-04-03
fix(modal): Bring new modals on top of the old ones.

stanislavgeorgiev:
Closes IBM/carbon-components-angular#462

I see that the `DialogService` is also using `placeholderService.createComponent()` so there may be unintended consequences from my change. I'm not sure how and what to test so please bear this in mind when doing to review.

### DIFF
--- a/src/placeholder/placeholder.service.ts
+++ b/src/placeholder/placeholder.service.ts
@@ -30,7 +30,7 @@ export class PlaceholderService {
 		if (!this.viewContainerRef) {
 			console.error("No view container defined! Likely due to a missing `ibm-placeholder`");
 		}
-		return this.viewContainerRef.createComponent(componentFactory, 0, injector);
+		return this.viewContainerRef.createComponent(componentFactory, undefined, injector);
 	}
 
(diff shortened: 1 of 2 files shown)


src/placeholder/placeholder.service.ts
```diff
@@ -30,7 +30,7 @@ export class PlaceholderService {
 		if (!this.viewContainerRef) {
 			console.error("No view container defined! Likely due to a missing `ibm-placeholder`");
 		}
-		return this.viewContainerRef.createComponent(componentFactory, 0, injector);
+		return this.viewContainerRef.createComponent(componentFactory, undefined, injector);
```
cal-smith: I think we generally use `null` for skipping arguments ... unless `undefined` explicitly works best here?
```suggestion
		return this.viewContainerRef.createComponent(componentFactory, null, injector);
```

stanislavgeorgiev: oh sure @cal-smith, that makes sense to make it consistent with your practices

PR #466, opened 2019-04-04
fix(datepicker): fixes issue#449

### LINKED ISSUE #449: DatePicker position within modal Dialog
## Detailed description

I found a bug for this component. As you can see below, when I use it on Modal Component the position is changed.
image
Following some code lines that may be cause the bug:
image
Please, let me know if the bug description was clear for you.
Thanks in advance and excuse me if I did not write according to the established standards.

cal-smith:
closes #449 

- ensure it has the correct classes when it opens
- ensure it has the correct position
- clean up as much as we can when our element is destroyed

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -1,6 +1,14 @@
-import { Component, Input, Output, EventEmitter } from "@angular/core";
+import {
+	Component,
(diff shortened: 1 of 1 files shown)


magedhennawy:
Thanks for fixing it! 👍 LGTM

PR #469, opened 2019-04-04
 feat(tooltip): Add definition and icon tooltip

youda97:


src/dialog/tooltip/tooltip-definition.component.ts
```diff
@@ -0,0 +1,43 @@
+import { Component, Input, HostBinding } from "@angular/core";
+
+@Component({
+	selector: "ibm-tooltip-definition",
+	template: `
+		<button class="bx--tooltip__trigger" [attr.aria-describedby]="id">{{triggerText}}</button>
```
cal-smith: It may be cleaner if `triggerText` was an `ng-content` instead

src/dialog/tooltip/tooltip-definition.component.ts
```diff
@@ -0,0 +1,41 @@
...
+	/**
+	 * The placement in which the `TooltipDefinition` appears.
+	 * Set to `"top"` to have it positioned above the trigger text
+	 */
+	@Input() placement: "bottom" | "top" = "bottom";
+
+	@HostBinding("class.bx--tooltip--definition") className = true;
+
+	constructor() {
+		TooltipDefinition.tooltipItemCount++;
```
cal-smith: this could be part of the `id` generation

PR #481, opened 2019-04-08
Add features to datepicker

youda97:


src/datepicker/datepicker.component.ts
```diff
@@ -141,6 +164,9 @@ export class DatePicker implements OnDestroy {
 		// add day classes and special case the "today" element based on `this.value`
 		Array.from(dayContainer).forEach(element => {
 			element.classList.add("bx--date-picker__day");
+			if (!this.value) {
```
cal-smith: Just curious what case this is handling?

youda97: When there is no assigned value I get a `TypeError: Cannot read property 'length' of undefined`

cal-smith: Ah, okay ... maybe a short comment noting that?

youda97: Well it's a simple function guard so not really sure what to add as a comment

PR #486, opened 2019-04-10
feat(overflow-menu): Add overflow with links

youda97:


src/dialog/overflow-menu/overflow-menu-option.component.ts
```diff
@@ -72,7 +93,7 @@ export class OverflowMenuOption implements AfterViewInit {
 	}
 
 	ngAfterViewInit() {
-		const button = this.elementRef.nativeElement.querySelector("button");
+		const button = this.elementRef.nativeElement.querySelector("button , a");
 		if (button.scrollWidth > button.offsetWidth) {
 			this.title = this.elementRef.nativeElement.querySelector("button").textContent;
```
cal-smith: I'm not sure why we ended up with two of the same `querySelectors` ... this should just be 
```suggestion
			this.title = button.textContent;
```

Not your fault, just a weird artifact of something 🤔

src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -119,6 +119,7 @@ export class OverflowMenuPane extends Dialog implements AfterViewInit {
 	}
 
 	protected listItems() {
-		return Array.from<any>(this.elementRef.nativeElement.querySelectorAll(".bx--overflow-menu-options__btn:not([disabled])"));
+		const selector = ".bx--overflow-menu-options__btn:not(.bx--overflow-menu-options__option--disable)";
+		return Array.from<any>(this.elementRef.nativeElement.querySelectorAll(selector));
```
cal-smith: This would be `Array.from<Element>` since we know what type `querySelectorAll` returns ... if anything we may not even need to cast to `<any>` in the `.from`

src/dialog/overflow-menu/overflow-menu-option.component.ts
```diff
@@ -32,8 +33,26 @@ import {
 			(click)="onClick($event)"
 			[disabled]="disabled"
 			[attr.title]="title">
-			<ng-content></ng-content>
+			<ng-container *ngTemplateOutlet="tempOutlet"></ng-container>
 		</button>
+
+		<a
+			*ngIf="href !== undefined"
```
cal-smith: could we just do a regular `*ngIf="href"` and `*ngIf="!href"` instead of explicitly checking against undefined? We should catch empty strings and other falsey values with the implicit check

src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -119,6 +119,9 @@ export class OverflowMenuPane extends Dialog implements AfterViewInit {
 	}
 
 	protected listItems() {
-		return Array.from<any>(this.elementRef.nativeElement.querySelectorAll(".bx--overflow-menu-options__btn:not([disabled])"));
+		const list = [];
+		const selector = ".bx--overflow-menu-options__option:not([disabled])";
+		this.elementRef.nativeElement.querySelectorAll(selector).forEach(item => list.push(item.firstElementChild));
```
cal-smith: What was wrong with `Array.from(this.elementRef.nativeElement.querySelectorAll(selector));`?

src/dialog/overflow-menu/overflow-menu-option.component.ts
```diff
@@ -32,8 +33,26 @@ import {
 			(click)="onClick($event)"
 			[disabled]="disabled"
 			[attr.title]="title">
-			<ng-content></ng-content>
+			<ng-container *ngTemplateOutlet="tempOutlet"></ng-container>
```
cal-smith: Why not `ng-content`?

src/dialog/overflow-menu/overflow-menu-pane.component.ts
youda97: the disabled option was still part of the disabled listItems so I had to do it in a different way

src/dialog/overflow-menu/overflow-menu-option.component.ts
youda97: Couldn't have two ng-content in the template

src/dialog/overflow-menu/overflow-menu-pane.component.ts
cal-smith: There's nothing your doing here that isn't the exact same as doing `Array.from(this.elementRef.nativeElement.querySelectorAll(selector));` ... so I'm still not sure why this changed

src/dialog/overflow-menu/overflow-menu-option.component.ts
cal-smith: Is there a specific error? Cause I think you can ... You should be able to, given that we're `ngIf`ing out a whole `ng-content`

If it's a compiler issue then this is fine :+1:

src/dialog/overflow-menu/overflow-menu-pane.component.ts
youda97: im getting the firstElementChild of each Node in the querySelectorAll thats why

cal-smith: Hm, so really you want `const selector = ".bx--overflow-menu-options__option:not([disabled]) .bx--overflow-menu-options__btn"`

src/dialog/overflow-menu/overflow-menu-option.component.ts
youda97: There's no error but the basic overflow completely breaks and you can't see any of the options

cal-smith: Gotcha, cool we can leave this then :+1:

PR #487, opened 2019-04-11
Add active/inactive states in loading and inline-loading components

youda97:


src/inline-loading/inline-loading.stories.ts
```diff
@@ -19,7 +19,7 @@ storiesOf("Inline Loading", module)
 		`,
 		props: {
 			onSuccess: action("onSuccess"),
-			loadingText: "Loading data...",
-			successText: "Data loaded."
+			loadingText: text("The loading text", "Loading data..."),
+			successText: text("The sucess text", "Data loaded.")
```
zvonimirfras:
```suggestion
			successText: text("The success text", "Data loaded.")
```

PR #488, opened 2019-04-11
 feat(select): Add invalid and optional text fields

youda97:


src/select/select.component.ts
```diff
@@ -49,9 +50,16 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from "@angular/forms";
 				<svg *ngIf="!skeleton" class="bx--select__arrow" width="10" height="5" viewBox="0 0 10 5">
 				<path d="M0 0l5 4.998L10 0z" fill-rule="evenodd" />
 				</svg>
+				<div *ngIf="helperText" class="bx--form__helper-text">{{helperText}}</div>
+				<div *ngIf="invalid" class="bx--form-requirement">{{invalidText}}</div>
 			</div>
 		</div>
 	`,
+	styles: [`
+		[data-invalid] ~ .bx--select__arrow {
```
zvonimirfras: This is not available in carbon-components?

src/select/select.stories.ts
```diff
@@ -1,6 +1,6 @@
 import { storiesOf, moduleMetadata } from "@storybook/angular";
 import { action } from "@storybook/addon-actions";
-import { withKnobs, select } from "@storybook/addon-knobs/angular";
+import { withKnobs, select, boolean, text } from "@storybook/addon-knobs/angular";
```
zvonimirfras:
```suggestion
import {
	withKnobs,
	select,
	boolean,
	text
} from "@storybook/addon-knobs/angular";
```

src/select/select.component.ts
youda97: Nope... react had to do something similar as well

cal-smith: This may be fixed in 10 ... I think at this point we're going to be stuck with it ... if it's not fixed in 10 then we should make an issue against carbon-components.

PR #494, opened 2019-04-12
fix(toggle): Add label and update knobs

youda97:


src/toggle/toggle.component.ts
```diff
@@ -140,27 +137,28 @@ export class Toggle extends Checkbox {
...
 	 */
 	@Input() skeleton = false;
 
 	/**
 	 * The unique id allocated to the `Toggle`.
 	 * @type {string}
-	 * @memberof Toggle
 	 */
 	id = "toggle-" + Toggle.toggleCount;
+	labelId = "toggle-" + Toggle.toggleCount + "-label";
```
cal-smith: `Checkbox` already provides a `ariaLabelledby` and `ariaLabel` input - we should use/override those instead. That way consumers can create an arbitrarily complex label and pass the id to the toggle, or they can create a simple label with the `label` input

src/toggle/toggle.component.ts
```diff
@@ -140,27 +137,28 @@ export class Toggle extends Checkbox {
 	get onText() {
 		return this._onText;
 	}
-
-
+	/**
+	 * Text that is set as the label of the toggle.
+	 * @type {(string)}
+	 */
+	@Input() labelText: string;
```
cal-smith: This should probably just be
```suggestion
	@Input() label: string;
```

PR #508, opened 2019-04-22
chore(position): switch to @carbon/utils-position

cal-smith:


cal-smith:
My branches and upstreams got confused, so this and #510 got swapped 🤦‍♂️

PR #522, opened 2019-05-02
Support render template in dropdown component

rudolpho1:
Extend the dropdown component to support rendering the current value via an angular template. This is similar to the template support the dropdown-list component already has.

#### Changelog

**New**

* new input parameter called *renderTpl* in dropdown component

### DIFF
--- a/src/pagination/pagination.component.ts
+++ b/src/pagination/pagination.component.ts
@@ -58,16 +58,10 @@ import { ExperimentalService } from "./../experimental.module";
 					[ngClass]="{
 						'bx--select__item-count': isExperimental
 					}">
-					<label
-						[for]="itemsPerPageSelectId"
-						class="bx--label bx--visually-hidden">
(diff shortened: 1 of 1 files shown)


src/dropdown/dropdown.component.ts
```diff
@@ -166,6 +172,10 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy {
 	 * Defaults to the `DROPDOWN.SELECTED` value from the i18n service.
 	 */
 	@Input() selectedLabel = this.i18n.get().DROPDOWN.SELECTED;
+	/**
+	 * Template to render the current value of the dropdown.
+	 */
+	@Input() renderTpl: string | TemplateRef<any> = null;
```
cal-smith: We should be able to re-use `displayValue` instead of introducing another Input ... we'll just need to check if `displayValue` contains a template or not and toggle the template appropriately

PR #523, opened 2019-05-02
Fix content-switcher button being unselected

jdcai:
Closes IBM/carbon-components-angular#450

Fix for "if I click on one of the content switcher buttons and then click on a button separate from the content switcher, the content switcher button I selected becomes unselected" part of IBM/carbon-components-angular#450 issue

### DIFF
--- a/src/content-switcher/content-switcher-option.directive.ts
+++ b/src/content-switcher/content-switcher-option.directive.ts
@@ -58,7 +58,7 @@ export class ContentSwitcherOption {
 
 	@HostListener("blur", ["$event"])
 	onBlur(event) {
-		if (event.relatedTarget && event.relatedTarget.className === "bx--content-switcher-btn") {
+		if (event.relatedTarget && event.relatedTarget.classList.contains("bx--content-switcher-btn")) {
 			this.active = false;
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/content-switcher/content-switcher-option.directive.ts
```diff
@@ -58,7 +58,7 @@ export class ContentSwitcherOption {
 
 	@HostListener("blur", ["$event"])
 	onBlur(event) {
-		if (event.relatedTarget) {
+		if (event.relatedTarget && event.relatedTarget.className === "bx--content-switcher-btn") {
```
cal-smith: Since this is a directive, we should prefer `classList.contains` ... I don't expect it will be needed, but we _should_ support cases where we have additional classes applied.

```suggestion
		if (event.relatedTarget && event.relatedTarget.classList.contains("bx--content-switcher-btn")) {
```

jdcai:
The pull request is for v9 but it should also go into master
Is there an easy way to do that on your side or does it need another pull request?

cal-smith:
The best way would be another pull request :+1: v9 and master have diverged quite a bit so far, and merging either into the other would be risky 😞

cal-smith:
Sure :+1: I thought I set the commit message properly, but I guess semantic-release didn't pick it up

Issue #497, opened 2019-04-15
Evaluate form control inputs and composition


cal-smith:
We should take inventory of all our form controls structure and inputs, and make sure they're as standardized as possible. 

Generally we should follow standard HTML conventions, and deviate where needed (for styling or usability reasons).

We should support simple built in labels, helper text, and error states, as well as fully customized variants. This may require adding an extra wrapper component and removing `bx--form-item` from each form control - or something completely different (discussion open)

More to come ...

cal-smith:
It seems like we've settled on an approach via #524 and #525 - next steps are just making sure we're consistent with all our form components

PR #524, opened 2019-05-03
Add invalid and disabled to form components

youda97:
Sort of touches on #497


src/input/input.stories.ts
```diff
@@ -12,26 +12,46 @@ storiesOf("Input", module).addDecorator(
...
+			invalidText: text("Form validation content", "Validation message here"),
+			label: text("Label", "Text Input label"),
+			helperText: text("Helper text", "Optional helper text."),
+			placeholder: text("Placeholder text", "Placeholder text")
 		}
 	}))
 	.add("TextArea", () => ({
 		template: `
-		<textarea ibmTextArea [theme]="theme" aria-label="textarea" placeholder="Optional placeholder text" rows="4" cols="50"></textarea>
+		<ibm-label [helperText]="helperText" [invalid]="invalid" [invalidText]="invalidText">
```
cal-smith: These should probably be on newlines

cal-smith: Applies to all the other cases of long lines in this file too :+1:

src/select/select.component.ts
```diff
@@ -39,7 +39,7 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from "@angular/forms";
 				<label *ngIf="skeleton" [attr.for]="id" class="bx--label bx--skeleton"></label>
 				<label *ngIf="!skeleton" [attr.for]="id" class="bx--label">{{label}}</label>
 				<div *ngIf="helperText" class="bx--form__helper-text">{{helperText}}</div>
-				<div class="bx--select-input__wrapper">
+				<div class="bx--select-input__wrapper" [attr.data-invalid]="(invalid ? '' : null)">
```
zvonimirfras: ?
```suggestion
				<div class="bx--select-input__wrapper" [attr.data-invalid]="(invalid ? true : null)">
```

cal-smith:
Also, conflicts in `number.component` @youda97

src/input/input.stories.ts
```diff
@@ -12,26 +12,46 @@ storiesOf("Input", module).addDecorator(
...
+			{{label}}
+			<textarea
+				ibmTextArea
+				[placeholder]="placeholder"
+				[invalid]="invalid"
+				[disabled]="disabled"
+				[theme]="theme"
+				[rows]="rows"
+				[cols]="cols"
+				aria-label="textarea"></textarea>
```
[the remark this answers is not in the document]
youda97: For some reason, if I don't have the closing tag on the same line the placeholder is not displayed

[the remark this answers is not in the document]
cal-smith: Since `textarea` uses it's inner content as it's value it will treat the newline as content and not display the placeholder

src/pagination/pagination.component.ts
```diff
@@ -158,6 +165,18 @@ export class Pagination {
...
 	 */
 	@Input() model: PaginationModel;
+	/**
+ 	 * Set to `true` to disable the backward/forward buttons.
+	 */
+	@Input() disabled = false;
+	/**
+	 * Set to `true` to disable the select box that changes the page.
+	 */
+	@Input() pageInputDisabled = false;
```
cal-smith: I'm not sure I understand when this would be used ... `disabled` should disable everything ... this removes the select box, but also disabled the buttons. 

Is there a spec or an issue for this particular change?

youda97: In [Carbon design systems](https://www.carbondesignsystem.com/components/pagination/code) the `disabled` feature just disabled the buttons. So I checked [react](http://react.carbondesignsystem.com/?selectedKind=Pagination&selectedStory=Pagination&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) and that was the way they disabled their select

cal-smith: Right, but `pageInputDisabled` there just removes the select, it doesn't remove the select _and_ disable the buttons.

src/datepicker-input/datepicker-input.component.ts
```diff
@@ -40,9 +40,7 @@ import {
 					[disabled]="disabled"
 					[attr.data-invalid]="(invalid ? true : null)"
 					(change) = "valueChange.emit(dateInput.value)"/>
-					<div *ngIf="invalid" class="bx--form-requirement">
-						{{invalidText}}
-					</div>
+					<div *ngIf="invalid" class="bx--form-requirement" [innerHtml]="invalidText"></div>
```
cal-smith: I'm not sure about this change. Now we're encouraging passing around raw HTML strings, which is really un-ergonomic and doesn't allow for any Angular template features ... It would be better if it was plain text or a template (for custom situations)

src/input/input.directive.ts
```diff
@@ -19,9 +19,14 @@ export class TextInput {
 	 * `light` or `dark` input theme
 	 */
 	@Input() theme: "light" | "dark" = "dark";
+	@Input() disabled = false;
```
cal-smith: Do we need to add this? I _think_ Angular already lets us bind to the native `disabled` attribute

src/input/input.directive.ts
```diff
@@ -19,9 +19,14 @@ export class TextInput {
...
 
 	@HostBinding("class.bx--text-input") inputClass = true;
+	@HostBinding("class.bx--text-input--invalid") @Input() invalid = false;
 	@HostBinding("class.bx--skeleton") @Input() skeleton = false;
+	@HostBinding("attr.disabled") get disabledValue() {
```
cal-smith: same as above

src/input/text-area.directive.ts
```diff
@@ -19,9 +19,14 @@ export class TextArea {
 	 * `light` or `dark` input theme
 	 */
 	@Input() theme: "light" | "dark" = "dark";
+	@Input() disabled = false;
```
cal-smith: same note as the input directive

src/number-input/number.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, Input, HostBinding, EventEmitter, Output } from "@angular/core";
+import { Component, Input, HostBinding, EventEmitter, Output, TemplateRef } from "@angular/core";
```
cal-smith:
```suggestion
import {
	Component,
	Input,
	HostBinding,
	EventEmitter,
	Output,
	TemplateRef
} from "@angular/core";
```

PR #528, opened 2019-05-06
Fix content-switcher button being unselected bug

jdcai:
Closes IBM/carbon-components-angular#450

Copying fix from v9 to master


cal-smith:
Related: #523 #475

PR #536, opened 2019-05-08
feat(table): refactor table header

youda97:


src/table/table-header.component.ts
```diff
@@ -0,0 +1,234 @@
...
+
+	@Input() showSelectionColumn: boolean;
+
+	@Input() selectAllCheckboxSomeSelected: boolean;
+
+	@Input() selectAllCheckbox: boolean;
+
+	@Input() skeleton: boolean;
+
+	@Input() set columnIndex(value: number) {
```
cal-smith: Just a note, no action required: Since we're splitting this up further we'll want to look at centralizing this in some sort of model/service thing we can share in a given table instance

src/table/table.component.ts
```diff
@@ -694,20 +561,6 @@ export class Table implements AfterViewInit {
...
-	 * @param {TableModel} model
-	 */
-	@Output() selectAll = new EventEmitter<Object>();
-
-	/**
-	 * Emits if all rows are deselected.
-	 *
-	 * @param {TableModel} model
-	 */
-	@Output() deselectAll = new EventEmitter<Object>();
```
cal-smith: pretty sure we still need this, it should just be propagated from the header

src/table/table.component.ts
```diff
@@ -694,20 +561,6 @@ export class Table implements AfterViewInit {
 	 */
 	@Output() sort = new EventEmitter<number>();
 
-	/**
-	 * Emits if all rows are selected.
-	 *
-	 * @param {TableModel} model
-	 */
-	@Output() selectAll = new EventEmitter<Object>();
```
cal-smith: pretty sure we still need this, it should just be propagated from the header

src/table/table.component.spec.ts
```diff
@@ -69,25 +72,25 @@ describe("Table", () => {
...
+		spyOn(tableHeaderInstance.componentInstance.deselectAll, "emit");
 
 		let checkbox = fixture.nativeElement.querySelector("th input[type='checkbox']");
 		checkbox.click();
 		fixture.detectChanges();
 		checkbox.click();
 		fixture.detectChanges();
 
-		expect(tableInstance.componentInstance.deselectAll.emit).toHaveBeenCalled();
+		expect(tableHeaderInstance.componentInstance.deselectAll.emit).toHaveBeenCalled();
```
cal-smith: So this test should have just worked 😄 we'll need to add another set of tests for just the header 👍

src/table/table-header.component.ts
```diff
@@ -0,0 +1,244 @@
...
+	@Input() stickyHeader: boolean;
+
+	@Input() checkboxHeaderLabel;
+	@Input() sortDescendingLabel;
+	@Input() sortAscendingLabel;
+	@Input() filterTitle;
+
+	/**
+	 * Emits an index of the column that wants to be sorted.
+	 *
```
zvonimirfras:
```suggestion
```

src/table/table-header.component.ts
```diff
@@ -0,0 +1,244 @@
...
+
+	setCheckboxIndex() {
+		this.checkboxIndex.emit();
+	}
+
+	/**
+	 * Triggered whenever the header checkbox is clicked.
+	 * Updates all the checkboxes in the table view.
+	 * Emits the `selectAll` or `deselectAll` event.
+	 *
```
zvonimirfras:
```suggestion
```

src/table/table-head.component.ts
```diff
@@ -0,0 +1,242 @@
...
+	HostBinding,
+	Input,
+	Output,
+	EventEmitter
+} from "@angular/core";
+
+import { TableModel } from "./table.module";
+import { I18n } from "./../i18n/i18n.module";
+
+@Component({
```
zvonimirfras: Add component description that explains where and how it is used.
This is especially important here since we also have a variety of `table-header*` components and this will easily create confusion among users.

src/table/table-head.component.ts
```diff
@@ -0,0 +1,242 @@
...
+	@Input() sortDescendingLabel;
+	@Input() sortAscendingLabel;
+	@Input() filterTitle;
+
+	/**
+	 * Emits an index of the column that wants to be sorted.
+	 */
+	@Output() sort = new EventEmitter<number>();
+
+	@Output() index = new EventEmitter<number>();
```
zvonimirfras: Needs documentation, it's not really obvious when this emits.

src/table/table.component.ts
```diff
@@ -696,7 +565,7 @@ export class Table implements AfterViewInit {
 	 */
 	@Output() sort = new EventEmitter<number>();
 
-	/**
+		/**
```
zvonimirfras:
```suggestion
	/**
```

src/table/table.module.ts
```diff
@@ -1,47 +1,60 @@
...
 import { Close16Module } from "@carbon/icons-angular/lib/close/16";
 
-import { DialogModule } from "./../dialog/dialog.module";
 import { NFormsModule } from "./../forms/forms.module";
+import { DialogModule } from "./../dialog/dialog.module";
+import { I18nModule } from "./../i18n/i18n.module";
+
 import { ButtonModule, SearchModule } from "../";
 
+// imports
```
zvonimirfras:
```suggestion
// components
```
?

Since `// modules` are also imports

src/table/table-head.component.ts
```diff
@@ -0,0 +1,242 @@
...
+	Output,
+	EventEmitter
+} from "@angular/core";
+
+import { TableModel } from "./table.module";
+import { I18n } from "./../i18n/i18n.module";
+
+@Component({
+	// tslint:disable-next-line:component-selector
+	selector: "[ibm-table-head]",
```
cal-smith: Should be 
```suggestion
	selector: "[ibmTableHead]",
```

since we're following the directive pattern

cal-smith:
Another note ... we'll be deprecating the `index` and `indexChange` Input/Output almost immediately - the data grid interaction stuff can be tied together a lot better with a `DataGridModel` (or whatever we want to call it), probably passed in through a `dataGrid` input ... 

Does it make sense to skip exporting these components until we have our datagrid model situation locked down? That way we're not risking blowing up a public API surface

cal-smith:
Same goes for my table PR

zvonimirfras:
> Another note ... we'll be deprecating the `index` and `indexChange` Input/Output almost immediately - the data grid interaction stuff can be tied together a lot better with a `DataGridModel` (or whatever we want to call it), probably passed in through a `dataGrid` input ...
> 
> Does it make sense to skip exporting these components until we have our datagrid model situation locked down? That way we're not risking blowing up a public API surface

It makes sense to do this in a separate table-refactor branch and merge it to master once it's ready.

PR #537, opened 2019-05-08
 fix(date-picker): validate pattern

### LINKED ISSUE #482: Date Picker does not have a manual input validation for Date
## Detailed description
The Date Picker (range or not) currently does not count on an input validation when the date is entered manually.

When creating it in the NG template, the desired format and regex are given, so the date input already knows what is a valid format or not. Looks like the validation is only happening using the date picker in the calendar

#Carbon Version used
```
    "carbon-components": "9.91.0",
    "carbon-components-angular": "2.41.0",
```

#Example: Date filter in Assess UX Redesign
https://pages.github.ibm.com/talent-incubator/assess-ux-redesign/invitations/2

1. Go to [Assess UX Redesign Github Pages](https://pages.github.ibm.com/talent-incubator/assess-ux-redesign/invitations/2)
2. Open the Date filter on the top of the table.
3. Type whatever you want (something preferably different from the placeholder format)
4. Move the focus out of the input field (hit enter)

*Current behaviour*: 
- Anything can be entered. 
- No error given (we also had to disable this no no broken thing is entered).
*Desired behaviour*: 
-The input field shows an 'invalid format message' when focus is lost.

## Additional information

* GIF of the issue:
DateNoValid2
* Error on console (for NaN characters entered - for numbers no error is given)
image

### LINKED ISSUE #554: (v2) Date Picker: Ignores `dateFormat` property value
## Detailed description

We're trying to localize the Date Picker's picked date through setting the `dateFormat` property. We tried `Y-m-d` (general) and `d.m.Y` (German) but both gets ignored and the value still shows as `m/d/Y` (US)

Bug

Date Picker

The component should ideally react to the browser language. At least it should respect the property setting.

Chrome

v2@latest

IBM ADDI

youda97:
Closes #482 
closes #554


src/datepicker-input/datepicker-input.component.ts
```diff
@@ -71,7 +75,7 @@ export class DatePickerInput {
 
 	@Input() placeholder = "mm/dd/yyyy";
 
-	@Input() pattern = "\\d{1,2}/\\d{1,2}/\\d{4}";
+	@Input() pattern = new RegExp("^\\d{1,2}/\\d{1,2}/\\d{4}$");
```
cal-smith: could this be a regex literal - `/\d{1,2}/\d{1,2}/\d{4}/`?

src/datepicker-input/datepicker-input.component.ts
```diff
@@ -84,4 +88,13 @@ export class DatePickerInput {
 	@Input() invalidText: string;
 
 	@Input() skeleton = false;
+
+	onChange() {
+		this.valueChange.emit(this.dateInput.nativeElement.value);
```
cal-smith: Since we're binding directly to the element, we could just pass down the event and pull the value from `event.target` ... then we can also remove the extra `dateInput` bindings 👍

src/datepicker-input/datepicker-input.component.ts
```diff
@@ -84,4 +88,13 @@ export class DatePickerInput {
...
+		if (this.dateInput && this.dateInput.nativeElement.value === "") {
+			this.invalid = false;
```
cal-smith: I'm not sure about this ... do we have anything else that does this by default? And how does this behave with the forms module?

src/datepicker-input/datepicker-input.component.ts
```diff
@@ -86,12 +95,29 @@ export class DatePickerInput {
...
-			this.invalid = false;
-		} else {
-			this.invalid = !new RegExp(this.pattern).test(event.target.value);
+		this.propagateChange(event.target.value);
+		this.onTouched();
+	}
+
+	public writeValue(value: any) {
+		if (this.elementRef.nativeElement.querySelector("input")) {
+			this.elementRef.nativeElement.querySelector("input").value = value;
```
cal-smith: This should write directly to `this.value` ... which appears to be missing?

src/datepicker/datepicker.stories.ts
```diff
@@ -7,13 +7,86 @@ import {
...
+	get invalidRange() {
+		return this.rForm.controls["range"].invalid && this.rForm.controls["range"].touched;
+	}
+
+	rForm: FormGroup;
+
+	constructor(protected formBuilder: FormBuilder) {
+		this.rForm = this.formBuilder.group({
+			single: [
+				[(new Date().getMonth() + 1) + "/" + new Date().getDate() + "/" + new Date().getFullYear()],
```
cal-smith: Couldn't this just be `new Date()`?

src/datepicker/datepicker.stories.ts
```diff
@@ -7,13 +7,86 @@ import {
...
+				Validators.required
+			],
+			range: [
+				[
+					(new Date().getMonth() + 1) + "/" + new Date().getDate() + "/" + new Date().getFullYear(),
+					(new Date().getMonth() + 2) + "/" + new Date().getDate() + "/" + new Date().getFullYear()
```
cal-smith: Ditto, or at least use template strings :+1:

src/datepicker/datepicker.stories.ts
```diff
@@ -7,13 +7,86 @@ import {
...
+	selector: "app-date-picker",
+	template: `
+		<form [formGroup]="rForm">
+			<ibm-date-picker
+				formControlName="single"
+				label="Date Picker Label"
+				invalidText="Invalid date format"
+				[invalid]="invalidSingle"
+				(valueChange)="valueChange.emit($event)">
+			</ibm-date-picker><br>
```
cal-smith: Beneath each of these, in a `<code>` block, could we write out the `.value`? something like `rForm.controls["single"].value | json` should work

src/datepicker/datepicker.stories.ts
zvonimirfras: `new Date()` will also get the current time, if we want the start of the day, then no?

src/datepicker/datepicker.component.ts
```diff
@@ -141,8 +155,15 @@ export class DatePicker implements OnDestroy {
 
 	constructor(protected elementRef: ElementRef) { }
 
+	@HostListener("focusin")
+	onFocus() {
+		console.log(this.flatpickrOptions);
```
cal-smith: Probably didn't mean for this 🤔 
```suggestion
```

PR #540, opened 2019-05-10
Support render template in dropdown component

rudolpho1:
Extend the dropdown component to support rendering the current value via an angular template. This is similar to the template support the dropdown-list component already has.

#### Changelog

**Changed**

* extend the dropdown components `displayValue` input parameter to accept a template as well as a string.

### DIFF
--- a/src/dropdown/dropdown.component.ts
+++ b/src/dropdown/dropdown.component.ts
@@ -369,14 +369,14 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy {
 			return;
 		}
 		let selected = this.view.getSelected();
-		if (selected && (!this.displayValue || typeof this.displayValue !== "string")) {
+		if (selected && (!this.displayValue || !this.isRenderString())) {
 			if (this.type === "multi") {
 				return of(this.placeholder);
 			} else {
 				return of(selected[0].content);
 			}
-		} else if (selected && typeof this.displayValue === "string") {
-			return of(this.displayValue);
+		} else if (selected && this.isRenderString()) {
+			return of(this.displayValue as string);
 		}
 		return of(this.placeholder);
 	}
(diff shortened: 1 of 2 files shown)


rudolpho1:
@cal-smith I addressed the change you proposed and also added a story to the story board. Unfortunately, I had some tslint problems when pushing the change and ended up force pushing my fork. Thats why I had to create a new pull request.

cal-smith:
Totally understand ... Force pushes usually update the same PR though 👍

src/dropdown/dropdown.component.ts
```diff
@@ -358,23 +364,41 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy {
...
 	 * otherwise the placeholder will be returned.
 	 */
-	getDisplayValue(): Observable<string> {
+	getDisplayStringValue(): Observable<string> {
 		if (!this.view) {
 			return;
 		}
 		let selected = this.view.getSelected();
-		if (selected && !this.displayValue) {
+		if (selected && (!this.displayValue || typeof this.displayValue !== "string")) {
```
cal-smith: Since we have the method...
```suggestion
		if (selected && (!this.displayValue || this.isRenderString())) {
```

src/dropdown/dropdown.component.ts
```diff
@@ -358,23 +364,41 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy {
...
 			if (this.type === "multi") {
 				return of(this.placeholder);
 			} else {
 				return of(selected[0].content);
 			}
-		} else if (selected) {
+		} else if (selected && typeof this.displayValue === "string") {
```
cal-smith: Ditto here
```suggestion
		} else if (selected && this.isRenderString()) {
```

rudolpho1: Done.

PR #541, opened 2019-05-12
Added possibility to give action to the X button on Filter Tag

Pablodotnet:
Closes IBM/carbon-components-angular#538

This pull requests adds the possibility to add an action to trigger with the X button on the Filter Tag.

#### Changelog

**New**

* Added possibility to give action to X button on Filter Tag

### DIFF
--- a/src/tag/tag-filter.component.ts
+++ b/src/tag/tag-filter.component.ts
@@ -28,7 +28,6 @@ import { Tag } from "./tag.component";
 export class TagFilter extends Tag {
 	/**
 	 * Function for close/delete the tag
-	 *
 	 */
 	@Output() close = new EventEmitter<any>();
 
--- a/src/tag/tag-filter.component.ts
+++ b/src/tag/tag-filter.component.ts
@@ -1,4 +1,9 @@
-import { Component, Output, EventEmitter, HostBinding } from "@angular/core";
+import {
+	Component,
+	Output,
+	EventEmitter,
+	HostBinding
+} from "@angular/core";
 import { Tag } from "./tag.component";
 
 @Component({
@@ -6,7 +11,7 @@ import { Tag } from "./tag.component";
 	template: `
 		<ng-content></ng-content>
 		<svg
-			(click)="closeFn()"
+			(click)="close.emit()"
 			focusable="false"
 			preserveAspectRatio="xMidYMid meet"
 			style="will-change: transform;"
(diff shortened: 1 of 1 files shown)


src/tag/tag-filter.component.ts
```diff
@@ -20,7 +21,17 @@ import { Tag } from "./tag.component";
...
+	 * Function for close/delete the tag
+	 *
+	 */
+	@Output() close: EventEmitter<any> = new EventEmitter<any>();
+
 	@HostBinding("attr.class") get attrClass() {
 		return `bx--tag bx--tag--filter bx--tag--${this.type} ${this.class}`;
 	}
+
+	closeFn() {
```
cal-smith: Just a FYI, generally we [follow these naming conventions](https://github.com/IBM/carbon-components-angular/wiki/Style-guide#event-handler-and-emitter-handler-naming)

So this should be 
```suggestion
	doClose() {
```

_however_ since we don't need to do any extra processing, you can just call `close.emit()` in the template 👍

src/tag/tag-filter.component.ts
```diff
@@ -20,7 +21,17 @@ import { Tag } from "./tag.component";
 	`
 })
 export class TagFilter extends Tag {
+	/**
+	 * Function for close/delete the tag
+	 *
+	 */
+	@Output() close: EventEmitter<any> = new EventEmitter<any>();
```
cal-smith: We can take advantage of type inference here:
```suggestion
	@Output() close = new EventEmitter<any>();
```

src/tag/tag-filter.component.ts
```diff
@@ -1,11 +1,12 @@
-import { Component, HostBinding } from "@angular/core";
+import { Component, Output, EventEmitter, HostBinding } from "@angular/core";
```
cal-smith: Another small style change, but we try to [limit imports to 3 on one line](https://github.com/IBM/carbon-components-angular/wiki/Style-guide#import-ordering)

So:
```suggestion
import { 
	Component, 
	Output, 
	EventEmitter, 
	HostBinding 
} from "@angular/core";
```

Pablodotnet:
I'll make the changes requested! Thanks for the review

Pablodotnet:
Changes have been done

src/tag/tag-filter.component.ts
```diff
@@ -20,6 +26,12 @@ import { Tag } from "./tag.component";
 	`
 })
 export class TagFilter extends Tag {
+	/**
+	 * Function for close/delete the tag
+	 *
```
cal-smith: We can probably remove this line 🙂 
```suggestion
```

Pablodotnet: Done 👍

PR #547, opened 2019-05-15
feat(pagination): Added possibility to define the options for the items per page select

Pablodotnet:
Closes IBM/carbon-components-angular#546

This pull requests adds the possibility to define the options for the items per page select, they were always defined as 10, 20, 30, 40, 50, now a user can define whatever numbers they want.

#### Changelog

**New**

* Now is possible to set the options for the items per page select on pagination


cal-smith:
FYI travis seems to be rate limited at the moment. As soon as it starts behaving we'll merge this in.

PR #557, opened 2019-05-21
V2 - fix(dialog): use mouseleave by default

cal-smith:
closes https://github.com/IBM/carbon-components-angular/issues/552


cal-smith:
Apologies @magedhennawy for the ping ... that'll be fixed once #551 lands

PR #566, opened 2019-05-22
fix(date-picker): Update flatpickr

### LINKED ISSUE #483: Date Picker does not properly trigger events on manual input
## Detailed description
Currently the manual input on the range date picker has a couple of accessibility issues:
- The manual input does not trigger the valueChange() on changing from one input to another
- If a `return` is hit after the date is manually entered in the first input field, the event is triggered, but the focus is moved to the calendar instead of the other input field. 

#Carbon Version used
```
    "carbon-components": "9.91.0",
    "carbon-components-angular": "2.41.0",
```

#Example: Date filter in Assess UX Redesign
https://pages.github.ibm.com/talent-incubator/assess-ux-redesign/invitations/2

1. Go to [Assess UX Redesign Github Pages](https://pages.github.ibm.com/talent-incubator/assess-ux-redesign/invitations/2)
2. Click on the Date Filter
3. In the first input, add some text (following the format given in the place holder - an issue for lack of validation was also created - ticket https://github.com/IBM/carbon-components-angular/issues/482)
4. Move focus out.

*Current behaviour*: 
- The date is only stored when `return` is hit.
- After hitting `return`, the focus moves to the calendar (even if the date is already entered) instead of the other input field
- Hitting `tab` leads to the other input field - which is correct - but no event is triggered. 

The success flow currently is:
1. Focus on first input field
2. Enter a valid (formatted) date : `04/04/2019`
3. Hit `return` (the focus goes to the calendar, on the date entered)
4. Hit `tab` (the focus goes to the other date input field)
5. Enter a valid (formatted) date : `04/05/2019` 
6. Hit `return` (the focus goes to the calendar, on the date entered)
7. Hit `tab` (the focus goes to the buttons)

*Desired behaviour*: 
**(suggestion)** We should enter a value in the first field, then a listener checks if both fields are filled, the valueChange event is triggered and dates are stored

The success flow should be
1. Focus on first input field
2. Enter a valid (formatted) date : `04/04/2019`
3. Hit `tab` (the focus goes to the other date input field)
4. Enter a valid (formatted) date : `04/05/2019`
5. Hit `tab` (the focus goes to the buttons). 

In this moment - we should be able to use the dates 

## Additional information
* GIF of the issue * - Note that the first time I was following the steps as 'the current success flow' and then what happens on the desired one.
DateManual

youda97:
closes #483

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -763,9 +763,9 @@
       }
     },
     "@babel/parser": {
-      "version": "7.4.4",
-      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.4.4.tgz",
-      "integrity": "sha512-5pCS4mOsL+ANsFZGdvNLybx4wtqAZJ0MJjMHxvzI3bvIsz6sQvzW8XX92EYIkiPtIvcfG3Aj+Ir5VNyjnZhP7w==",
+      "version": "7.4.5",
+      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.4.5.tgz",
+      "integrity": "sha512-9mUqkL1FF5T7f0WDFfAoDdiMVPWsdD1gZYzSnaXsxUCUqzuch/8of9G3VUSNiZmMBoRxT3neyVsqeiL/ZPcjew==",
       "dev": true
     },
     "@babel/plugin-proposal-async-generator-functions": {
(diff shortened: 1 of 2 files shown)


cal-smith:
Any notable improvements with this?

youda97:
It fixes issue #483. At first, the date was not being saved when you tab out of the first input in the range selection.

PR #572, opened 2019-05-27
feat(radio-group): Vertial align

### LINKED ISSUE #569: Add ability to specify orientation of the radio buttons in a group
## Detailed description

There is currently no way to specify the direction that radio buttons are displayed within the <ibm-radio-group> control.  To satisfy my requirement, I've overridden the bx--radio-button-group class, but if I ever need to display a radio button group horizontally I'll be in trouble.  Please add the ability to specify if radio buttons are displayed horizontally or vertically within a radio button group.

## Additional information

Screen Shot 2019-05-23 at 1 26 55 PM

youda97:
closes #569


cal-smith:
Could you open an issue for `carbon-components` on the carbon monorepo about this:
image

src/radio/radio-group.component.ts
```diff
@@ -88,6 +93,10 @@ export class RadioGroup implements AfterContentInit, AfterViewInit, ControlValue
 	 */
 	static radioGroupCount = 0;
 
+	@Input() orientation: "horizontal" | "vertical" = "horizontal";
+
+	@Input() placement: "left" | "right" =  "left";
```
cal-smith: We have `placement` here and `labelLeft` in radio.component ... we should standardize on one ... 

Maybe `labelPlacement: "left" | "right"`? `placement` is fairly generic and I'm not sure if it would be confusing here

cal-smith:
Issue reference: https://github.com/carbon-design-system/carbon/issues/2286

cal-smith:
If you're feeling adventerious you could contribute the fix back to carbon-components ... I don't see any specific PRs open around this issue

src/radio/radio-group.component.ts
```diff
@@ -239,10 +248,13 @@ export class RadioGroup implements AfterContentInit, AfterViewInit, ControlValue
 	/**
 	 * Synchronizes the names of the radio items with the name of the `RadioGroup`.
 	 */
-	updateRadioNames() {
+	updateRadios() {
```
zvonimirfras: Why the name change? Technically, it's a public function so technically, this is a breaking change...

If this is something we actually want, we go the deprecate and substitute route.

src/radio/radio-group.component.ts
```diff
@@ -236,13 +246,21 @@ export class RadioGroup implements AfterContentInit, AfterViewInit, ControlValue
 		this.onTouched();
 	}
 
+	updateRadioNames() {
+		console.warn("updateRadioNames had been deprecated. Use updateRadios instead");
+		this.updateRadios();
+	}
+
 	/**
 	 * Synchronizes the names of the radio items with the name of the `RadioGroup`.
```
zvonimirfras: Update the documentation to reflect that it's not only the names.

PR #574, opened 2019-05-28
fix(tile-group): dynamic content 

### LINKED ISSUE #573: Fix tile-group dynamic content
`ngAfterContentInit` is only run once and it doesn't react to any changes to content.
Also, looks like subscribe is never unsubscribed, so, in dynamic situations, it leaks.

zvonimirfras:
Closes #573

### DIFF
--- a/src/tiles/tile-group.component.ts
+++ b/src/tiles/tile-group.component.ts
@@ -6,11 +6,14 @@ import {
 	EventEmitter,
 	HostBinding,
 	ContentChildren,
-	QueryList
+	QueryList,
+	OnDestroy
 } from "@angular/core";
 import { SelectionTile } from "./selection-tile.component";
 import { NG_VALUE_ACCESSOR } from "@angular/forms";
 import { TileSelection } from "./tile-selection.interface";
+import { Subject } from "rxjs";
+import { takeUntil } from "rxjs/operators";
 
 @Component({
 	selector: "ibm-tile-group",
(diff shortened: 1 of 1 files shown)


src/tiles/tile-group.component.ts
```diff
@@ -61,18 +66,32 @@ export class TileGroup implements AfterContentInit {
...
+				tile.multiple = this.multiple;
 			});
-			tile.multiple = this.multiple;
-		});
+		};
+		updateTiles();
+
+		this.selectionTiles.changes
+			.pipe(takeUntil(this.unsubscribe$))
+			.subscribe(_ => updateTiles());
```
cal-smith: You'll probably want to do a `this.unsubscribe$.next` here ... otherwise you may run into multiple `selected` events emitting

zvonimirfras: Wouldn't that unsubscribe right away? It happens onDestroy currently.

cal-smith: I mean, currently nothing is every _actually_ unsubscribed, it just stops processing events ... I'm actually not sure why `this.selectionTiles.changes` has a `takeUntil` ... it should be cleaned up/completed by angular in the destroy phase and any subscriptions will also stop receiving when the final `unsubscribe` is emitted.

package.json
```diff
@@ -85,13 +85,13 @@
     "@carbon/icons-angular": "10.2.0",
     "@commitlint/cli": "7.5.2",
     "@commitlint/config-conventional": "7.5.0",
-    "@compodoc/compodoc": "1.1.7",
+    "@compodoc/compodoc": "1.1.9",
```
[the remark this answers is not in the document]
zvonimirfras: It does look fine locally.

PR #583, opened 2019-05-31
fix(datepicker): Make flatpickrOptions an input

### LINKED ISSUE #575: `flatpickrOptions` and `flatpickrOptionsRange` should be inputs
This will allow devs to easily provide custom ones for, i.e. disabling dates.

youda97:
closes #575

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -25,13 +25,13 @@ import rangePlugin from "flatpickr/dist/plugins/rangePlugin";
 	template: `
 	<div class="bx--form-item">
 		<ng2-flatpickr
-			[setDate]= "value"
-			[config]= "range ? flatpickrOptionsRange : flatpickrOptions"
+			[setDate]="value"
+			[config]="flatpickrOptions"
 			[hideButton]="true">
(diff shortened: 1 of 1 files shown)


src/datepicker/datepicker.component.ts
```diff
@@ -120,20 +120,38 @@ export class DatePicker implements OnDestroy {
...
+	get flatpickrOptions(): FlatpickrOptions {
+		const plugins = [...this.plugins];
+		if (this.range) {
+			plugins.push(rangePlugin({ input: "#" + this.id + "-rangeInput"}));
+		}
+		return Object.assign({}, this._flatpickrOptions, this.flatpickrBaseOptions, {plugins});
+	}
+
+	@Input()
+	set flatpickrOptionsRange (options: FlatpickrOptions) {
```
zvonimirfras: Didn't we talk about deprecating range and just using `flatpickrOptions` with `range` controlling the switch between the versions?

Everything else seems correct except `flatpickrOptionsRange` became an input except getting deprecated.

cal-smith: lol yeah that shouldn't be an `Input` ... the rest of the implementation is valid I think though

PR #591, opened 2019-06-05
chore(storybook): Add title to header actions

### LINKED ISSUE #588: UI Shell - Hover on icon in Header is displaying 'undefined'

1. Visit https://angular.carbondesignsystem.com/?path=/story/ui-shell--header
2. Hover over the extreme right hand side two icons

youda97:
closes #588


cal-smith:
Should probably also give `title` a default value ... I think `null` should clear it out

youda97:
`null` will just print out null instead of undefined. So I used`""` to clear it out

PR #603, opened 2019-06-12
fix(datepicker): fix dateFormat option generation

cal-smith:


imandyie:
@cal-smith  was the console.log added on purpose?

cal-smith:
That was definitely a mistake @imandyie ... I believe #612 removes that (in addition to a number of other fixes)

PR #617, opened 2019-06-24
fix(dropdown): Swallows space keys after it's been opened

### LINKED ISSUE #604: Dropdown (v2): Swallows space keys after it's been opened
## Detailed description

I have a dropdown with an item template which adds a search box in the items list. The issue is that when I type something in the search box it gets through the `Dropdown.onKeyDown()` method first which swallows the spaces.

image

Bug

Dropdown

I expect the component to handle spaces only when it needs to open the menu and leave them to propagate otherwise.

Chrome

2.45.0

IBM ADDI

youda97:
closes #604

### DIFF
--- a/src/dropdown/list/dropdown-list.component.ts
+++ b/src/dropdown/list/dropdown-list.component.ts
@@ -67,7 +67,7 @@ import { Observable, isObservable, Subscription } from "rxjs";
 				}">
 				<div
 					#listItem
-					tabindex="0"
+					tabindex="-1"
 					class="bx--list-box__menu-item__option">
 					<div
 						*ngIf="!listTpl && type === 'multi'"
--- a/src/dropdown/list/dropdown-list.component.ts
+++ b/src/dropdown/list/dropdown-list.component.ts
@@ -54,8 +54,6 @@ import { Observable, isObservable, Subscription } from "rxjs";
 			class="bx--list-box__menu bx--multi-select"
 			[attr.aria-label]="ariaLabel">
 			<li
-				#listItem
-				tabindex="-1"
 				role="option"
 				*ngFor="let item of displayItems; let i = index"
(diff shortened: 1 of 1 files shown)


src/dropdown/list/dropdown-list.component.ts
```diff
@@ -405,10 +406,12 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 	doKeyDown(event: KeyboardEvent, item: ListItem) {
 		// "Spacebar", "Down", and "Up" are IE specific values
 		if (event.key === "Enter" || event.key === " " || event.key === "Spacebar") {
-			event.preventDefault();
-			if (event.key === "Enter") {
-				this.doClick(event, item);
-			}
+				if (this.listElementList.some(option => option.nativeElement === event.target)) {
+					event.preventDefault();
```
cal-smith: Did you check this with an `input` as a custom template? That's where the original issue ran into problems ...

src/dropdown/list/dropdown-list.component.ts
```diff
@@ -67,7 +65,10 @@ import { Observable, isObservable, Subscription } from "rxjs";
 					'bx--list-box__menu-item--active': item.selected,
 					disabled: item.disabled
 				}">
-				<div class="bx--list-box__menu-item__option">
+				<div
+					#listItem
+					tabindex="0"
```
cal-smith: Should this be `-1`? AFAIK we don't want users tabbing through the list

src/dropdown/list/dropdown-list.component.ts
youda97: Yup I tested it with the search and it was working

cal-smith: Awesome, we should be good to go then

PR #631, opened 2019-07-08
feat(Table): Added `rowsClass` property.

stanislavgeorgiev:
Related to IBM/carbon-components-angular#626. Can close it after merging the code into master

To test, just add:

```
simpleModel.rowsContext[0] = “error”;
simpleModel.rowsClass[0] = “111";
simpleModel.rowsClass[1] = “222”;
```

in the table.stories.ts, line 414 and then first data row should show contain `<tr class="111">` and second data row should contain the `222` class.

### DIFF
--- a/src/table/table.component.ts
+++ b/src/table/table.component.ts
@@ -313,6 +313,7 @@ import { I18n } from "./../i18n/i18n.module";
 				<tr *ngIf="!model.isRowFiltered(i)"
 					(click)="onRowSelect(i)"
 					[attr.data-parent-row]="(model.isRowExpandable(i) ? 'true' : null)"
+					class="{{model.rowsClass[i] ? model.rowsClass[i] : null}}"
 					[ngClass]="{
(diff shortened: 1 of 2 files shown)


src/table/table.component.ts
```diff
@@ -313,6 +313,7 @@ import { I18n } from "./../i18n/i18n.module";
 				<tr *ngIf="!model.isRowFiltered(i)"
 					(click)="onRowSelect(i)"
 					[attr.data-parent-row]="(model.isRowExpandable(i) ? 'true' : null)"
+					class="{{model.rowsClass[i] ? model.rowsClass[i] : null}}"
```
cal-smith: Is it possible to bind directly to `class`? so - 
```suggestion
					[class]="model.rowsClass[i] ? model.rowsClass[i] : null"
```

or 

```suggestion
					[attr.class]="model.rowsClass[i] ? model.rowsClass[i] : null"
```

stanislavgeorgiev: Sure, makes sense!

PR #634, opened 2019-07-08
fix(pagination): improve pagination performance

### LINKED ISSUE #633: Pagination Module: HUGE Performance Issue
# Pagination Module: HUGE Performance Issue

## Detailed description

For `totalDataLength` if a large value is assigned, the entire app freezes / slows down for 5-10 seconds. If the value is small (<1000) the performance is fine, anything larger like 200,000 causes a HUGE performance drag.

BUG

PaginationModule. Possible the `range` method in `common/utils`. We suspect it might be the use of `.fill` and also possibly `.map`. A `for` loop would be the most efficient way to generate the page list.

Instant rendering after the data is set; instead we saw a 5-10 second delay and the app / browser freezing.

Chrome

`3.11.2`

This is impacting 2 products in PRODUCTION! A release ASAP would be appreciated.

Internal Link to code: 

https://github.ibm.com/client-success/chi-explorer/blob/master/packages/ui/src/app/records/records.component.ts#L230

_(Ping me if you'd like a demo of it)_

cal-smith:
- add pageSelectThreshold to configure when to switch to an input
instead of a select
- remove calls to range

closes #633

### DIFF
--- a/src/pagination/pagination.component.ts
+++ b/src/pagination/pagination.component.ts
@@ -119,7 +119,7 @@ import { ExperimentalService } from "./../experimental.module";
 						[id]="currentPageSelectId"
 						class="bx--select-input"
 						[(ngModel)]="currentPage">
-						<option *ngFor="let _ of pageOptions; let i = index;" class="bx--select-option" [value]="i + 1">{{i + 1}}</option>
+						<option *ngFor="let page of pageOptions; let i = index;" class="bx--select-option" [value]="i + 1">{{i + 1}}</option>
 					</select>
 					<ibm-icon-chevron-down16
 						*ngIf="pageOptions.length <= 1000"
--- a/src/pagination/pagination.component.ts
+++ b/src/pagination/pagination.component.ts
@@ -6,7 +6,6 @@ import {
 	EventEmitter
 } from "@angular/core";
 
-import { range } from "../common/utils";
 import { I18n } from "./../i18n/i18n.module";
 import { ExperimentalService } from "./../experimental.module";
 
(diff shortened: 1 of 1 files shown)


src/pagination/pagination.component.ts
```diff
@@ -106,10 +105,24 @@ import { ExperimentalService } from "./../experimental.module";
 					'bx--select__page-number' : isExperimental
 				}">
 					<label [for]="currentPageSelectId" class="bx--label bx--visually-hidden">{{itemsPerPageText | async}}</label>
-					<select [id]="currentPageSelectId" class="bx--select-input" [(ngModel)]="currentPage">
-						<option *ngFor="let i of range(lastPage + 1, 1)" class="bx--select-option" [value]="i">{{i}}</option>
+					<input
+						*ngIf="pageOptions.length > pageSelectThreshold"
+						style="padding-right: 1rem; margin-right: 1rem;"
```
youda97: Is this needed? What does Carbon do?

src/pagination/pagination.component.ts
```diff
@@ -106,10 +105,24 @@ import { ExperimentalService } from "./../experimental.module";
...
+						min="1"
+						[max]="pageOptions.length"
+						class="bx--select-input"
+						[(ngModel)]="currentPage">
+					<select
+						*ngIf="pageOptions.length <= pageSelectThreshold"
+						[id]="currentPageSelectId"
+						class="bx--select-input"
+						[(ngModel)]="currentPage">
+						<option *ngFor="let _ of pageOptions; let i = index;" class="bx--select-option" [value]="i + 1">{{i + 1}}</option>
```
youda97: Is `let _ of pageOptions` needed?

cal-smith: yep, it's part of the nfForOf syntax .... otherwise we won't actually iterate anything

src/pagination/pagination.component.ts
cal-smith: The wrong thing 😛 this is a short term hack until we can get the design sorted out for >1000 pages

src/pagination/pagination.component.ts
youda97: Can we name it `page` or something instead of `_`

cal-smith: Sure ... my only concern with that is that someone might expect it to contain a valid value

PR #647, opened 2019-07-12
fix(datepicker): valueChange emits on change

### LINKED ISSUE #643: Changing date picker through keyboard does not work anymore
#600 did not fix it.

DatePickerInput emits a valueChange event and nothing in DatePicker is listening to this ... this is still the case.

cal-smith:
The primary fix here is switching control of value propagation away from
flatpickr and instead handling it ourselves. This has an additional
benefit of simplifying value handling for ngModel, reactive forms,
double binding, etc.

closes #643

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -113,7 +113,16 @@ export class DatePicker implements OnDestroy, AfterViewInit, OnChanges {
 
 	@Input() id = `datepicker-${DatePicker.datePickerCount++}`;
 
-	@Input() value: (Date | string)[] = [];
+	@Input() set value(v: (Date | string)[]) {
+		if (!v) {
+			v = [];
+		}
+		this._value = v;
+	}
+
+	get value() {
+		return this._value;
+	}
 
 	@Input() theme: "light" | "dark" = "dark";
 
@@ -155,6 +164,8 @@ export class DatePicker implements OnDestroy, AfterViewInit, OnChanges {
 
 	@Output() valueChange: EventEmitter<any> = new EventEmitter();
 
+	protected _value = [];
+
 	protected _flatpickrOptions = {
 		allowInput: true
 	};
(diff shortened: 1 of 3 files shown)


src/datepicker/datepicker.component.ts
```diff
@@ -177,16 +177,19 @@ export class DatePicker implements OnDestroy, AfterViewInit, OnChanges {
...
 			this.flatpickrInstance = flatpickr(`#${this.id}`, this.flatpickrOptions);
-			this.flatpickrInstance.setDate(dates);
 			this.setDateValues(dates);
 		}
 	}
 
 	ngAfterViewInit() {
 		this.flatpickrInstance = flatpickr(`#${this.id}`, this.flatpickrOptions);
+		if (this.value === null) {
+			this.value = [];
```
youda97: Isn't it already set to this when it is initialized?

src/datepicker/datepicker.component.ts
```diff
@@ -196,12 +199,68 @@ export class DatePicker implements OnDestroy, AfterViewInit, OnChanges {
...
-	doSelect(selectedValue) {
-		this.valueChange.emit(selectedValue);
-		this.propagateChange(selectedValue);
+	/**
+	 * Writes a value from the model to the component. Expects the value to be `null` or `(Date | string)[]`
+	 * @param value value recived from the model
+	 */
+	writeValue(value: (Date | string)[]) {
+		if (!value) {
+			value = [];
```
youda97: Couldn't this be set in the parameters?

src/datepicker/datepicker.component.ts
cal-smith: It is initialized to `[]`, however `ngModel` and friends can override that with an explicit `null`, so in this case we just reset it (since nothing else in the component is set up to handle `null`s)

**However** you bring up a good point ... I'll change this (and the other comment) out to use a setter/getter on the input instead so we can easilly catch any instance of `null` value passing

PR #649, opened 2019-07-15
feat(breadcrumb): add template option for items

### LINKED ISSUE #618: Breadcrumb: Need feature for custom component and threshold
Hi Team,

We have two ways to create breadcrumbs using carbon breadcrumb component:

1. 1st way is, by writing breadcrumb-item in which we can pass our own ng-content. However, this method **doesn't support threshold**.
2. 2nd way is, without breadcrumb-item in which we pass items property as array of content and href. This way does support threshold but **don't allow us to send our custom ng-content**.

In our project, we want both threshold and custom template support.

Probable Solution- 
As per my finding carbon team will have to fix it in the below code in <ibm-breadcrumb> component in both showforoverflow if conditions:

Existing code is:
<ibm-breadcrumb-item *ngFor="let item of items" [href]="item.href">
 {{item.content}}
</ibm-breadcrumb-item>

changes needed:
You first need to check instance of item.content. If it is string then you can render as it is. However, if it is templateRef, you need to render it inside ng-container templateOutlet. So, user will be able to send custom template in content property. (something similar is already implemented in tooltip component)

(I am using below version of carbon:
"carbon-components": "^10.2.0",
"carbon-components-angular": "^3.7.0".)

@zvonimirfras

cal-smith:
closes #618

### DIFF
--- a/src/breadcrumb/breadcrumb-item.interface.ts
+++ b/src/breadcrumb/breadcrumb-item.interface.ts
@@ -13,7 +13,7 @@ export interface BreadcrumbItem {
 	 */
 	href: string;
 	/**
-	 * Optional `TemplateRef` for the breadcrumb item. Recvices the content as an `$implicit` template variable
+	 * Optional `TemplateRef` for the breadcrumb item. Receives the content as an `$implicit` template variable
 	 */
 	template?: TemplateRef<any>;
 }
--- a/src/breadcrumb/breadcrumb-item.interface.ts
+++ b/src/breadcrumb/breadcrumb-item.interface.ts
@@ -1,20 +1,19 @@
+import { TemplateRef } from "@angular/core";
+
 /**
  * A structure that represents a breadcrumb item.
- *
- * @export
- * @interface BreadcrumbItem
  */
 export interface BreadcrumbItem {
 	/**
 	 * Content to be displayed in the breadcrumb item.
-	 * @type {string}
-	 * @memberof BreadcrumbItem
 	 */
 	content: string;
 	/**
 	 * Href for the breadcrumb item.
-	 * @type {string}
-	 * @memberof BreadcrumbItem
 	 */
 	href: string;
+	/**
+	 * Optional `TemplateRef` for the breadcrumb item. Recvices the content as an `$implicit` template variable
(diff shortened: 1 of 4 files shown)


src/breadcrumb/breadcrumb-item.interface.ts
```diff
@@ -1,20 +1,19 @@
...
 	 */
 	content: string;
 	/**
 	 * Href for the breadcrumb item.
-	 * @type {string}
-	 * @memberof BreadcrumbItem
 	 */
 	href: string;
+	/**
+	 * Optional `TemplateRef` for the breadcrumb item. Recvices the content as an `$implicit` template variable
```
youda97:
```suggestion
	 * Optional `TemplateRef` for the breadcrumb item. Receives the content as an `$implicit` template variable
```

src/breadcrumb/breadcrumb.stories.ts
```diff
@@ -59,6 +61,25 @@ storiesOf("Breadcrumb", module)
 		items: createBreadcrumbItems
 	}
 }))
+.add("Model with templates", () => ({
+	template: `
+	<ng-template #breadcrumbTemplate let-item>
+		Hello {{ item.content }}
```
youda97: Maybe make this a knob that allows them to change what's in the template

cal-smith: Yeah... could also make the `item.content` configurable as well 🤔

src/datepicker/datepicker.stories.ts
```diff
@@ -260,7 +260,7 @@ storiesOf("Date Picker", module)
 			date: new Date(new Date().getFullYear(), 5, 15),
 			rangeDates: [
 				new Date(new Date().getFullYear(), 5, 15),
-				new Date(new Date().getFullYear(), 9, 19)
+				new Date(new Date().getFullYear(), 8, 19)
```
youda97: Not sure if this was intentional or not 😅

cal-smith: I was off by a year before 😏 so yeah, intentional ... just an easter egg

PR #650, opened 2019-07-15
Add autocomplete input

### LINKED ISSUE #630: Pass down autocomplete to native input
Relevant for search, but potentially relevant for other wrapped inputs ... we should pass down the native HTML autocomplete attribute - https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

youda97:
closes #630

### DIFF
--- a/src/search/search.component.ts
+++ b/src/search/search.component.ts
@@ -109,6 +109,7 @@ export class Search implements ControlValueAccessor {
 	@Input() value = "";
 	/**
 	 * Sets the autocomplete attribute on the `input` element.
+	 * For refernece: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#Values
 	 */
 	@Input() autocomplete = "on";
 	/**
--- a/src/search/search.component.ts
+++ b/src/search/search.component.ts
@@ -108,6 +108,10 @@ export class Search implements ControlValueAccessor {
 	 */
 	@Input() value = "";
 	/**
+	 * Sets the autocomplete attribute on the `input` element.
+	 */
+	@Input() autocomplete = "on";
+	/**
 	 * Sets the text inside the `label` tag.
 	 */
 	@Input() label = this.i18n.get().SEARCH.LABEL;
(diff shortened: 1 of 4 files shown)


src/search/search.component.ts
```diff
@@ -107,6 +107,10 @@ export class Search implements ControlValueAccessor {
 	 * Sets the value attribute on the `input` element.
 	 */
 	@Input() value = "";
+	/**
+	 * Sets the autocomplete attribute on the `input` element.
+	 */
+	@Input() autocomplete = "on";
```
cal-smith: We should probably type hint the possible values

```suggestion
	@Input() autocomplete: "off" | "on" = "on";
```

src/search/search.stories.ts
```diff
@@ -11,12 +11,19 @@ storiesOf("Search", module).addDecorator(
...
+				[autocomplete]="autocomplete"
+				[disabled]="disabled"
+				[size]="size">
+			</ibm-search>
 		`,
 		props: {
 			size: select("size", ["lg", "sm"], "lg"),
 			theme: select("theme", ["dark", "light"], "dark"),
 			disabled: boolean("disabled", false),
+			autocomplete: text("autocomplete", "on"),
```
cal-smith: Maybe a `select` would make more sense here ...
```suggestion
			autocomplete: text("autocomplete", ["off", "on"], "on")
```

src/search/search.component.ts
youda97: There are other options though. Why are we limiting the user to just on and off?

cal-smith: I totally didn't realize there were more options ... maybe instead we should link out to somewhere that lists the options - like https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#Values

youda97: you want me to include this as a comment?

cal-smith: Yeah, something to that effect ... just so people are aware

PR #656, opened 2019-07-17
fix(search): valueChange

### LINKED ISSUE #651: Search Component: clear button doesn't have separate event
Right now, Search have same change event emitter for typing and cross button.
So, it is not possible to detect where user click the cross button or simply empty the text field manually because in both the scenarios, search is emitting empty field

Expectation: 
Separate event emitter for cross button click or something to identify separately.

youda97:
closes #651


src/search/search.component.ts
```diff
@@ -11,20 +11,6 @@ import {
 import { NG_VALUE_ACCESSOR, ControlValueAccessor } from "@angular/forms";
 import { I18n } from "../i18n/i18n.module";
 
-/**
- * Used to emit changes performed on search components.
- */
-export class SearchChange {
```
cal-smith: We can't remove this just yet ... we can definitely mark it deprecated however :+1:

src/search/search.component.ts
```diff
@@ -125,9 +111,13 @@ export class Search implements ControlValueAccessor {
 	 */
 	@Input() clearButtonTitle = this.i18n.get().SEARCH.CLEAR_BUTTON;
 	/**
-	 * Emits event notifying other classes when a change in state occurs in the input.
+	 * Emits an event when value is changed.
 	 */
-	@Output() change = new EventEmitter<SearchChange>();
```
cal-smith: Same as above, we'll need to keep this as deprecated until a major version

src/search/search.component.ts
```diff
@@ -125,9 +126,18 @@ export class Search implements ControlValueAccessor {
...
 	@Input() clearButtonTitle = this.i18n.get().SEARCH.CLEAR_BUTTON;
 	/**
+	 * @deprecated in the next major carbon-components-angular version in favour of `valueChange`
 	 * Emits event notifying other classes when a change in state occurs in the input.
 	 */
 	@Output() change = new EventEmitter<SearchChange>();
+	/**
+	 * Emits an event when value is changed.
+	 */
+	@Output() valueChange = new EventEmitter<any>();
```
cal-smith: In this case we know what types we're dealing with ...

```suggestion
	@Output() valueChange = new EventEmitter<string>();
```

src/search/search.component.ts
```diff
@@ -12,6 +12,7 @@ import { NG_VALUE_ACCESSOR, ControlValueAccessor } from "@angular/forms";
 import { I18n } from "../i18n/i18n.module";
 
 /**
+ * @deprecated in the next major carbon-components-angular version in favour of `valueChange`
```
zvonimirfras:
```suggestion
 * @deprecated in favor of `valueChange`, to be removed in the next major carbon-components-angular version
```

src/search/search.component.ts
```diff
@@ -125,9 +126,18 @@ export class Search implements ControlValueAccessor {
 	 */
 	@Input() clearButtonTitle = this.i18n.get().SEARCH.CLEAR_BUTTON;
 	/**
+	 * @deprecated in the next major carbon-components-angular version in favour of `valueChange`
```
zvonimirfras:
```suggestion
	 * @deprecated in favor of `valueChange`, to be removed in the next major carbon-components-angular version
```

PR #663, opened 2019-07-19
Add aria-label and focusout to tabs

### LINKED ISSUE #654: Need support for labels on tabs
## Detailed description

In trying to use the tab component, I need a way to add labels to the tab nav element. When there are multiple navigational elements in a page. A unique label is required. 

From DAP Chrome tool: 

https://aat.w3ibm.mybluemix.net/token/fc6655fc-6799-47a2-be20-9a6907013539/6dda174d-0092-4854-87be-d748c891b788/latest///doc/w3/help/en-US/idhi_accessibility_check_g1164.html

Currently using: 3.12.1

### LINKED ISSUE #659: Tab dropdown left open when I click out without selecting
## Detailed description

So when using the tabs in small mode, I cannot leave the dropdown without selecting an option.  I can click away and the tab dropdown remains open. 

Bug

Tabs in small mode

I would expect the dropdown to close if I focus off the dropdown regardless of making a selection. 

Chrome  Version 75.0.3770.100

    "carbon-components": "10.3.2",
    "carbon-components-angular": "3.12.1",

IWR 

## Additional information
https://angular.carbondesignsystem.com/?path=/story/tabs--basic  - You can see the issue in the storybook itself. 

In the dropdown storybook, I noticed i can leave the dropdown without making a selection.
https://angular.carbondesignsystem.com/?path=/story/dropdown--basic

youda97:
closes #654 
closes #659

### DIFF
--- a/src/tabs/tabs.component.ts
+++ b/src/tabs/tabs.component.ts
@@ -92,11 +92,11 @@ export class Tabs implements AfterContentInit {
 	 */
 	@Input() isNavigation = false;
 	/**
-	 * Sets the aria label on the `TableHeader`'s nav element.
+	 * Sets the aria label on the `TabHeader`s nav element.
 	 */
 	@Input() ariaLabel: string;
 	/**
-	 * Sets the aria labelledby on the `TableHeader`'s the nav element.
+	 * Sets the aria labelledby on the `TabHeader`s nav element.
 	 */
 	@Input() ariaLabelledby: string;
 
--- a/src/tabs/tabs.component.ts
+++ b/src/tabs/tabs.component.ts
@@ -51,7 +51,8 @@ import { TabHeaders } from "./tab-headers.component";
 				[cacheActive]="cacheActive"
 				[contentBefore]="before"
 				[contentAfter]="after"
-				[ariaLabel]="ariaLabel">
+				[ariaLabel]="ariaLabel"
+				[ariaLabelledby]="ariaLabelledby">
 			</ibm-tab-headers>
 			<ng-content></ng-content>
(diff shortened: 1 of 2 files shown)


youda97:
@cal-smith the DAP issue says "must have unique labels specified with aria-label or aria-labelledby". So if `aria-labelledby` is better here should I just use that?

cal-smith:
`aria-label` is for labeling directly on the element, `labelledby` is an ID that points to some other element that acts like a label. So we need both.

src/tabs/tabs.component.ts
```diff
@@ -89,6 +91,14 @@ export class Tabs implements AfterContentInit {
 	 * Set to `true` to have the tabIndex of the all tabpanels be -1.
 	 */
 	@Input() isNavigation = false;
+	/**
+	 * Sets the aria label on the `TableHeader`'s nav element.
```
cal-smith: `TabHeader`s ... no `'` needed either. Also on the comment for `ariaLabelledby`

```suggestion
	 * Sets the aria label on the `TabHeader`s nav element.
```

src/tabs/tabs.component.ts
```diff
@@ -89,6 +91,14 @@ export class Tabs implements AfterContentInit {
...
+	 */
+	@Input() ariaLabel: string;
+	/**
+	 * Sets the aria labelledby on the `TableHeader`'s the nav element.
```
cal-smith:
```suggestion
	 * Sets the aria labelledby on the `TabHeader`s nav element.
```

PR #667, opened 2019-07-22
fix(tooltip): pointer position

### LINKED ISSUE #512: bug(tooltip): tooltip caret always positioned at the top
## Detailed description

When changing the placement of the tooltip to left, right or top and caret it always positioned at the top and does not readjust.

bug

tooltip

caret should readjust instead of staying at the top of the tooltip. For example, if its placement is to the left and caret should be on the right of the tooltip.

Chrome

10.0.0

## Additional information

[storybook](https://angular.carbondesignsystem.com/?path=/story/tooltip--basic)

youda97:
closes #512


cal-smith:
image

Changing the props seems to blow away the content ...

PR #691, opened 2019-08-06
fix(i18n): allow for values to be overriden

cal-smith:
Adds the notion of an "Overridable" to `I18N`. From the comments:

> Largely an internal usecase. There are situations where we want an `Observable` that can emit events from a centralized source **OR** an `Observable` that will emit events from a component local source. The key example being on/off text in a `Toggle` - In some cases we want the `Toggle` to use `I18n`s global translations, but in others we'd prefer to use a local override. We don't ever need to return to an non-overriden state, but we do need the ability to switch _to_ an overridden sate.

### DIFF
--- a/src/i18n/i18n.service.ts
+++ b/src/i18n/i18n.service.ts
@@ -40,7 +40,7 @@ export const replace = (subject, variables) => subject.pipe(
  * can emit events from a centralized source **OR** an `Observable` that will emit events
  * from a component local source. The key example being on/off text in a `Toggle` - In some cases
  * we want the `Toggle` to use `I18n`s global translations, but in others we'd prefer to use a local
- * override. We don't ever need to return to an non-overriden state, but we do need the ability to
+ * override. We don't ever need to return to a non-overridden state, but we do need the ability to
  * switch _to_ an overridden sate.
  */
 export class Overridable {
--- a/src/i18n/i18n.service.ts
+++ b/src/i18n/i18n.service.ts
@@ -46,21 +46,31 @@ export const replace = (subject, variables) => subject.pipe(
  * switch _to_ an overridden sate.
  */
 export class Overridable {
-	public get value(): string {
+	public get value(): string | Observable<string> {
 		return this._value;
 	}
 
 	public get subject(): Observable<string> {
-		console.log(this.value);
+		/**
+		 * since inputs are bound on template instantiation (and thusly will always have _some_ value)
+		 * We can use a simple boolean and the `iif` function to determine which subject to return on subscription
+		 */
 		return iif(() => this.isOverridden, this.$override, this._subject);
 	}
 
-	protected _value = this.i18n.getValueFromPath(this.path);
-	protected $override = new BehaviorSubject<string>(this.value);
+	protected _value = this.i18n.getValueFromPath(this.path) as string;
+	protected $override: BehaviorSubject<string>;
 	protected _subject: Observable<string> = this.i18n.get(this.path);
 	protected isOverridden = false;
 
-	constructor(protected path: string, protected i18n: I18n) {	}
+	constructor(protected path: string, protected i18n: I18n) {
+		/**
+		 * ensure `$override` is initlized with the correct default value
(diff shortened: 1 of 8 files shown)


src/i18n/i18n.service.ts
```diff
@@ -125,4 +214,24 @@ export class I18n {
 		}
 		return value;
 	}
+
+	/**
+	 * Helper method that returns an observable from the interal cache based on path
```
youda97:
```suggestion
	 * Helper method that returns an observable from the internal cache based on the path
```

cal-smith:
Thanks for the spell check 🙈

src/i18n/i18n.service.ts
```diff
@@ -1,7 +1,7 @@
 import { Injectable } from "@angular/core";
-import { BehaviorSubject } from "rxjs";
+import { BehaviorSubject, Observable, isObservable, iif } from "rxjs";
```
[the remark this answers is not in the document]
cal-smith: My bad 🤦‍♂

src/i18n/i18n.service.ts
```diff
@@ -27,14 +32,88 @@ export const replace = (subject, variables) => subject.pipe(
...
+ * override. We don't ever need to return to a non-overridden state, but we do need the ability to
+ * switch _to_ an overridden sate.
+ */
+export class Overridable {
+	/**
+	 * The raw value of the translation. Defaults to the string value, but will return the value passed to `override`
+	 *
+	 * @readonly
+	 */
+	public get value(): string | Observable<string> {
```
zvonimirfras: Why not also have a setter? Instead of override or could call override.

src/i18n/i18n.service.ts
```diff
@@ -125,4 +219,24 @@ export class I18n {
...
+	/**
+	 * Helper method that returns an observable from the internal cache based on the path
+	 *
+	 * @param path looks like `"NOTIFICATION.CLOSE_BUTTON"`
+	 */
+	protected getSubject(path: string): Observable<string> {
+		try {
+			// we run this here to validate the path exists before adding it to the translation map
+			const value = this.getValueFromPath(path) as string;
+			if (this.translations.has(path)) {
```
zvonimirfras: Tiny performance improvement would be to use `get` right away and store the result, then check for not undefined to return the result.

src/toggle/toggle.component.ts
```diff
@@ -159,6 +160,14 @@ export class Toggle extends Checkbox {
 		Toggle.toggleCount++;
 	}
 
+	getOffText() {
```
zvonimirfras: getOffTextSubject perhaps? It looks confusing with the same name getters there.
same for getOnText
same for things in Table

src/i18n/i18n.service.ts
cal-smith: Given that a `Map` should have a `O(1)` worst case I don't think we'll see a tangible speedup from that. Plus that would involve another assignment and more garbage collection, so if anything it could be slower.

The current way is also very explicit

src/toggle/toggle.component.ts
cal-smith: On one hand I don't disagree, on the other if I type it to `getOffText(): Observable<string>` that may be more explicit and accomplish the same idea ... If we change in the future we won't have to rename all the methods

src/i18n/i18n.service.ts
cal-smith: Calling override is more explicit, while assigning to `.value` could be misinterpreted. I'm on the fence.

zvonimirfras: It looks very straightforward to do it and we can do both.

src/i18n/i18n.service.ts
zvonimirfras: If you feel this strongly about it, keep it.

src/i18n/i18n.service.ts
cal-smith: I'll add both and we can see which one makes more sense as we use it more ... it's all very subjective

PR #694, opened 2019-08-08
fix(tiles): Add link component to clickable tile

youda97:


cal-smith:
There's no changes?

youda97:
The other pr didn't trigger a release

PR #695, opened 2019-08-12
Add scroll utilities

cal-smith:
- fixes the dropdown scrolling away from it's list when the body scrolls
- migrates dialog to use the fancy new scroll utilities

### DIFF
--- a/src/dialog/dialog.component.ts
+++ b/src/dialog/dialog.component.ts
@@ -16,12 +16,12 @@ import {
 	fromEvent,
 	merge
 } from "rxjs";
-import { throttleTime } from "rxjs/operators";
+import { throttleTime, map } from "rxjs/operators";
 // the AbsolutePosition is required to import the declaration correctly
 import Position, { position, AbsolutePosition, Positions } from "@carbon/utils-position";
 import { cycleTabs, getFocusElementList } from "./../common/tab.service";
 import { DialogConfig } from "./dialog-config.interface";
-
+import { scrollableParentsObservable, isVisibleInContainer } from "./../utils/scroll";
 
 /**
  * Implements a `Dialog` that can be positioned anywhere on the page.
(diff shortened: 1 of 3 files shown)


src/dialog/dialog.component.ts
```diff
@@ -133,43 +133,14 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
...
-				computedStyle["overflow-x"] === "scroll"
-			);
-		};
-
-		const isVisibleInContainer = (element, container) => {
-			const elementRect = element.getBoundingClientRect();
-			const containerRect = container.getBoundingClientRect();
-			return elementRect.bottom <= containerRect.bottom && elementRect.top >= containerRect.top;
-		};
+		const parentElement: HTMLElement = this.dialogConfig.parentRef.nativeElement;
```
youda97: Does it need to be defined as an `HTMLElement`?

src/utils/scroll.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Returns an observable that emits whenever any scrollable parent element scrolls
+ *
+ * @param node root element to start finding scrolling parents from
+ */
+export const scrollableParentsObservable = (node: HTMLElement) => {
+	const windowScroll = fromEvent(window, "scroll").pipe(map(event => (
+		// update the event target to be something useful. In this case `body` is a
+		// sensible replacement
```
youda97: Could probably fit in the comment line above

src/dialog/dialog.component.ts
cal-smith: Let me check, but maybe 🤔

src/utils/scroll.ts
cal-smith: True, I'll tweak that

src/dialog/dialog.component.ts
cal-smith: nope! it's already typed as `HTMLElement` from `.nativeElement`

PR #697, opened 2019-08-12
Update updateSelected update pill on multi select

magedhennawy:
Closes IBM/carbon-components-angular#640


cal-smith:
Can you also open a PR with this for the `v9` branch? The original issue was for v2 carbon-angular

PR #698, opened 2019-08-13
Fix `ng build --prod` errors

youda97:


src/toggle/toggle.component.ts
```diff
@@ -105,8 +105,8 @@ export class Toggle extends Checkbox {
 	 * Text that is set on the left side of the toggle.
 	 */
 	@Input()
-	set offText(value: string) {
-		this._offText.next(value);
+	set offText(value) {
+		this._offText.next(value.OFF);
```
cal-smith: why `value.OFF`? That would require passing `[offText]="{OFF: 'text`}"`, the value _should_ be a string.

src/toggle/toggle.component.ts
```diff
@@ -117,8 +117,8 @@ export class Toggle extends Checkbox {
 	 * Text that is set on the right side of the toggle.
 	 */
 	@Input()
-	set onText(value: string) {
-		this._onText.next(value);
+	set onText(value) {
+		this._onText.next(value.ON);
```
cal-smith: ditto

src/toggle/toggle.component.ts
cal-smith: In any event this should be sorted by https://github.com/IBM/carbon-components-angular/pull/691

PR #724, opened 2019-08-27
fix(tile): Expression changed after checked

youda97:


src/tiles/selection-tile.component.ts
```diff
@@ -69,12 +69,13 @@ export class SelectionTile {
...
 	 */
-	name: string;
+	name = "tile-group-unbound";
 	/**
 	 * Defines whether or not the `SelectionTile` supports selecting multiple tiles as opposed to single
 	 * tile selection.
 	 */
-	multiple: boolean;
+	multiple = true;	// Set to true because of the way tile group sets it up.
+						// If it is first undefined then set to true, the type will change from radio to checkbox and deselects the inputs.
```
cal-smith: Was the old default radio or checkbox? Undefined seems to imply that it was radio?

youda97: ya it was radio first, and that was the problem

PR #740, opened 2019-09-09
fix template tab display wrong

fkaibj:

{{When the tab is template and the window is resized to be changed to a dropdown list; The selection will not make the tab name change}}

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{Add template processing, when the tab is template}}

**Removed**

* {{removed thing}}


cal-smith:
Additionally, could you amend your commit to take the form `fix(tabs): template tab display wrong`, just to ensure we trigger a release

PR #742, opened 2019-09-10
(Number) Ensure type consistency in change event

betoroth:
(Number) Ensure type consistency in change event

When a value is typed in the component, the event returns a string value (NumberChange.value). It should be better to have consistent types when we click on up/down arrows and also when we enter a manual value.

#### Changelog

**Changed**

Force the event.value to be a number


src/number-input/number.component.ts
```diff
@@ -246,7 +246,7 @@ export class Number implements ControlValueAccessor {
 	emitChangeEvent(): void {
 		let event = new NumberChange();
 		event.source = this;
-		event.value = this.value;
+		event.value = +this.value;
```
cal-smith: It may be more explicit/consistent to just call `Number` ... like:
```suggestion
		event.value = Number(this.value);
```

betoroth: I think I tried to use that, but `Number` was colliding with the component name. I will give it a try tomorrow.

cal-smith: Ah shoot, that's right ... This is a touch more complicated, but can we rename the component to `NumberComponent` and add an `export { NumberComponent as Number };` to the end of the component and the end of the module? We'll have to fix the naming with v11

betoroth:
Hi @cal-smith would you mind taking a look at the changes I did after your last comment? Thanks ;)

src/number-input/number.component.ts
cal-smith: Does `Number(this.value);` work now? If not we can take this for now

betoroth:
Oops! Did all the other changes and forgot to add `Number(this.value)` which was the most important one. Now we should be all set.

cal-smith:
Awesome!

PR #757, opened 2019-09-18
Add tests to accordion

Donisius:


src/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+		element = fixture.debugElement.query(By.css(".bx--accordion__heading"));
+		spyOn(wrapper, "onClick");
+		element.triggerEventHandler("click", null);
+		fixture.detectChanges();
+		expect(wrapper.onClick).toHaveBeenCalled();
+	});
+
+	it("should expand on button click", () => {
+		fixture = TestBed.createComponent(AccordionTest);
+		wrapper = fixture. componentInstance;
```
cal-smith: an extra space seems to have wandered in between the `fixture.` and `componentInstance` (for the next three tests)

src/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+import { Component } from "@angular/core";
+import { ChevronRight16Module } from "@carbon/icons-angular/lib/chevron--right/16";
+import { AccordionItem } from "./accordion-item.component";
+import { Accordion } from "./accordion.component";
+
+@Component({
+	template: `
+	<ibm-accordion>
+		<ibm-accordion-item
+		(selected)="onClick()"
```
zvonimirfras: This block should be indented.
It looks like `<ibm-accordion>` isn't closed.

src/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+describe("Accordion", () => {
+	let fixture, wrapper, element;
+	beforeEach(() => {
+		TestBed.configureTestingModule({
+			declarations: [
+				Accordion,
+				AccordionItem,
+				AccordionTest
+			],
+
```
zvonimirfras:
```suggestion
```

src/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+		expect(fixture.componentInstance instanceof Accordion).toBe(true);
+	});
+
+	it("should call onClick on button click", () => {
+		fixture = TestBed.createComponent(AccordionTest);
+		wrapper = fixture.componentInstance;
+		fixture.detectChanges();
+		element = fixture.debugElement.query(By.css(".bx--accordion__heading"));
+		spyOn(wrapper, "onClick");
+		element.triggerEventHandler("click", null);
```
zvonimirfras: Why not just use [click](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click) instead of this (everywhere its used)?

```suggestion
		element.click();
```

src/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+		fixture.detectChanges();
+		expect(element.nativeElement.getAttribute("aria-expanded")).toEqual("true");
+	});
+
+	it("should set test-content into accordion item", () => {
+		fixture = TestBed.createComponent(AccordionTest);
+		wrapper = fixture.componentInstance;
+		fixture.detectChanges();
+		element = fixture.debugElement.query(By.css("ibm-accordion .bx--accordion__content"));
+		expect(element.nativeElement.textContent).toEqual(" test-content ");
```
zvonimirfras: What's with the spaces?
Maybe use contain?

src/accordion/accordion.component.spec.ts
Donisius: It tells me that element.click() is not a function. Another way I could've done it is to use dispatchEvent on fixture.nativeElement is that would be better?

src/accordion/accordion.component.spec.ts
Donisius: It does not run when I remove these two declarations.

zvonimirfras: If you commit this suggestion, you'll notice that it's only removing the empty line :)

src/accordion/accordion.component.spec.ts
cal-smith: `triggerEventHandler` is the correct API to use :+1:

src/accordion/accordion.component.spec.ts
Donisius: Oh I see :)

src/accordion/accordion.component.spec.ts
cal-smith: For clarity you may want to rename `element` to `debugElement`

src/accordion/accordion.component.spec.ts
```diff
@@ -44,13 +45,13 @@ describe("Accordion", () => {
...
 		fixture = TestBed.createComponent(AccordionTest);
-		wrapper = fixture.componentInstance;
+		element = fixture.debugElement.query(By.css("ibm-accordion-item"));
 		fixture.detectChanges();
-		element = fixture.debugElement.query(By.css(".bx--accordion__heading"));
-		spyOn(wrapper, "onClick");
-		element.triggerEventHandler("click", null);
+		spyOn(element.componentInstance.selected, "emit");
+		let change = fixture.nativeElement.querySelector(".bx--accordion__heading");
+		change.dispatchEvent(new Event("click"));
```
cal-smith: `triggerEventHandler` was correct

src/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+import { AccordionItem } from "./accordion-item.component";
+import { Accordion } from "./accordion.component";
+
+@Component({
+	template: `
+	<ibm-accordion>
+		<ibm-accordion-item
+		[title]="title"
+		[skeleton]="skeleton">
+		test-content
```
zvonimirfras:
```suggestion
			test-content
```

src/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+	it("should set test-content into accordion item", () => {
+		fixture = TestBed.createComponent(AccordionTest);
+		wrapper = fixture.componentInstance;
+		fixture.detectChanges();
+		debugElement = fixture.debugElement.query(By.css("ibm-accordion .bx--accordion__content"));
+		expect(debugElement.nativeElement.textContent).toContain("test-content");
+	});
+
+	it("should set ibm-accordion item title to Section 1", () => {
```
zvonimirfras:
```suggestion
	it("should set ibm-accordion item title to 'Section 1'", () => {
```

PR #758, opened 2019-09-19
Add tests to timepicker-select

Donisius:

PR #763, opened 2019-09-23
Add tests to timepicker

Donisius:


src/timepicker/timepicker.component.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+@Component({
+	template: `
+	<ibm-timepicker
+		theme="dark"
+		(valueChange)="onChange()"
+		[(value)]="value"
+		label="test"
+		disabled="false"
+		>
```
cal-smith: `>` Should be on the previous line

src/timepicker/timepicker.component.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+	it("should set ng-reflect-disabled to true", () => {
+		fixture = TestBed.overrideComponent(TimePickerTest, {
+			set: {
+				template: `<ibm-timepicker disabled="true"></ibm-timepicker>`
+			}
+		}).createComponent(TimePickerTest);
+		fixture.detectChanges();
+		element = fixture.debugElement.query(By.css("ibm-timepicker")).nativeElement;
+		expect(element.getAttribute("ng-reflect-disabled")).toEqual("true");
```
cal-smith: Same comment as #758 - A better (more reliable) test would be checking for disabled on the underlying input ... ng-reflect-* aren't guaranteed to be stable/exist.

src/timepicker/timepicker.component.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+		element = fixture.debugElement.query(By.css(".bx--time-picker__input-field"));
+		fixture.detectChanges();
+		expect(element.nativeElement.value).toEqual("12:12");
+	});
+
+	it("should call onChange", () => {
+		fixture = TestBed.createComponent(TimePickerTest);
+		wrapper = fixture.componentInstance;
+		fixture.detectChanges();
+		let de = fixture.debugElement.query(By.css(".bx--time-picker__input-field"));
```
zvonimirfras:
```suggestion
		const de = fixture.debugElement.query(By.css(".bx--time-picker__input-field"));
```

src/timepicker/timepicker.component.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+		spyOn(wrapper, "onChange");
+		de.triggerEventHandler("change", {target: {value: ""}});
```
zvonimirfras: This vs `dispatchEvent`?

src/timepicker/timepicker.component.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+		fixture.detectChanges();
+		expect(wrapper.onChange).toHaveBeenCalled();
+	});
+
+	it("should change the value to 04:33", () => {
```
zvonimirfras: Can't this be part of `should call onChange` with 2 calls to `expect`?

It seems like they do the same thing, in different ways. Having one test case with 2 expects would speed up testing since it won't have to create everything twice.

src/timepicker/timepicker.component.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+		element = fixture.debugElement.query(By.css(".bx--time-picker__input-field")).nativeElement;
+		fixture.detectChanges();
+		element.value = "04:33";
+		fixture.detectChanges();
+		element.dispatchEvent(new Event("change"));
+		fixture.detectChanges();
+		expect(wrapper.value).toEqual("04:33");
+	});
+
+	it("should set ng-reflect-disabled to true", () => {
```
zvonimirfras: It's not really testing for that, is it? :)

src/timepicker/timepicker.component.spec.ts
cal-smith: https://angular.io/api/core/DebugElement

`dispatchEvent` is a HTML API, `triggerEventHandler` should be preferred since this is testing an Angular element

src/timepicker/timepicker.component.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+	it("should call onChange", () => {
+		fixture = TestBed.createComponent(TimePickerTest);
+		wrapper = fixture.componentInstance;
+		fixture.detectChanges();
+		const de = fixture.debugElement.query(By.css(".bx--time-picker__input-field"));
+		de.value = "04:33";
+		spyOn(wrapper, "onChange");
+		de.triggerEventHandler("change", {target: {value: "04:33"}});
+		fixture.detectChanges();
+		expect(wrapper.onChange).toHaveBeenCalled();
```
zvonimirfras: Second `expect` (from the removed test) is missing
```suggestion
		expect(wrapper.onChange).toHaveBeenCalled();
		expect(wrapper.value).toEqual("04:33");
```

PR #767, opened 2019-09-24
Add unit tests for datepicker

Donisius:


src/datepicker/datepicker.component.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+		[disabled]="disabled"
+		[invalid]="invalid"
+		[invalidText]="invalidText"
+		[dateFormat]="dateFormat"
+		(valueChange)="valueChange($event)">
+	</ibm-date-picker>
+	`
+})
+class DatePickerTest {
+	label = "Date picker label";
```
zvonimirfras: Some of these properties can be set inline in the template since they are never changed.

src/datepicker/datepicker.component.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+		fixture.detectChanges();
+		expect(element.componentInstance.invalid).toBe(true);
+	});
+
+	it("should call onValueChange on valueChange event", () => {
+		fixture = TestBed.createComponent(DatePickerTest);
+		element = fixture.debugElement.query(By.css("ibm-date-picker"));
+		fixture.detectChanges();
+		spyOn(element.componentInstance.valueChange, "emit");
+		let change = fixture.nativeElement.querySelector("ibm-date-picker-input");
```
zvonimirfras:
```suggestion
		const change = fixture.nativeElement.querySelector("ibm-date-picker-input");
```

src/datepicker/datepicker.component.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+		element = fixture.debugElement.query(By.css("ibm-date-picker"));
+		fixture.detectChanges();
+		expect(element.componentInstance.dateFormat).toEqual("m/d/Y");
+	});
+
+	it("should set theme to dark", () => {
+		fixture = TestBed.createComponent(DatePickerTest);
+		element = fixture.debugElement.query(By.css("ibm-date-picker"));
+		fixture.detectChanges();
+		expect(element.componentInstance.theme).toEqual("dark");
```
cal-smith: This should probably also check for the dark classes to be applied

src/datepicker/datepicker.component.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+	});
+
+	it("should set invalid to true", () => {
+		fixture = TestBed.createComponent(DatePickerTest);
+		element = fixture.debugElement.query(By.css("ibm-date-picker"));
+		fixture.componentInstance.invalid = true;
+		fixture.detectChanges();
+		expect(element.componentInstance.invalid).toBe(true);
```
cal-smith: Is there some other change we can look for? An additional element or class?

src/datepicker/datepicker.component.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+		element = fixture.debugElement.query(By.css("ibm-date-picker"));
+		fixture.detectChanges();
+		expect(element.nativeElement.textContent).toEqual("Date picker label");
+	});
+
+	it("should set input value to 10/19/2019", () => {
+		fixture = TestBed.createComponent(DatePickerTest);
+		fixture.detectChanges();
+		element = fixture.debugElement.query(By.css("input"));
+		expect(element.nativeElement.value).toEqual("10/19/2019");
```
cal-smith: Should also check for changing the value

src/datepicker/datepicker.component.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+		expect(element.componentInstance.invalid).toBe(true);
+	});
+
+	it("should call onValueChange on valueChange event", () => {
+		fixture = TestBed.createComponent(DatePickerTest);
+		element = fixture.debugElement.query(By.css("ibm-date-picker"));
+		fixture.detectChanges();
+		spyOn(element.componentInstance.valueChange, "emit");
+		let change = fixture.nativeElement.querySelector("ibm-date-picker-input");
+		change.dispatchEvent(new Event("valueChange"));
```
cal-smith: `triggerEventHandler` may be a better choice? Also we should make sure (either in this test case or another) that a useful value is emitted

src/datepicker/datepicker.component.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+	});
+
+	it("should set date format to m/d/Y", () => {
+		fixture = TestBed.createComponent(DatePickerTest);
+		element = fixture.debugElement.query(By.css("ibm-date-picker"));
+		fixture.detectChanges();
+		expect(element.componentInstance.dateFormat).toEqual("m/d/Y");
+	});
+
+	it("should set theme to light", () => {
```
cal-smith: There used to be a test for the `dark` theme ... is that something we could add?

Donisius: A class is introduced when the theme is set to light:
Screen Shot 2019-09-27 at 2 33 26 PM

But remains default when set to dark:
Screen Shot 2019-09-27 at 2 33 40 PM

I thought it may be more interesting to test for the existence of the light class.

cal-smith: Ah perfect, that makes sense

src/datepicker/datepicker.component.spec.ts
Donisius: I can't seen to find anything else besides ng-reflect attributes.

PR #777, opened 2019-09-26
feat(Notification) Add option to remove close icon

### LINKED ISSUE #738: Notification Component - No option to hide close icon
Notification messages don't have an option to hide the close icon. Need to add an attribute for the same.

Donisius:
Closes #738


src/notification/notification.component.ts
```diff
@@ -86,10 +87,15 @@ export class Notification {
 	@HostBinding("class.bx--inline-notification--success") get isSuccess() { return this.notificationObj.type === "success"; }
 	@HostBinding("class.bx--inline-notification--warning") get isWarning() { return this.notificationObj.type === "warning"; }
 
+	get isShowClose() {
```
cal-smith: We could probably just leave this as `showClose` ... or rename it to something that makes sense grammatically 👍

src/notification/toast.component.ts
```diff
@@ -64,6 +65,10 @@ export class Toast extends Notification implements OnInit {
 	@HostBinding("class.bx--toast-notification--success") get isSuccess() { return this.notificationObj["type"] === "success"; }
 	@HostBinding("class.bx--toast-notification--warning") get isWarning() { return this.notificationObj["type"] === "warning"; }
 
+	get isShowClose() {
```
cal-smith: Same comment as for notification here ... we may even be able to just use the getter from notification

Donisius: Good point!

PR #778, opened 2019-09-26
fix(Dropdown) Fix error when the list of items is empty

### LINKED ISSUE #771: Dropdown throws an error when the list of items is empty
## Detailed description

> Component: Dropdown

> When the item list is empty and we try to open the dropdown, we have this error in the console: `ERROR TypeError: Cannot read property 'nativeElement' of undefined`. The component is trying to focus on the first element of the list even when it is empty.

>According to the stacktrace, it fails on dropdown-list.component.ts:348 and it seems to be caused by the call to the function `initFocus()`, when the item list is empty, it set `index = 0`  and then it invokes to `getCurrentElement()` where `listElementList.toArray()[0]` is undefined, so asking for `.nativeElement` throws the error.

## Additional information

* Screenshots

image

image

Donisius:
Closes #771

### DIFF
--- a/src/dropdown/list/dropdown-list.component.ts
+++ b/src/dropdown/list/dropdown-list.component.ts
@@ -374,21 +374,26 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 		if (!Array.isArray(value)) {
 			console.error(`${this.constructor.name}.propagateSelected expects an Array<ListItem>, got ${JSON.stringify(value)}`);
 		}
-		for (let newItem of value) {
+		// loop through the list items and update the `selected` state for matching items in `value`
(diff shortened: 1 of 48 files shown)


src/dropdown/list/dropdown-list.component.ts
```diff
@@ -404,7 +404,7 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 		if (this.index < 0) {
 			this.index = 0;
 		}
-		this.getCurrentElement().focus();
+		this.getCurrentElement().focus();
```
cal-smith: Looks like a trailing space crept in here ... 
```suggestion
		this.getCurrentElement().focus();
```

PR #782, opened 2019-09-27
fix(ui-shell): remove close icon

cal-smith:
- removes the close icon (consumers can add it back with an ngIf on their end, not needed with the updated carbon styles)
- updates the header to take a route


cal-smith:
It's not clear to me that the close icon is part of the carbon specs ... it doesn't appear to be 😕 https://www.carbondesignsystem.com/components/UI-shell-right-panel/code

PR #791, opened 2019-10-01
Update unit tests for dropdown

Donisius:


src/dropdown/dropdown.component.spec.ts
```diff
@@ -10,26 +10,27 @@ import { I18nModule } from "../i18n/i18n.module";
...
 import { ChevronDown16Module } from "@carbon/icons-angular/lib/chevron--down/16";
+import { FormsModule } from "@angular/forms";
 
 @Component({
 	template: `
 	<ibm-dropdown
 		placeholder="test"
 		class="custom-class"
-		(selected)="onSelect($event)">
+		(selected)="onSelect($event)"
```
cal-smith: If `onSelect` doesn't take an argument we can remove the `$event`

PR #793, opened 2019-10-01
Add tests for progress-indicator

Donisius:


src/progress-indicator/progress-indicator.component.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+	});
+
+	it("should set current step to Fourth step and set warning icon when step is in error state", () => {
+		fixture = TestBed.createComponent(ProgressIndicatorTest);
+		wrapper = fixture.componentInstance;
+		wrapper.current = 3;
+		fixture.detectChanges();
+		element = fixture.debugElement.query(By.css("ibm-progress-indicator"));
+		expect(element.nativeElement.querySelector(".bx--progress-step--current").textContent).toContain("Fourth step");
+		expect(element.nativeElement.querySelector(".bx--progress-step--current ibm-icon-warning16")).toBeTruthy();
```
cal-smith: Is there a class we can check for instead of the component name? Just because that _will_ be changing

PR #796, opened 2019-10-02
Add tests to slider

Donisius:


src/slider/slider.component.spec.ts
```diff
@@ -0,0 +1,104 @@
+import { Component } from "@angular/core";
+import { TestBed } from "@angular/core/testing";
+import { By	 } from "@angular/platform-browser";
```
cal-smith: Maybe it's just github, but I'm seeing some weird spacing here
```suggestion
import { By } from "@angular/platform-browser";
```

src/slider/slider.component.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+		expect(element.componentInstance.isMouseDown).toBe(true);
+	});
+
+	it("should increment value on ArrowRight KeyboardEvent by the given shiftMultiplier", () => {
+		spyOn(wrapper, "onValueChange");
+		const arrowRight = new KeyboardEvent("keydown", {
+			"key": "ArrowRight",
+			shiftKey: true
+		});
+		element.nativeElement.querySelector(".bx--slider__thumb").dispatchEvent(new Event("mousedown"));
```
cal-smith: Do we need the mousedown for the test to work?

src/slider/slider.component.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+			]
+		});
+	});
+
+	beforeEach(() => {
+		fixture = TestBed.createComponent(SliderTest);
+		element = fixture.debugElement.query(By.css("ibm-slider"));
+		wrapper = fixture.componentInstance;
+		fixture.detectChanges();
+	});
```
natashadecoste: Some of the tests assign the fixture, element, wrapper within the` it()` blocks and some are using a second  `beforeEach()`.
I favour this way, as it reduces lots of duplicate code, but maybe we can commit to one way or another and update the tests for consistency and ease of contribution in the future

PR #801, opened 2019-10-03
fix(ui-shell): add header router changes from v3

cal-smith:
Closes IBM/carbon-components-angular#779

### DIFF
--- a/src/ui-shell/header.component.ts
+++ b/src/ui-shell/header.component.ts
@@ -1,6 +1,23 @@
-import { Component, Input, HostBinding } from "@angular/core";
-import { I18n } from "./../i18n/i18n.module";
+import {
+	Component,
+	Input,
+	Optional,
+	Output,
+	EventEmitter
+} from "@angular/core";
+import { DomSanitizer } from "@angular/platform-browser";
+import { Router } from "@angular/router";
+import { I18n } from "../i18n/i18n.module";
 
+/**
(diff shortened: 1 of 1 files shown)


src/ui-shell/header.component.ts
```diff
@@ -9,13 +26,17 @@ import { I18n } from "./../i18n/i18n.module";
...
 				[href]="skipTo"
 				tabindex="0">
 				{{ i18n.get("UI_SHELL.SKIP_TO") | async }}
 			</a>
 			<ng-content select="ibm-hamburger"></ng-content>
-			<a class="bx--header__name" href="#">
+			<a
+				class="bx--header__name"
+				href="#"
+				(click)="navigate($event)">
```
rapzo: Sorry for the intrusion here but wouldn't it be better to have `(selected)="navigate($event)"` to have the full hamburger functionality available for test?
I'm still getting my kicks around angular so i don't even know if that's the right MO.

cal-smith: `click` is the standard HTML click event :+1: `selected` is just our convention for generic item selection events

cal-smith: (And in many places we're shifting towards `value` + `valueChange` or `change` where that makes sense)

rapzo: Got it. Thanks.

PR #808, opened 2019-10-07
feat(ui shell): Make skipToContent optional

### LINKED ISSUE #772: UIShell - Header : make skipToContent optional
Hi,

Would it be possible to also make this skipToContent `<a href>` to become optional, using `*ngIf="skipTo"` ? This skipToContent is not always necessarily needed, implementable.

Thanks.

Donisius:
Closes #772


cal-smith:
there seems to be some conflics @Donisius ... if you sort those out we should be good to go :+1:

Donisius:
> there seems to be some conflics @Donisius ... if you sort those out we should be good to go 👍

I think this was already fixed in #782.

PR #810, opened 2019-10-08
fix(ui shell): change to correct hamburger icon

### LINKED ISSUE #773: UIShell - Header: Incorrect hamburger icon
Hi,

Our UX team is pointing that the hamburger icon that is used here is incorrect, it should be one with the 4 lines like this:

image

Thanks.

Donisius:
Closes #773


zvonimirfras:
You need to also change the commit message ;)

PR #811, opened 2019-10-08
fix(select): align inline select arrow

### LINKED ISSUE #765: Inline select arrow not aligned correctly
## Detailed description

The inline select's arrow is not aligned correctly.

It's a bug.

The select component with the `display` input set to "inline".

I expected the arrow to be aligned correctly. It's not aligned correctly. I would like it to be aligned correctly.

Chrome

`10.6.2`, but the issue can be observed in the Carbon Angular Storybook.

1. Use `ibm-select` with the `display` input set to "inline"

or

1. Go to the select component demo in the Carbon Angular Storybook
2. Under the **Knobs** tab set **Display** to "inline"

Donisius:
Closes #765

### DIFF
--- a/src/select/select.component.ts
+++ b/src/select/select.component.ts
@@ -63,8 +63,20 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from "@angular/forms";
 						class="bx--select__invalid-icon"
 						style="display: inherit;">
 					</ibm-icon-warning-filled16>
-					<ibm-icon-chevron-down16 *ngIf="!skeleton" class="bx--select__arrow" style="display: inherit;"></ibm-icon-chevron-down16>
-				</div>
+					<svg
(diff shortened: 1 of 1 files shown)


src/select/select.component.ts
```diff
@@ -63,8 +63,20 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from "@angular/forms";
...
+						style="will-change: transform;"
+						xmlns="http://www.w3.org/2000/svg"
+						class="bx--select__arrow"
+						width="10"
+						height="6"
+						viewBox="0 0 10 6"
+						aria-hidden="true">
+						<path d="M5 6L0 1 .7.3 5 4.6 9.3.3l.7.7z"></path>
+					</svg>
+					</div>
```
zvonimirfras: Indentation :)
```suggestion
				</div>
```

PR #820, opened 2019-10-09
feat(combobox): add invalid state for combobox

Donisius:


src/combobox/combobox.component.ts
```diff
@@ -86,6 +91,7 @@ import { filter } from "rxjs/operators";
 						ariaLabel="Close menu">
 					</ibm-icon-chevron-down16>
 			</div>
+			<div *ngIf="invalid" class="bx--form-requirement">A valid value is required</div>
```
Pablodotnet: invalidText input can be integrated to as in dropdowns or inputs to customize this invalid message

cal-smith: 💯 agreed, there should be an `invalidText` input like in `dropdown.component`

src/combobox/combobox.component.ts
```diff
@@ -39,6 +39,7 @@ import { filter } from "rxjs/operators";
 		</div>
 		<div
 			class="bx--combo-box bx--list-box"
+			[attr.data-invalid]="invalid ? false : null"
```
zvonimirfras: Shouldn't it be
```suggestion
			[attr.data-invalid]="(invalid ? true : null)"
```

src/combobox/combobox.component.ts
```diff
@@ -151,6 +158,14 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit {
...
 	 */
 	@Input() helperText: string | TemplateRef<any>;
+	/**
+	 * Set to `true` for invalid state.
+	 */
+	@Input() invalid = false;
+	/**
+	 * Value displayed if dropdown is in invalid state.
+	 */
+	@Input() invalidText = "";
```
zvonimirfras: We usually eventually need string or template options.
Any reason why that won't be the case here?

src/combobox/combobox.component.ts
Donisius: I agree, it should be like that.

src/combobox/combobox.component.ts
```diff
@@ -92,6 +97,8 @@ import { filter } from "rxjs/operators";
 				<ng-content></ng-content>
 			</div>
 		</div>
+
+		<div *ngIf="invalid" class="bx--form-requirement">{{ invalidText }}</div>
```
zvonimirfras: Template also needs a template outlet to be displayed :)

Combo box does it for example.

Donisius: Oh right 🙃 , I updated it to include the outlet and an example in the storybook

PR #821, opened 2019-10-09
feat: add basic support for angular ivy

cal-smith:
Closes IBM/carbon-components-angular#641

- `en.json` is now `en.ts` - the angular compiler doesn't package `.json` by default, and there's no reason it needs to be `.json` formatted for our purposes anyway
- fixed a bunch of circular imports
- fixed some `mkrdir` deploy errors
- renamed the `Modal` decorator to `ModalDecorator` ... we'll need to have a v4 to fully support ivy, so we can drop it then ... I don't believe there will be compat issues since the decorator _doesn't work anyway_. But definitely open to suggestions on this issue.

### DIFF
--- a/src/modal/modal.module.ts
+++ b/src/modal/modal.module.ts
@@ -17,7 +17,7 @@ import { PlaceholderModule } from "./../placeholder/placeholder.module";
 import { ExperimentalModule } from "./../experimental.module";
 
 // exports
-export { default as Modal } from "./modal.decorator";
+export { default as ModalDecorator } from "./modal.decorator";
 export { ModalService } from "./modal.service";
 export * from "./alert-modal.interface";
 export * from "./base-modal.class";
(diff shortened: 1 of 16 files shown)


src/modal/modal.module.ts
```diff
@@ -17,7 +17,7 @@ import { PlaceholderModule } from "./../placeholder/placeholder.module";
 import { ExperimentalModule } from "./../experimental.module";
 
 // exports
-export { default as Modal } from "./modal.decorator";
+export { default as ModalDecorator } from "./modal.decorator";
```
zvonimirfras: Wouldn't this be a breaking change for everybody importing `Modal`?

cal-smith: As far as I can tell it already was ... this just corrects the exports to match what the code was doing (there were two `Modal` exports, so I _think_ the named one takes precedence ... there's no way to `import { Modal }` and get the decorator right now)

cal-smith: It also won't compile unless we remove the duplicate export names, so that's a thing 😱

PR #831, opened 2019-10-11
feature(file-uploader): Add input to set button type

Pablodotnet:
Closes IBM/carbon-components-angular#830

There was no way to change the type of the button used on file uploader, an input has been added so devs can set the button type using the types available (primary, secondary, etc...).

#### Changelog

**New**

* Now is possible to set the type of button you want to use on file uploader, using `buttonType` input.

### DIFF
--- a/src/file-uploader/file-uploader.stories.ts
+++ b/src/file-uploader/file-uploader.stories.ts
@@ -25,6 +25,7 @@ import { NotificationService } from "../notification/notification.service";
 			[title]="title"
 			[description]="description"
 			[buttonText]="buttonText"
+			[buttonType]="buttonType"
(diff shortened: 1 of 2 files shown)


src/file-uploader/file-uploader.stories.ts
```diff
@@ -46,6 +47,7 @@ class FileUploaderStory {
 	@Input() title;
 	@Input() description;
 	@Input() buttonText;
+	@Input() buttonType = "primary";
```
zvonimirfras: It could use the typing.
```suggestion
	@Input() buttonType: "primary" | "secondary" | "tertiary" | "ghost" | "danger" = "primary";
```

PR #835, opened 2019-10-11
fix(file uploader): Prevent fileChanges event from getting fired on init

### LINKED ISSUE #733: File upload component - On the initial component load fileChanges event is getting fired
## Detailed description
On the initial component load also fileChanges event is getting fired. Ideally, it should fire on file change. I found in carbon component source code that on ngOnInit() we are emitting this event.
example: In file (https://angular.carbondesignsystem.com/documentation/components/FileUploader.html#source)
ngOnInit() {
       // overrides the undefined files value set by the user
       if (!this.files) {
           this.files = new Set();
           this.filesChange.emit(this.files);
       }
   }

I am using the below carbon version:
"carbon-components": "^10.5.1",
"carbon-components-angular": "^3.14.4"

Donisius:
Closes #733

### DIFF
--- a/src/file-uploader/file-uploader.component.ts
+++ b/src/file-uploader/file-uploader.component.ts
@@ -3,8 +3,7 @@ import {
 	Input,
 	Output,
 	ViewChild,
-	EventEmitter,
-	OnInit
+	EventEmitter
 } from "@angular/core";
 import { NG_VALUE_ACCESSOR } from "@angular/forms";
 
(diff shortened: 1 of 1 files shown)


src/file-uploader/file-uploader.component.ts
```diff
@@ -147,7 +147,6 @@ export class FileUploader implements OnInit {
 		// overrides the undefined files value set by the user
 		if (!this.files) {
 			this.files = new Set();
-			this.filesChange.emit(this.files);
```
zvonimirfras: Can `this.files` be initialized when declared on line 123 `@Input() files: Set<FileItem>;` as `@Input() files = new Set<FileItem>();` (or something similar)?
If everything still works as intended (it should), then remove `ngOnInit` completely.
I thought that was what we talked about.

PR #840, opened 2019-10-15
fix(date picker): Show DatePicker when calendar icon is clicked

### LINKED ISSUE #837: DatePicker Range component - DatePicker should also show when we click on the calendar icon inside the field
DatePicker is only showing when we click on the text Field. It should also show when we click on the calendar icon inside the field, On hover its showing pointer icon but on click nothing is happening.

Donisius:
Closes #837 

DatePicker could not be opened when the calendar icon is clicked. A toggle has been added to open and close the DatePicker when the calendar icon is clicked.

#### Changelog

**New**

* It is now possible to open and close the calendar when the calendar icon is clicked.

### DIFF
--- a/src/datepicker-input/datepicker-input.component.ts
+++ b/src/datepicker-input/datepicker-input.component.ts
@@ -36,9 +36,7 @@ import { NG_VALUE_ACCESSOR } from "@angular/forms";
 						[id]= "id"
 						[disabled]="disabled"
 						(change)="onChange($event)"/>
-						<ibm-icon-calendar16
-							class="bx--date-picker__icon"
-							(click)="onClick()">
+						<ibm-icon-calendar16 class="bx--date-picker__icon">
 						</ibm-icon-calendar16>
 				</div>
 				<div *ngIf="invalid" class="bx--form-requirement">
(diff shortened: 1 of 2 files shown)


src/datepicker-input/datepicker-input.component.ts
```diff
@@ -78,7 +78,7 @@ export class DatePickerInput {
 
 	@Output() valueChange: EventEmitter<string> = new EventEmitter();
 
-	@Output() toggleCalendar: EventEmitter<boolean> = new EventEmitter();
+	@Output() selected = new EventEmitter();
```
cal-smith: Not sure that the semantics of `selected` as used elsewhere in the codebase make sense here as well ... there's no value emitted for one thing

Donisius: Would something like openCalendar be better?

cal-smith: You may even just be able to forward the `click` event - that way all the cases are covered

cal-smith: Otherwise `open` or `openCalendar` would probably make more sense ... but you'd want to make sure they fire for all cases of clicking on the input

Donisius: Okay, that makes sense

cal-smith:
fyi @Donisius we'll need to add these changes for v2 as well 👍

Donisius:
@cal-smith it already works as expected on v2

cal-smith:
Oh awesome!

PR #853, opened 2019-10-21
fix(dropdown): emit empty array when last selected item is unselected

### LINKED ISSUE #823: Angular dropdown multi-select component not working as expected
## Detailed description

Using following version of carbon components:
`"carbon-components": "^10.6.2",
 "carbon-components-angular": "^3.16.3"`

Angular dropdown multi-select component does not fire an event when the last value is unchecked from a list of selections.

Bug

Angular dropdown multi-select component

Expected that on unselecting the last selection in a multi-select component, it fires an event.
Instead, no event was fired and hence the model always contain the last selection.

Bug is observed on all major browsers like Chrome, Firefox and Edge

`"carbon-components": "^10.6.2",
 "carbon-components-angular": "^3.16.3"`

As can be seen, I have selected two roles, and have printed the selection at top for convenience

image

Now, I remove role "REGULAR" by unselecting it and it works fine

image

I proceed to remove the remaining role "ADMINISTRATOR" but it is not removed! It is just unchecked from UI but the model is not changed (as no event is fired on unchecking the last remaining item)

image

Donisius:
Closes #823

### DIFF
--- a/src/dropdown/dropdown.component.ts
+++ b/src/dropdown/dropdown.component.ts
@@ -474,13 +474,13 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy, ControlVal
 			return;
 		}
 		let selected = this.view.getSelected();
-		if (selected && (!this.displayValue || !this.isRenderString())) {
+		if (selected.length && (!this.displayValue || !this.isRenderString())) {
 			if (this.type === "multi") {
 				return of(this.placeholder);
 			} else {
 				return of(selected[0].content);
 			}
-		} else if (selected && this.isRenderString()) {
+		} else if (selected.length && this.isRenderString()) {
 			return of(this.displayValue as string);
 		}
 		return of(this.placeholder);
--- a/src/dropdown/dropdown.component.ts
+++ b/src/dropdown/dropdown.component.ts
@@ -328,6 +328,9 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy, ControlVal
 			// only emit selected for "organic" selections
 			if (event && !event.isUpdate) {
 				this.selected.emit(event);
+			} else {
+				// emit an empty array when the last selected item in a multiselect is unselected
+				this.selected.emit([]);
 			}
 			this.appRef.tick();
 		});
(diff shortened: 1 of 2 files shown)


src/dropdown/dropdown.component.ts
```diff
@@ -328,6 +328,9 @@ export class Dropdown implements OnInit, AfterContentInit, OnDestroy, ControlVal
 			// only emit selected for "organic" selections
 			if (event && !event.isUpdate) {
 				this.selected.emit(event);
+			} else {
+				// emit an empty array when the last selected item in a multiselect is unselected
+				this.selected.emit([]);
```
cal-smith: This is going to end up emitting an empty array every time `select` emits, which could be quite a lot - `isUpdate` exists to let us switch between updates for programmatic reasons and updates due to user selection, we only want to emit `selected` when it's a user selection

src/dropdown/list/dropdown-list.component.ts
```diff
@@ -361,6 +361,9 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 	getSelected(): ListItem[] {
 		let selected = this.getListItems().filter(item => item.selected);
 		if (selected.length === 0) {
+			if (this.type === "multi") {
+				return [];
+			}
 			return null;
```
cal-smith: Instead of special casing this return, would it make sense to return `[]` in all cases? What else in the component would that affect?

Donisius: If we return [] in all cases, it gives the following error if the dropdown is not of type 'multi':

```
VM928 Dropdown.ngfactory.js:109 ERROR TypeError: Cannot read property 'content' of undefined
    at Dropdown.webpackHotUpdate../src/dropdown/dropdown.component.ts.Dropdown.getDisplayStringValue (dropdown.component.ts:483)
    at Object.eval [as updateRenderer] (VM710 Dropdown.ngfactory.js:113)
    at Object.debugUpdateRenderer [as updateRenderer] (core.js:11088)
    at checkAndUpdateView (core.js:10464)
    at callViewAction (core.js:10700)
    at execEmbeddedViewsAction (core.js:10663)
    at checkAndUpdateView (core.js:10460)
    at callViewAction (core.js:10700)
    at execComponentViewsAction (core.js:10642)
    at checkAndUpdateView (core.js:10465)
```

cal-smith: Let's fix those issues where the come up, and return `[]` for all cases ... I don't expect this to be a breaking change in practice

Donisius: Fixed!

PR #858, opened 2019-10-22
fix(date picker): update dates on UI on focus change

### LINKED ISSUE #850: DatePicker Component- If the user manually adding the dates it should also update on UI
if the user is manually adding dates than on blur or focus change it's not emitting any event, it's only emitting when the user has pressed the tab key. also it's not swapping the dates on UI.

Donisius:
Closes #850

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -330,29 +330,30 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 			// we can either set a date value or an empty string, so we start with an empty string
 			let singleDate = "";
 			// if date is a string, parse and format
-			if (typeof dates[0] === "string") {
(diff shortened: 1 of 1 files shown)


src/datepicker/datepicker.component.ts
```diff
@@ -330,29 +330,30 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 			// we can either set a date value or an empty string, so we start with an empty string
 			let singleDate = "";
 			// if date is a string, parse and format
-			if (typeof dates[0] === "string") {
-				singleDate = this.flatpickrInstance.parseDate(dates[0], this.dateFormat);
+			if (typeof this.flatpickrInstance.selectedDates[0] === "string") {
```
cal-smith: Ah, we use the dates from the flatpickr instance since they'll already be sorted correctly? Makes sense to me 👍

PR #860, opened 2019-10-22
fix(i18n): unsubscribe previous subscription before assigning a new subscription

### LINKED ISSUE #838: Slowdown in table component after 3.17.0
## Description

Significant performance slowdown has appeared after table component refactoring in 3.17.0. 

1. Create table with any data. 
2. Choose 4x CPU throttling in Chrome dev tools.
3. Click on the select row checkbox many times quickly.

You should see freezes and problems with interface responsiveness. On my 2012 MacBook Pro it can cause total page freeze. Moreover, if you just place table on your page and interact with other components presented on the page, you should see same problems.

There is no such problem in 3.16.3.

Thank you.

Donisius:
Closes #838

### DIFF
--- a/src/i18n/i18n.service.ts
+++ b/src/i18n/i18n.service.ts
@@ -89,6 +89,9 @@ export class Overridable {
 	 */
 	protected baseTranslation: Observable<string> = this.i18n.get(this.path);
 
+	/**
+	 * Subscription to the observable provided as an override (if any)
+	 */
 	protected subscription: Subscription;
 	/**
 	 * A boolean to flip between overridden and non-overridden states.
@@ -110,14 +113,16 @@ export class Overridable {
 	 */
 	override(value: string | Observable<string>) {
 		this.isOverridden = true;
+		// To ensure that there are not multiple subscriptions created for the same observable, we
+		// unsubscribe if a subscription already exists for an observable before creating a new one.
+		if (this.subscription) {
+			this.subscription.unsubscribe();
+			this.subscription = null;
+		}
+
 		this._value = value;
+
 		if (isObservable(value)) {
-			// To ensure that there are not multiple subscriptions created for the same observable, we
-			// unsubscribe if a subscription already exists for an observable before creating a new one.
-			if (this.subscription) {
-				this.subscription.unsubscribe();
-				this.subscription = null;
-			}
 			this.subscription = value.subscribe(v => {
 				this.$override.next(v);
 			});
--- a/src/i18n/i18n.service.ts
+++ b/src/i18n/i18n.service.ts
@@ -3,7 +3,8 @@ import {
 	BehaviorSubject,
 	Observable,
 	isObservable,
-	iif
+	iif,
+	Subscription
 } from "rxjs";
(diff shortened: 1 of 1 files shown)


src/i18n/i18n.service.ts
```diff
@@ -87,6 +88,8 @@ export class Overridable {
 	 * Our base non-overridden translation.
 	 */
 	protected baseTranslation: Observable<string> = this.i18n.get(this.path);
+
+	protected subscription: Subscription;
```
cal-smith: Could you add a one line comment explaining the purposes of this property?

Donisius: Two lines it is

cal-smith: Whatever works 😆

cal-smith: Also, maybe something like 
```suggestion
	/**
	 * Subscription to the observable provided as an override (if any)
	 */
	protected subscription: Subscription;
```
?

src/i18n/i18n.service.ts
```diff
@@ -107,9 +113,14 @@ export class Overridable {
 	 */
 	override(value: string | Observable<string>) {
 		this.isOverridden = true;
+		// To ensure that there are not multiple subscriptions created for the same observable, we
+		// unsubscribe if a subscription already exists for an observable before creating a new one.
+		if (this.subscription) {
```
zvonimirfras: Shouldn't this go inside of `if (isObservable(value)) {` block so we only unsubscribe if we'll subscribe after?

src/i18n/i18n.service.ts
```diff
@@ -107,9 +113,14 @@ export class Overridable {
...
+			this.subscription.unsubscribe();
```
zvonimirfras: `subscription.unsubscribe` doesn't appear to be setting the subscription to a falsy value so maybe
```suggestion
			this.subscription.unsubscribe();
			this.subscription = null;
```
to make sure we enter here repeatedly?

src/i18n/i18n.service.ts
cal-smith: What happens if we have a subscription but someone passes a raw value in (for whatever reason)?

zvonimirfras: Yup, makes sense.

src/i18n/i18n.service.ts
```diff
@@ -109,7 +112,13 @@ export class Overridable {
 		this.isOverridden = true;
 		this._value = value;
 		if (isObservable(value)) {
-			value.subscribe(v => {
+			// To ensure that there are not multiple subscriptions created for the same observable, we
+			// unsubscribe if a subscription already exists for an observable before creating a new one.
+			if (this.subscription) {
```
cal-smith: FYI @Donisius gotta put this back outside the `if (isObservable(value)) {`, sorry!

src/i18n/i18n.service.ts
cal-smith: This seems to still need done?

PR #870, opened 2019-10-25
feat(search): add output to listen to open and close event of toolbar search

### LINKED ISSUE #726: How to listen the open/close event of the 'toolbar search'?

Donisius:
Closes #726

### DIFF
--- a/src/search/search.component.ts
+++ b/src/search/search.component.ts
@@ -130,6 +130,7 @@ export class Search implements ControlValueAccessor {
 	 * Emits an event when value is changed.
 	 */
 	@Output() valueChange = new EventEmitter<string>();
+	@Output() opened = new EventEmitter<boolean>();
 	/**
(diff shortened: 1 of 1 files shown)


src/search/search.component.ts
```diff
@@ -130,6 +130,7 @@ export class Search implements ControlValueAccessor {
 	 * Emits an event when value is changed.
 	 */
 	@Output() valueChange = new EventEmitter<string>();
+	@Output() opened = new EventEmitter<boolean>();
```
zvonimirfras: Any reason why it can't be
```suggestion
	@Output() open = new EventEmitter<boolean>();
```
?

It seems like that would go more in line with our usual output naming.

cal-smith: Ah that's a good point ... I'm not super picky, but `open` may make more sense yeah

PR #887, opened 2019-11-05
fix(dropdown): Clear all selected options when number icon is clicked

### LINKED ISSUE #878: [IE]click the number icon should clear all selected options for ibm-dropdown
Use carbon-components-angular version 2.48.8. could not clear the selected options when click the number icon[IE11].

image

Donisius:
Closes #878


cal-smith:
@Donisius we should also check this against v3 and IE just in case ...

Donisius:
> @Donisius we should also check this against v3 and IE just in case ...

Yeah... there's issues with that as well

PR #890, opened 2019-11-05
fix(modal): scroll overflow content

youda97:
Content needs to be focusable to allow keyboard users the ability to scroll the content.

### DIFF
--- a/src/modal/alert-modal.component.ts
+++ b/src/modal/alert-modal.component.ts
@@ -1,6 +1,8 @@
 import {
 	Component,
-	Inject
+	Inject,
+	ViewChild,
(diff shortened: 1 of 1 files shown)


src/modal/alert-modal.component.ts
```diff
@@ -106,4 +111,9 @@ export class AlertModal extends BaseModal {
 		}
 		this.closeModal();
 	}
+
+	isOverflow() {
+		const element = this.content.nativeElement;
```
cal-smith: We should guard this for cases where the view may be delayed ... just a simple `if (!this.content) { return false; }` should be fine 👍

Issue #880, opened 2019-10-31
Selection tile - unable to make the selection using the key board


immanenipavankumar:
## Detailed description

> Describe in detail the issue you're having.

I am having an issue with using an ibm-selection-tile inside an ibm-tile-group. I am able to select a tile from group of selectable tiles by using the mouse but I am unable to make the selection using the key board (by hitting the SPACE or ENTER). This is an Accessibility concern and issue.

> Is this a feature request (new component, new icon), a bug, or a general issue?

Bug

> Is this issue related to a specific component?

selection-tile.component

> What did you expect to happen? What happened instead? What would you like to see changed?

I should be able to navigate through the group of selection tiles and make the selection using the key board but I am unable to make the selection using the key board (by hitting the SPACE or ENTER).

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

The issue seems to be happening in V3.0 since I tried it here and it didn't work
https://angular.carbondesignsystem.com/?path=/story/tiles--multi-select
The issue happens in carbon-angular v2.49.8 as well.

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

IBM ADI (Application Delivery Intelligence). Would like to get a fix as soon as possible to ship as part of our upcoming release.

PR #893, opened 2019-11-06
fix(selection tile): trigger selected event on tile-group on keyboard selection

Donisius:
issue #880 

When a selection is made using the keyboard, the selected event on the tile-group was not being triggered. Now the selected event gets triggered as expected.

### DIFF
--- a/src/tiles/selection-tile.component.ts
+++ b/src/tiles/selection-tile.component.ts
@@ -88,6 +88,7 @@ export class SelectionTile {
 	keyboardInput(event) {
 		if (event.key === "Enter" || event.key === "Spacebar" || event.key === " ") {
 			this.selected = !this.selected;
+			this.input.nativeElement.dispatchEvent(new Event("change"));
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/tiles/selection-tile.component.ts
```diff
@@ -88,6 +88,7 @@ export class SelectionTile {
 	keyboardInput(event) {
 		if (event.key === "Enter" || event.key === "Spacebar" || event.key === " ") {
 			this.selected = !this.selected;
+			this.input.nativeElement.dispatchEvent(new Event("change"));
```
cal-smith: What's the benefit of doing this over calling `this.change.emit`?

Donisius: Good point, it accomplishes the same thing in the end.

src/tiles/selection-tile.component.ts
```diff
@@ -88,6 +88,7 @@ export class SelectionTile {
 	keyboardInput(event) {
 		if (event.key === "Enter" || event.key === "Spacebar" || event.key === " ") {
 			this.selected = !this.selected;
+			this.change.emit();
```
cal-smith: `change` should emit the `event` :+1:

PR #894, opened 2019-11-06
fix(selection tile): add ability to make selections using keyboard

Donisius:


src/tiles/selection-tile.component.ts
```diff
@@ -82,6 +83,14 @@ export class SelectionTile {
 		SelectionTile.tileCount++;
 	}
 
+	@HostListener("keydown", ["$event"])
+	keyboardInput(event) {
+		if (event.key === "Enter" || event.key === "Spacebar" || event.key === " ") {
```
theiliad: We should probably use `event.keyCode` and `event.which` here too.

Although they are both deprecated, some older browsers will only support those fields.

`.key`
image


`.which`
image


`.keyCode`
image

cal-smith: Looks like `.key` is supported down to IE9 ... we only support back to IE11 ... we're also only concerned with the latest version of everything else, and Firefox ESR

theiliad: 👍

PR #895, opened 2019-11-06
fix(date picker): stop click propagation when calendar is clicked

### LINKED ISSUE #876: DatePicker is closing parent tooltip/dialog.
When I use a DatePicker inside a tooltip or dialog if I click any of the buttons to switch the month or the year it closes the calendar and the containing tooltip/dialog.

DatePickerDefect1

[Codesandbox](https://codesandbox.io/s/carbon-components-angular-pwqzh)

1. Click on the 'Trigger' to open the tooltip
2. Click the '>' to move to the next month

The tooltip closes.

## Additional information

Or you can also try this download from CodeSandbox [carbon-components-angular-simple-reproduce.zip](https://github.com/IBM/carbon-components-angular/files/3784969/carbon-components-angular-simple-reproduce.zip)

Donisius:
Closes #876 

Clicking on the flatpickr calendar causes the click event to propagate. It now stops the click propagation.

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -172,7 +172,10 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 		mode: "single",
 		dateFormat: "m/d/Y",
 		plugins: this.plugins,
-		onOpen: () => { this.updateClassNames(); },
+		onOpen: () => {
+			this.updateClassNames();
+			this.updateCalendarListeners();
+		},
 		value: this.value
 	};
 
@@ -276,6 +279,14 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 		this.flatpickrInstance.open();
 	}
 
+	protected updateCalendarListeners() {
+		const calendarContainer = document.querySelectorAll(".flatpickr-calendar");
+		Array.from(calendarContainer).forEach(calendar => {
+			calendar.removeEventListener("click", this.preventCalendarClose);
+			calendar.addEventListener("click", this.preventCalendarClose);
+		});
+	}
+
 	/**
 	 * Carbon uses a number of specific classnames for parts of the flatpickr - this idempotent method applies them if needed.
 	 */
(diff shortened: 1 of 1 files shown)


src/datepicker/datepicker.component.ts
```diff
@@ -271,7 +271,7 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 	 */
 	protected updateClassNames() {
 		if (!this.elementRef) { return; }
-
+		document.querySelector(".flatpickr-calendar").addEventListener("click", (event) => { event.stopPropagation(); });
```
youda97: Probably want to add the `.open` class to your query selector in case there are two date pickers

Donisius: good point!

src/datepicker/datepicker.component.ts
```diff
@@ -271,14 +271,17 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
...
 		// get all the possible flatpickrs in the document - we need to add classes to (potentially) all of them
 		const calendarContainer = document.querySelectorAll(".flatpickr-calendar");
 		const monthContainer = document.querySelectorAll(".flatpickr-month");
 		const weekdaysContainer = document.querySelectorAll(".flatpickr-weekdays");
 		const weekdayContainer = document.querySelectorAll(".flatpickr-weekday");
 		const daysContainer = document.querySelectorAll(".flatpickr-days");
 		const dayContainer = document.querySelectorAll(".flatpickr-day");
+		const openCalendarContainer = document.querySelectorAll(".flatpickr-calendar.open");
+
+		Array.from(openCalendarContainer)
```
cal-smith: I'm not sure we need to query for only `.open` calendars ... also there may be value in removing the event listener before adding it (since we may end up adding quite a few listeners over the lifetime of the component).

src/datepicker/datepicker.component.ts
```diff
@@ -280,6 +279,12 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 		const daysContainer = document.querySelectorAll(".flatpickr-days");
 		const dayContainer = document.querySelectorAll(".flatpickr-day");
 
+		Array.from(calendarContainer)
+			.forEach(calendar => {
+				calendar.removeEventListener("click", event => event.stopPropagation());
+				calendar.addEventListener("click", event => event.stopPropagation());
```
cal-smith: In order to remove it you'll have to add the listener as a property - something like `protected preventCalendarClose = event => event.stopPropagation();` and then pass that to add/remove listener

src/datepicker/datepicker.component.ts
```diff
@@ -284,6 +283,12 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 		const daysContainer = document.querySelectorAll(".flatpickr-days");
 		const dayContainer = document.querySelectorAll(".flatpickr-day");
 
+		Array.from(calendarContainer)
```
zvonimirfras: This doesn't belong to `updateClassNames`. It's not doing anything to class names and it's not necessary to update any class names.
Either function needs a better name or this needs to be moved out to a different function and used in a different way. Since it's a protected function and not used anywhere else, former might make more sense.

Donisius: That makes sense

PR #899, opened 2019-11-07
feature(dropdown-list): add ability to set output to emit scroll event

Pablodotnet:
Closes IBM/carbon-components-angular#398

Adds the ability to set an output to get the scroll event of the ibm-dropdown-list so that user can have the opportunity to set a handler.

#### Changelog

**New**

* scroll output set to ibm-dropdown-list container

#### Usage

The event received has the next structure:
`{ atTop: boolean, atBottom: boolean, event: Scroll }`

````
<ibm-dropdown
    [label]="label"
    [helperText]="helperText"
    [invalid]="invalid"
    [invalidText]="invalidText"
    [theme]="theme"
    placeholder="Select"
    [disabled]="disabled"
    (selected)="selected($event)"
    (onClose)="onClose($event)">
        <ibm-dropdown-list [items]="items" (scroll)="handleScroll($event)"></ibm-dropdown-list>
</ibm-dropdown>
````

### DIFF
--- a/src/dropdown/list/scroll-custom-event.interface.ts
+++ b/src/dropdown/list/scroll-custom-event.interface.ts
@@ -1,5 +1,3 @@
-import { Scroll } from "@angular/router";
-
 /**
  * A custom structure that represents a custom event to emit when scroll.
  *
@@ -7,7 +5,7 @@ import { Scroll } from "@angular/router";
  * export interface ScrollCustomEvent {
  * 	atTop: boolean;
  * 	atBottom: boolean;
- * 	event: Scroll;
+ * 	event: Event;
  * }
  * ```
  */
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/dropdown/list/scroll-custom-event.interface.ts [hunk 1]
```diff
@@ -0,0 +1,27 @@
+import { Scroll } from "@angular/router";
```
cal-smith: I don't think we want to use this event here 🤔 I think we can just use the generic `Event` interface (based on this that seems to be correct: https://developer.mozilla.org/en-US/docs/Web/API/Document/scroll_event there's no speicifc `ScrollEvent`)

Pablodotnet: Ok, changing :D

PR #901, opened 2019-11-08
fix(combobox): add check to avoid error when empty array

Pablodotnet:
Closes IBM/carbon-components-angular#900

When selected is an empty array it causes an error of content of undefined, a check has been added to be sure it has the used element of the array

#### Changelog

**Changed**

* updated expression to be sure first place of array is defined


JehanX:
@zvonimirfras Can you please approve and merge this pr? Thanks

PR #904, opened 2019-11-11
Breadcrumb

stanislavgeorgiev:
Closes IBM/carbon-components-angular#902

@cal-smith @zvonimirfras I need help fixing the tests please. I'm getting the error below and I cannot figure out how to fix it. I figured as much that if I revert the changes from the `breadcrumb-item.component.ts` then the tests are passing but I just cannot connect why would they fail with the small update.
```
HeadlessChrome 0.0.0 (Mac OS X 10.15.1)
Breadcrumb should create a breadcrumb with five items and 
an overflow menu in second position FAILED
        Error: Expected 'a' to equal 'ibm-overflow-menu'.
```

### DIFF
--- a/src/breadcrumb/breadcrumb-item.component.ts
+++ b/src/breadcrumb/breadcrumb-item.component.ts
@@ -62,6 +62,10 @@ export class BreadcrumbItemComponent {
 
 	constructor(protected domSanitizer: DomSanitizer, @Optional() protected router: Router) { }
 
+	useTemplate() {
+		return this.skeleton || this._href || this.route;
+	}
+
 	navigate(event) {
 		if (this.router && this.route) {
 			event.preventDefault();
--- a/src/breadcrumb/breadcrumb-item.component.ts
+++ b/src/breadcrumb/breadcrumb-item.component.ts
@@ -17,7 +17,7 @@ import { Router } from "@angular/router";
 		[href]="(skeleton ? '/#' : href)"
 		(click)="navigate($event)"
 		[attr.aria-current]="(current ? ariaCurrent : null)"
-		*ngIf="skeleton || href; else content">
+		*ngIf="useTemplate(); else content">
 		<ng-container *ngTemplateOutlet="content"></ng-container>
 	</a>
 	<ng-template #content>
--- a/src/breadcrumb/breadcrumb-item.component.ts
+++ b/src/breadcrumb/breadcrumb-item.component.ts
@@ -58,7 +58,7 @@ export class BreadcrumbItemComponent {
 
 	@HostBinding("class.bx--breadcrumb-item") itemClass = true;
 
-	protected _href = "javascript:void(0)";
+	protected _href;
 
 	constructor(protected domSanitizer: DomSanitizer, @Optional() protected router: Router) { }
 
--- a/src/breadcrumb/breadcrumb-item.component.ts
+++ b/src/breadcrumb/breadcrumb-item.component.ts
@@ -14,7 +14,7 @@ import { Router } from "@angular/router";
 	template: `
 	<a
 		class="bx--link"
-		href="{{skeleton ? '/#' : href}}"
+		[href]="(skeleton ? '/#' : href)"
 		(click)="navigate($event)"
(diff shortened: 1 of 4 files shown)


src/breadcrumb/breadcrumb-item.component.ts
```diff
@@ -1,15 +1,21 @@
...
 } from "@angular/core";
+import { DomSanitizer } from "@angular/platform-browser";
+import { Router } from "@angular/router";
 
 @Component({
 	selector: "ibm-breadcrumb-item",
 	template: `
 	<a
 		class="bx--link"
 		href="{{skeleton ? '/#' : href}}"
```
cal-smith: There's some warns around this
```suggestion
		[href]="(skeleton ? '/#' : href)"
```

src/breadcrumb/breadcrumb-item.component.ts
```diff
@@ -1,15 +1,21 @@
...
+		(click)="navigate($event)"
 		[attr.aria-current]="(current ? ariaCurrent : null)"
 		*ngIf="skeleton || href; else content">
```
cal-smith:
```suggestion
		*ngIf="useTemplate(); else content">
```

src/breadcrumb/breadcrumb-item.component.ts
```diff
@@ -28,4 +57,16 @@ export class BreadcrumbItemComponent {
 	@HostBinding("class.bx--breadcrumb-item--current") @Input() current = false;
```
cal-smith: This can be undefined, so we preserve the old behaviour
```suggestion
	protected _href;
```

src/breadcrumb/breadcrumb-item.component.ts
```diff
@@ -28,4 +57,16 @@ export class BreadcrumbItemComponent {
...
 
 	@HostBinding("class.bx--breadcrumb-item") itemClass = true;
+
+	protected _href = "javascript:void(0)";
```
cal-smith: Check if we are in a skeleton, have a href, or have a route

```suggestion
	useTemplate() {
		return this.skeleton || this._href || this.route;
	}
```

src/breadcrumb/breadcrumb-item.component.ts
cal-smith: This note should be with the `protected _href = "";` and not replacing the `@HostBinding`

src/breadcrumb/breadcrumb-item.component.ts
cal-smith: This also isn't attached to the right place....

src/breadcrumb/breadcrumb-item.component.ts
```diff
@@ -28,4 +57,16 @@ export class BreadcrumbItemComponent {
...
+
+	constructor(protected domSanitizer: DomSanitizer, @Optional() protected router: Router) { }
+
```
cal-smith: Check if we are in a skeleton, have a href, or have a route

```suggestion

	useTemplate() {
		return this.skeleton || this._href || this.route;
	}
```

src/breadcrumb/breadcrumb-item.component.ts
cal-smith: This can be undefined, so we preserve the old behaviour
```suggestion
	protected _href;
```

stanislavgeorgiev:
Thanks for the help @cal-smith! Tests are now passing!

src/breadcrumb/breadcrumb-item.interface.ts
```diff
@@ -12,6 +12,16 @@ export interface BreadcrumbItem {
 	 * Href for the breadcrumb item.
 	 */
 	href: string;
```
cal-smith: Optional suggest, since we have the `route` key now, `href` should be optional

```suggestion
	href?: string;
```

stanislavgeorgiev: Yea, I was thinking about the same thing but wasn't sure about it

PR #907, opened 2019-11-12
fix(date picker): position flatpickr calendar on scroll

### LINKED ISSUE #905: Calendar position when scrolling modal.
## Calendar position when scrolling modal.

When using date picker in a modal and you open the calendar and scroll down or up, the calendar maintains open and does not scroll with the picker, as in the below screenshots.

## Screenshots

Screen Shot 2019-11-11 at 11 27 04 AM

Screen Shot 2019-11-11 at 11 27 26 AM

Donisius:
Closes #905


src/datepicker/datepicker.component.ts
```diff
@@ -211,6 +212,13 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 		this.onTouched();
 	}
 
+	@HostListener("document:scroll")
```
cal-smith: This is only going to catch the document scrolling, which won't catch the modal scrolling - see here for a demo https://stackblitz.com/edit/typescript-hvdqcp?file=index.ts

You should be able to use `scrollableParentsObservable` from `utils/scroll`

src/datepicker/datepicker.component.ts
```diff
@@ -176,8 +179,21 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 
 	protected flatpickrInstance = null;
 
+	protected scrollSubscription: Subscription = new Subscription();
+
 	constructor(protected elementRef: ElementRef) { }
 
+	ngOnInit() {
```
cal-smith: Should be `AfterViewInit` ... the view isn't guaranteed to be instantiated in `OnInit`

src/datepicker/datepicker.component.ts
```diff
@@ -196,6 +212,7 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 	ngAfterViewChecked() {
 		if (!this.isFlatpickrLoaded()) {
 			this.flatpickrInstance = flatpickr(`#${this.id}`, this.flatpickrOptions);
+			this.flatpickrInstance._positionCalendar();
```
cal-smith: Do we need this?

Donisius: The first time you open the calendar in the modal it is positioned slightly off:
Screen Shot 2019-11-13 at 12 35 17 PM
Compare with:
Screen Shot 2019-11-13 at 12 35 31 PM
Using positionCalendar here fixes that.

src/datepicker/datepicker.component.ts
Donisius: Done!

src/datepicker/datepicker.component.ts
cal-smith: Interesting ... is this something we could hook into the `flatpickrBaseOptions.onOpen` callback?

Donisius: I can't seem to reproduce this anymore, might be some weird margins affecting the positioning before

src/datepicker/datepicker.component.ts
```diff
@@ -190,13 +195,23 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
...
 	}
 
+	ngAfterViewInit() {
+		const scrollObservable = scrollableParentsObservable(this.elementRef.nativeElement);
+		this.scrollSubscription = scrollObservable.subscribe((event: any) => {
+			if (!isVisibleInContainer(this.elementRef.nativeElement, event.target)) {
+				this.flatpickrInstance.close();
+			} else {
+				this.flatpickrInstance._positionCalendar();
+			}
```
cal-smith: Sorry, one last thing, I think we'll need to check `isFlatpickrLoaded` ... these unchecked calls may error

Donisius: done!

src/datepicker/datepicker.component.ts
```diff
@@ -203,13 +208,25 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 		}
 	}
 
+	ngAfterViewInit() {
+		const scrollObservable = scrollableParentsObservable(this.elementRef.nativeElement);
+		this.scrollSubscription = scrollObservable.subscribe((event: any) => {
```
cal-smith: Just so it doesn't get lost, this should use `ElementService` 👍

Donisius: Working on that now 👍

cal-smith: Awesome 💯

src/datepicker/datepicker.component.spec.ts
```diff
@@ -41,7 +43,8 @@ describe("DatePicker", () => {
 			imports: [
 				Calendar16Module,
 				FormsModule
-			]
+			],
+			providers: [ElementService, AnimationFrameServiceSingleton]
```
cal-smith: You should be able to just import the `UtilsModule` from `utils/` and just use that :+1:

src/datepicker/datepicker.component.ts
```diff
@@ -199,7 +202,9 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 
 	protected flatpickrInstance = null;
 
-	constructor(protected elementRef: ElementRef) { }
+	protected visibilitySubscription = new Subscription();
+
+	constructor(protected elementRef: ElementRef, protected elementService: ElementService = null) { }
```
cal-smith:
```suggestion
	constructor(protected elementRef: ElementRef, protected elementService: ElementService) { }
```

You shouldn't need to null out the service 🤔

src/datepicker/datepicker.component.ts
```diff
@@ -213,13 +218,27 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 		}
 	}
 
+	ngAfterViewInit() {
+		setTimeout(() => {
+			if (this.isFlatpickrLoaded() && this.elementService) {
```
cal-smith: `this.elementService` should be available right away ... there's no subclasses of datepicker and it's not optional so we should be fine to just check if it's loaded

cal-smith: The loaded check should also likely go inside the subscribe ... we can't know that `isFlatpickrLoaded` will be true by the time the timeout and `AfterViewInit` run

src/datepicker/datepicker.module.ts
```diff
@@ -14,6 +16,7 @@ import { DatePicker } from "./datepicker.component";
 	imports: [
 		CommonModule,
 		DatePickerInputModule
-	]
+	],
+	providers: [ElementService, AnimationFrameServiceSingleton]
```
cal-smith: Same comment as the test, just import `UtilsModule` 👍

src/datepicker/datepicker.component.ts
```diff
@@ -213,13 +218,25 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 		}
 	}
 
+	ngAfterViewInit() {
+		this.visibilitySubscription = this.elementService
+			.visibility(this.elementRef.nativeElement, this.elementRef.nativeElement)
+			.subscribe(value => {
+				if (this.isFlatpickrLoaded()) {
```
cal-smith: I think you'll want to do something like 
```suggestion
				if (this.isFlatpickrLoaded() && this.flatpickrInstance.isOpen) {
```

I'm not 100% sure of the property names though 👍

PR #915, opened 2019-11-15
feature(combobox): add output to emit the searching text

Pablodotnet:
Closes IBM/carbon-components-angular#914

This PRs adds an output so the developer can set a handler to get the text that is being input to the combobox.

#### Changelog

**New**

* Add output to emit the searched value.


Pablodotnet:
thought setting keyup directly to ibm-combobox was not working as expected but it is, so this is not needed.

PR #921, opened 2019-11-18
feat: update position utility and improve resize detection

cal-smith:
- updates `@carbon/utils-position` to `1.1.2`
- adds `AnimationFrameService` and `AnimationFrameServiceSingleton` so we have a consistent and effeciant way of hooking into `requestAnimationFrame`
- add `ElementService` to simplify visibility checks (and more in future)
- refactor `dropdown`, `dropdown.service`, and `dialog` to use `ElementService` and `AnimationFrameService`


src/utils/scroll.ts
```diff
@@ -30,6 +30,17 @@ export const isVisibleInContainer = (element: HTMLElement, container: HTMLElemen
 	return elementRect.bottom <= containerRect.bottom && elementRect.top >= containerRect.top;
 };
 
+export const getScrollableParents = (node: HTMLElement) => {
+	const elements = [document.body];
```
zvonimirfras: Check if we're in the browser to use `document`?

src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -30,8 +31,12 @@ import { ExperimentalService } from "./../../experimental.module";
 	`
 })
 export class OverflowMenuPane extends Dialog implements AfterViewInit {
-	constructor(protected elementRef: ElementRef, protected i18n: I18n, protected experimental: ExperimentalService) {
-		super(elementRef);
+	constructor(
+		protected elementRef: ElementRef,
+		protected i18n: I18n,
+		protected experimental: ExperimentalService,
+		protected elementService: ElementService) {
```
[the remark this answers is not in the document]
cal-smith: Sure 👍

src/utils/scroll.ts
cal-smith: Shouldn't be necessary given that we only query when the components have live DOM ... Really we need a proper strategy to make the components SSR safe, as-is there's a few that won't even initialize

src/utils/element.service.ts
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	visibility(target: HTMLElement, parentElement: HTMLElement = target): Observable<{ visible: boolean, frame: number }> {
+		const scrollableParents = getScrollableParents(parentElement);
+		return this.tick.pipe(map(frame => {
+			for (const parent of scrollableParents) {
+				if (!isVisibleInContainer(target, parent)) {
+					return {
+						visible: false,
+						frame
```
Donisius: I don't see a reason why we need to return the frame along with visibility.

src/dialog/dialog.component.ts
```diff
@@ -63,13 +64,15 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
 	/**
 	 * `Subscription` used to update placement in the event of a window resize.
 	 */
-	protected resizeSubscription: Subscription;
+	protected resizeSubscription: Subscription = new Subscription();
```
theiliad: you can remove `: Subscription` now that the type is implicitly defined through `new Subscription()`

src/dialog/dialog.component.ts
```diff
@@ -63,13 +64,15 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
...
 	/**
 	 * Subscription to all the scrollable parents `scroll` event
 	 */
 	// add a new subscription temporarily so that contexts (such as tests)
 	// that don't run ngAfterViewInit have something to unsubscribe in ngOnDestroy
 	protected scrollSubscription: Subscription = new Subscription();
```
theiliad: remove `: Subscription`
```suggestion
	protected scrollSubscription = new Subscription();
```

src/dialog/dialog.component.ts
theiliad:
```suggestion
	protected resizeSubscription = new Subscription();
```

src/dialog/dialog.component.ts
```diff
@@ -101,9 +108,12 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
...
 		this.data = this.dialogConfig.data;
 
-		this.resizeSubscription = Dialog.resizeObservable.subscribe(() => {
-			this.placeDialog();
-		});
+		// fallback if elementService isn't available
+		if (!this.elementService) {
+			this.resizeSubscription = Dialog.resizeObservable.subscribe(() => {
+				this.placeDialog();
+			});
```
theiliad: Why not do:
```js
this.resizeSubscription = Dialog.resizeObservable.subscribe(this.placeDialog);
```

src/dropdown/dropdown.service.ts
```diff
@@ -21,14 +21,17 @@ export class DropdownService {
 	protected menuInstance: HTMLElement;
 
 	/**
-	 * Maintains an Event Observable Subscription for tracking window resizes.
-	 * Window resizing is tracked if the `Dropdown` is appended to the body, otherwise it does not need to be supported.
+	 * Maintains an Event Observable Subscription for the global requestAnimationFrame.
+	 * requestAnimationFrame is tracked only if the `Dropdown` is appended to the body otherwise we don't need it
 	 */
-	protected resize: Subscription;
+	protected animationFrameSubscription: Subscription = new Subscription();
```
theiliad:
```suggestion
	protected animationFrameSubscription = new Subscription();
```

src/dropdown/dropdown.service.ts
```diff
@@ -56,10 +59,9 @@ export class DropdownService {
 
 		this.menuInstance = dropdownWrapper;
 
-		this.positionDropdown(parentRef, dropdownWrapper);
-		this.resize = fromEvent(window, "resize")
-			.pipe(throttleTime(100))
-			.subscribe(() => this.positionDropdown(parentRef, dropdownWrapper));
+		this.animationFrameSubscription = this.animationFrameService.tick.subscribe(() => {
+			this.positionDropdown(parentRef, dropdownWrapper);
+		});
```
theiliad: What about?
```js
this.animationFrameSubscription = this.animationFrameService.tick
   .subscribe(() => this.positionDropdown(parentRef, dropdownWrapper));
```

src/dialog/dialog.component.ts
cal-smith: I don't _think_ we'll run into scoping issues with that ... I'll give it a try 👍

src/dropdown/dropdown.service.ts
cal-smith: More or less the same thing ... Currently it's marginally more readable to my eyes

PR #923, opened 2019-11-19
feat(table): add stickyHeader support

### LINKED ISSUE #917: Table: Bring back the sticky headers from Neutrino
## Detailed description

I'd like to limit the height of the table and have scrollable content with fixed table headers. I found that old issue which is about the same thing: #435 

@cal-smith mentioned that the React library has that support and can be mirrored in the Angular library too.

Feature request

Table

3.26

IBM ADDI

Donisius:
Closes #917

### DIFF
--- a/src/table/body/table-row.component.ts
+++ b/src/table/body/table-row.component.ts
@@ -19,6 +19,7 @@ import { Observable } from "rxjs";
 			<td
 				*ngIf="model.hasExpandableRows()"
 				ibmTableExpandButton
+				class="bx--table-expand-v2"
 				[expanded]="expanded"
 				[expandable]="expandable"
 				[skeleton]="skeleton"
@@ -28,6 +29,7 @@ import { Observable } from "rxjs";
 			<td
 				*ngIf="!skeleton && showSelectionColumn && !enableSingleSelect"
 				ibmTableCheckbox
+				class="bx--table-column-checkbox"
 				[size]="size"
 				[selected]="selected"
 				[label]="getCheckboxLabel()"
--- a/src/table/body/table-row.component.ts
+++ b/src/table/body/table-row.component.ts
@@ -24,6 +24,7 @@ import { Observable } from "rxjs";
 				[expandable]="expandable"
 				[skeleton]="skeleton"
 				[ariaLabel]="getExpandButtonAriaLabel()"
+				[headers]="model.getHeaderId('expand')"
 				(expandRow)="expandRow.emit()">
 			</td>
 			<td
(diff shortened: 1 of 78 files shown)


src/table/body/table-row.component.ts
```diff
@@ -19,6 +19,7 @@ import { Observable } from "rxjs";
 			<td
 				*ngIf="model.hasExpandableRows()"
 				ibmTableExpandButton
+				class="bx--table-expand-v2"
```
theiliad: do we really want to use a version number in the classname?

Donisius: It's the class name defined in the carbon components css. I needed to use the v2 class because there was styling that the css ignores if it is v2 but doesn't ignore if its the regular bx--table-expand.

theiliad: 🤦‍♂ I guess we can't do much there then

src/table/head/table-head.component.ts
```diff
@@ -24,7 +26,8 @@ import { Observable } from "rxjs";
 	template: `
 	<ng-container *ngIf="model">
 		<tr>
-			<th ibmTableHeadExpand *ngIf="model.hasExpandableRows()"></th>
+			<th ibmTableHeadExpand *ngIf="model.hasExpandableRows() && !stickyHeader"></th>
+			<th ibmTableHeadExpand *ngIf="model.hasExpandableRows() && stickyHeader" class="bx--table-expand-v2" style="padding-left: 2.5rem"></th>
```
cal-smith: Could the class be applied conditionally with an `ngClass`? Also We may want to move the `style` to the `styles` property on the component if we need it (what does react do for that case?)

Donisius: Their expanded for does this:
Screen Shot 2019-11-25 at 1 49 48 PM

Donisius:
> image
> 
> There's something weird going on with the styling and offsets 🤔

We tried to compensate for that by offsetting it based on the width of the scroll bar width but the computed scroll bar width was always off.

cal-smith:
There also seems to be a conflict with the table head component

Donisius:
> Solve conflicts please :)

Done!

src/table/stories/app-expansion-table.component.ts
```diff
@@ -40,6 +40,10 @@ class CustomHeaderItem extends TableHeaderItem {
 			[model]="model"
 			[size]="size"
 			[showSelectionColumn]="showSelectionColumn"
+<<<<<<< HEAD
+			[stickyHeader]="stickyHeader"
+=======
+>>>>>>> 28598a2a185bd578990271ea080d2250f321ed33
```
cal-smith: I don't think you meant to commit this ...

Donisius: .........I was just testing you 🙃

cal-smith: 😂

PR #924, opened 2019-11-19
feat(table): show blue action bar only when actions are defined

### LINKED ISSUE #919: Table: Don't show the blue actions bar in the table header if there are no actions defined
## Detailed description

We have a use case where we don't want to have an actions bar showing up when one or more rows have been selected in the table. We haven't defined an actions bar in the table header but it still shows up. We'd love if the table can either automatically "sense" that lack of selected actions and don't show the actions bar (preferred) or have a property to toggle.

Screen Shot 2019-11-18 at 3 09 15 PM

Feature request

Table

3.26

IBM ADDI

Donisius:
Closes #919

### DIFF
--- a/src/table/toolbar/table-toolbar.component.ts
+++ b/src/table/toolbar/table-toolbar.component.ts
@@ -42,7 +42,7 @@ import { I18n, Overridable } from "../../i18n/i18n.module";
 	template: `
 	<section class="bx--table-toolbar">
 		<div
-			*ngIf="showToolbarActions"
+			*ngIf="showActions && model"
 			class="bx--batch-actions"
 			[ngClass]="{
 				'bx--batch-actions--active': selected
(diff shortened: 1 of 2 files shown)


src/table/toolbar/table-toolbar.component.ts
```diff
@@ -70,13 +71,21 @@ export class TableToolbar {
 	@Input() set cancelText(value: { CANCEL: string }) {
 		this._cancelText.override(value.CANCEL);
 	}
+	@Input() showToolbarActions = true;
```
cal-smith: If we keep this ...

```suggestion
	@Input() showActions = true;
```

Since we already know it's a toolbar 🙂

src/table/toolbar/table-toolbar.component.ts
```diff
@@ -42,6 +42,7 @@ import { I18n, Overridable } from "../../i18n/i18n.module";
 	template: `
 	<section class="bx--table-toolbar">
 		<div
+			*ngIf="showToolbarActions"
```
cal-smith: What if we toggle it based on the existence of the model? If there's no model then there's no need to show the actions at all ... 🤔

PR #927, opened 2019-11-21
feat(File Uploader): Add drag and drop option

Donisius:

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "cz-conventional-changelog-components": "1.0.0",
     "extract-text-webpack-plugin": "3.0.2",
     "file-loader": "3.0.1",
+    "file-type": "12.4.0",
     "gulp": "4.0.0",
     "gulp-concat": "2.6.1",
     "gulp-tap": "1.0.1",
--- a/src/file-uploader/file-uploader.component.ts
+++ b/src/file-uploader/file-uploader.component.ts
@@ -154,8 +154,6 @@ export class FileUploader {
 	 */
 	protected dragOver = false;
 
-	protected _fileList = [];
-
 	constructor(protected i18n: I18n) {
 		FileUploader.fileUploaderCount++;
 	}
@@ -177,15 +175,8 @@ export class FileUploader {
 		this.onTouchedCallback();
 	}
 
-	set fileList(files: Array<any>) {
-		this._fileList = files;
-	}
-
 	get fileList() {
-		if (this.fileInput.nativeElement.files.length) {
-			return Array.from(this.fileInput.nativeElement.files);
-		}
-		return this._fileList;
+		return Array.from(this.fileInput.nativeElement.files);
 	}
 
 	/**
@@ -197,18 +188,22 @@ export class FileUploader {
 		}
 	}
 
+	createFileItem(file): FileItem {
+		return {
+			uploaded: false,
+			state: "edit",
+			invalid: false,
+			invalidText: "",
+			file: file
+		};
+	}
+
 	onFilesAdded() {
 		if (!this.multiple) {
 			this.files.clear();
 		}
 		for (let file of this.fileList) {
-			const fileItem: FileItem = {
-				uploaded: false,
-				state: "edit",
-				invalid: false,
-				invalidText: "",
-				file: file
-			};
+			const fileItem = this.createFileItem(file);
 			this.files.add(fileItem);
 			this.filesChange.emit(this.files);
 		}
@@ -235,29 +230,30 @@ export class FileUploader {
 		const transferredFiles = Array.from(event.dataTransfer.files);
 
 		if (!this.accept.length) {
-			this.fileList = transferredFiles;
-			this.onFilesAdded();
+			transferredFiles.forEach(file => {
+				const fileItem = this.createFileItem(file);
+				this.files.add(fileItem);
+				this.filesChange.emit(this.files);
+			});
+
+			this.value = this.files;
 			this.dragOver = false;
 			return;
 		}
 
-		const fileExtensionRegExp = new RegExp(/\.[0-9a-z]+$/, "i");
-
-		const acceptedFiles = transferredFiles.filter(({ name, type }) => {
-			if (!fileExtensionRegExp.test(name)) {
-				return false;
-			}
-			const [fileExtension] = name.match(fileExtensionRegExp);
+		transferredFiles.filter(({ name, type }) => {
+			// Get the file extension and add a "." to the beginning.
+			const fileExtension = name.split(".").pop().replace(/^/, ".");
+			// Check if the accept array contains the mime type or extension of the file.
 			return this.accept.includes(type) || this.accept.includes(fileExtension);
+		}).forEach(file => {
+			const fileItem = this.createFileItem(file);
+			this.files.add(fileItem);
+			this.filesChange.emit(this.files);
 		});
 
-		this.fileList = acceptedFiles;
-		this.onFilesAdded();
+		this.value = this.files;
 		this.dragOver = false;
-
-		// Emit the files for the possibility of further filtration.
-		// See storybook for an example.
-		this.filesChange.emit(this.files);
 	}
 
 	removeFile(fileItem) {
--- a/src/file-uploader/file-uploader.component.ts
+++ b/src/file-uploader/file-uploader.component.ts
@@ -35,7 +35,8 @@ const noop = () => {};
 						(dragover)="onDragOver($event)"
 						(dragleave)="onDragLeave($event)"
 						(drop)="onDrop($event)">
-						{{ dropText }}
+						<ng-container *ngIf="!isTemplate(dropText)">{{dropText}}</ng-container>
+						<ng-template *ngIf="isTemplate(dropText)" [ngTemplateOutlet]="dropText"></ng-template>
 					</div>
 				</label>
 				<button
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -105,6 +105,7 @@
     "element-closest-polyfill": "1.0.0",
     "extract-text-webpack-plugin": "3.0.2",
     "file-loader": "3.0.1",
+    "file-type": "12.4.0",
```
theiliad: this seems like it'd be a `dependency` not a `devDependency`

src/file-uploader/file-uploader.stories.ts
```diff
@@ -84,6 +86,119 @@ class FileUploaderStory {
...
+					if (type) {
+						fileExtension = type.ext.replace(/^/, ".");
+						mime = type.mime;
+					} else {
+						// If a file type can not be determined using magic numbers
+						// then use file extension or mime type.
+						[fileExtension] = fileObj.file.name.match(fileExtensionRegExp);
+						mime = fileObj.file.type;
+					}
+					resolve({file: fileObj, accept: (this.accept.includes(fileExtension) || this.accept.includes(mime))});
```
theiliad: can we break this object into multiple lines?

src/file-uploader/file-uploader.stories.ts
```diff
@@ -84,6 +86,119 @@ class FileUploaderStory {
...
+				};
+			});
+		};
+
+		transferredFiles.forEach(file => {
+			promises.push(readFileAndAddToMap(file));
+		});
+
+		Promise.all(promises).then(filePromiseArray =>
+			filePromiseArray.filter(filePromise => filePromise.accept).map(acceptedFile => acceptedFile.file))
```
theiliad: should `.map` move onto the next line?

theiliad:
Also the spinner positioning seems to be a bit off
image

package.json
Donisius: file-type is only used in the storybook to show an example of a more extensive test for file-type checking after the preliminary checking done in the component.

theiliad: oh ok

src/file-uploader/file-uploader.stories.ts
```diff
@@ -168,9 +171,10 @@ class DragAndDropStory {
 		});
 
 		Promise.all(promises).then(filePromiseArray =>
-			filePromiseArray.filter(filePromise => filePromise.accept).map(acceptedFile => acceptedFile.file))
-				.then(acceptedFiles => this.files = new Set(acceptedFiles))
-					.catch(error => console.log(error));
+			filePromiseArray.filter(filePromise => filePromise.accept)
+				.map(acceptedFile => acceptedFile.file))
+					.then(acceptedFiles => this.files = new Set(acceptedFiles))
+						.catch(error => console.log(error));
```
theiliad: Do we need more indentation for `then` and `catch`?

src/file-uploader/file-uploader.component.ts
```diff
@@ -24,7 +25,21 @@ const noop = () => {};
...
+					<div
+						class="bx--file__drop-container"
+						[ngClass]="{'bx--file__drop-container--drag-over': dragOver}"
+						role="button"
+						(click)="fileInput.click()"
+						[attr.for]="fileUploaderId"
+						(dragover)="onDragOver($event)"
+						(dragleave)="onDragLeave($event)"
+						(drop)="onDrop($event)">
+						{{ dropText }}
```
cal-smith: The `dropText` says it takes a `TemplateRef` but there's no output for it here?

src/file-uploader/file-uploader.component.ts
```diff
@@ -122,9 +145,16 @@ export class FileUploader {
 
 	@Output() filesChange = new EventEmitter<any>();
 
+	@Output() dropped = new EventEmitter<any>();
```
cal-smith: Do we need a separate event for dropping files? Or can we just use `filesChange`?

src/file-uploader/file-uploader.component.ts
```diff
@@ -155,12 +189,11 @@ export class FileUploader {
...
 	}
 
-	onFilesAdded() {
-		const files = this.fileInput.nativeElement.files;
+	onFilesAdded(fileList) {
 		if (!this.multiple) {
 			this.files.clear();
 		}
-		for (let file of files) {
+		for (let file of fileList) {
```
cal-smith: Can we just use `this.fileList` directly here instead of passing it through the template first?

src/file-uploader/file-uploader.stories.ts
```diff
@@ -18,6 +18,8 @@ import {
 } from "../";
 import { NotificationService } from "../notification/notification.service";
 
+const fileType = require("file-type");
```
cal-smith: Can we `import { fileType } from "file-type";` or `import * as fileType from "file-type";`?

src/file-uploader/file-uploader.component.ts
```diff
@@ -125,6 +149,13 @@ export class FileUploader {
 	protected onTouchedCallback: () => void = noop;
 	protected onChangeCallback: (_: Set<FileItem>) => void = noop;
 
+	/**
+	 * Controls the state of the drag and drop file container
+	 */
+	protected dragOver = false;
+
+	protected _fileList = [];
```
cal-smith: I'm not totally sure why we have `fileList` _and_ `files`? Is there something we can't do with the `files` Set that merits having both?

src/file-uploader/file-uploader.component.ts
```diff
@@ -175,12 +216,60 @@ export class FileUploader {
...
+			return;
+		}
+
+		const fileExtensionRegExp = new RegExp(/\.[0-9a-z]+$/, "i");
+
+		const acceptedFiles = transferredFiles.filter(({ name, type }) => {
+			if (!fileExtensionRegExp.test(name)) {
+				return false;
+			}
+			const [fileExtension] = name.match(fileExtensionRegExp);
```
cal-smith: This line is kinda confusing ... We don't have to reuse the React code

src/file-uploader/file-uploader.stories.ts
```diff
@@ -84,6 +86,122 @@ class FileUploaderStory {
...
+		const transferredFiles = Array.from(event);
+
+		const fileExtensionRegExp = new RegExp(/\.[0-9a-z]+$/, "i");
+
+		let promises: Array<Promise<any>> = [];
+
+		// Creates a promise which resolves to a file and whether or not the file should be accepted.
+		const readFileAndAddToMap = fileObj => {
+			return new Promise((resolve, reject) => {
+				let fileExtension, mime;
```
cal-smith: I'm not sure why we need to forward declare these?

src/file-uploader/file-uploader.stories.ts
```diff
@@ -84,6 +86,122 @@ class FileUploaderStory {
...
+					const type = fileType(reader.result);
+					if (type) {
+						fileExtension = type.ext.replace(/^/, ".");
+						mime = type.mime;
+					} else {
+						// If a file type can not be determined using magic numbers
+						// then use file extension or mime type.
+						[fileExtension] = fileObj.file.name.match(fileExtensionRegExp);
+						mime = fileObj.file.type;
+					}
```
cal-smith: This logic isn't totally clear ...

src/file-uploader/file-uploader.stories.ts
```diff
@@ -84,6 +86,122 @@ class FileUploaderStory {
...
+					});
+				};
+
+				reader.onerror = error => {
+					reject(error);
+				};
+			});
+		};
+
+		transferredFiles.forEach(file => {
```
cal-smith: Could we `.map` the transferred files instead and just initialize a `const` promises array here?

src/file-uploader/file-uploader.stories.ts
```diff
@@ -84,6 +86,122 @@ class FileUploaderStory {
...
+	// preliminary filtration and is optional.
+	onDropped(event) {
+		const transferredFiles = Array.from(event);
+
+		const fileExtensionRegExp = new RegExp(/\.[0-9a-z]+$/, "i");
+
+		let promises: Array<Promise<any>> = [];
+
+		// Creates a promise which resolves to a file and whether or not the file should be accepted.
+		const readFileAndAddToMap = fileObj => {
```
cal-smith: I don't see the `AndAddToMap` part of this function?

src/file-uploader/file-uploader.stories.ts
```diff
@@ -84,6 +86,122 @@ class FileUploaderStory {
...
+		};
+
+		transferredFiles.forEach(file => {
+			promises.push(readFileAndAddToMap(file));
+		});
+
+		Promise.all(promises).then(filePromiseArray =>
+			filePromiseArray.filter(filePromise => filePromise.accept)
+				.map(acceptedFile => acceptedFile.file))
+					.then(acceptedFiles => this.files = new Set(acceptedFiles)).catch(error => console.log(error));
```
cal-smith: This indentation is awkward, and the `catch` should be on a newline

src/file-uploader/file-uploader.stories.ts
Donisius: If a file type can not be determined using magic numbers the fileType method will return undefined. So in that case it uses the file extension or mime type. Would a comment be enough to explain that or should I rethink the logic?

Donisius: I just realized that the else block is the exact same as the preliminary filter in the component. In that case would it be better to just accept the file automatically if it reaches that point?

cal-smith: Most likely, seems redundant otherwise

PR #930, opened 2019-11-22
feat(notification): add support for actions

### LINKED ISSUE #928: Notification component support for buttons/links similar to React version
Notification component in Angular is missing a much needed ability to launch actions from it. Either by rendering buttons or links. Support for using any component in the notification.

>
## Detailed description

> UX team has designed several scenarios that have buttons or email links in the notification/toast component.  The component does not appear to support additional non text elements as part of its construction.  See image below of what our design is expecting.  

> Version of the Carbon Angular 3.21.2 . carbon 10.6.0

> IBM Study Advance - Mid January is next release.

## Additional information

image

image

Donisius:
Closes #928

### DIFF
--- a/src/notification/notification-content.interface.ts
+++ b/src/notification/notification-content.interface.ts
@@ -1,18 +1,25 @@
+import { Subject } from 'rxjs';
+
 export interface NotificationContent {
 	type: string;
 	title: string;
 	target?: string;
-	actionText?: string;
 	duration?: number;
 	smart?: boolean;
 	closeLabel?: any;
 	message?: string;
 	showClose?: boolean;
 	lowContrast?: boolean;
-	showAction?: boolean;
+	actions?: NotificationAction[];
 }
 
 export interface ToastContent extends NotificationContent {
 	subtitle: string;
 	caption: string;
+}
+
+interface NotificationAction {
+	text: string;
+	click: Subject<{event: Event, action: any}> | ((event: {event: Event, action: any}) => any);
+	[x:string]: any;
 }
--- a/src/notification/notification-content.interface.ts
+++ b/src/notification/notification-content.interface.ts
@@ -2,12 +2,14 @@ export interface NotificationContent {
 	type: string;
 	title: string;
 	target?: string;
+	actionText?: string;
 	duration?: number;
 	smart?: boolean;
 	closeLabel?: any;
 	message?: string;
 	showClose?: boolean;
 	lowContrast?: boolean;
+	showAction?: boolean;
 }
 
 export interface ToastContent extends NotificationContent {
--- a/src/notification/notification-content.interface.ts
+++ b/src/notification/notification-content.interface.ts
@@ -1,4 +1,4 @@
-import { Subject } from 'rxjs';
+import { Subject } from "rxjs";
 
 export interface NotificationContent {
 	type: string;
@@ -21,5 +21,5 @@ export interface ToastContent extends NotificationContent {
 interface NotificationAction {
 	text: string;
 	click: Subject<{event: Event, action: any}> | ((event: {event: Event, action: any}) => any);
-	[x:string]: any;
+	[x: string]: any;
 }
--- a/src/notification/notification.component.ts
+++ b/src/notification/notification.component.ts
@@ -43,7 +43,7 @@ import { of, isObservable, Subject } from "rxjs";
 		</div>
 		<div *ngFor="let action of notificationObj.actions">
 			<button
-				(click)="onClick(action.click, $event)"
+				(click)="onClick(action, $event)"
 				ibmButton="ghost"
 				size="sm"
 				class="bx--inline-notification__action-button"
@@ -121,13 +121,13 @@ export class Notification {
 	}
 
 	onClick(action, event) {
-		if (!action) {
+		if (!action.click) {
 			return;
 		}
-		if (isObservable(action)) {
-			(action as Subject<{event: Event, action: any}>).next({event, action: event.action});
+		if (isObservable(action.click)) {
+			(action.click as Subject<{event: Event, action: any}>).next({event, action});
 		} else {
-			action({event, action: event.action});
+			action.click({event, action});
 		}
 	}
 
--- a/src/notification/notification.component.ts
+++ b/src/notification/notification.component.ts
@@ -121,6 +121,9 @@ export class Notification {
 	}
 
 	onClick(action) {
+		if (!action) {
+			return;
+		}
 		if (typeof action === "function") {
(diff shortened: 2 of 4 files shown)


src/notification/notification-content.interface.ts
```diff
@@ -2,12 +2,14 @@ export interface NotificationContent {
...
 	title: string;
 	target?: string;
+	actionText?: string;
 	duration?: number;
 	smart?: boolean;
 	closeLabel?: any;
 	message?: string;
 	showClose?: boolean;
 	lowContrast?: boolean;
+	showAction?: boolean;
```
cal-smith: Would it make more sense to have it be `actions: NotificationAction[]` where `NotificationAction` looks like

```typescript
interface NotificationAction {
    text: string;
    click: Subject<ClickEvent> | (event: ClickEvent) => any;
}
```

So that a user could define multiple actions?

src/notification/notification.component.ts
```diff
@@ -124,11 +120,12 @@ export class Notification {
 		this.close.emit();
 	}
 
-	/**
-	 * Emits action event.
-	 */
-	onAction() {
-		this.action.emit();
+	onClick(action) {
+		if (typeof action === "function") {
```
cal-smith: `isObservable` may be more reliable ... Also we'll want to check if the callback/observable even exists first and bail out if it doesn't (not sure why someone would do that, but people do strange things)

src/notification/notification.stories.ts
```diff
@@ -1,12 +1,69 @@
...
+			actions: actions}">
+		</ibm-notification>
+		`,
+	providers: [NotificationService]
+})
+class NotificationActionStory {
+	@Input() showClose = true;
+	@Input() lowContrast = false;
+
+	subject = new Subject<any>();
```
cal-smith: Maybe `actionSubject`? Just `subject` is pretty vague 🤔

src/notification/notification.stories.ts
```diff
@@ -1,12 +1,69 @@
 import { storiesOf, moduleMetadata } from "@storybook/angular";
-import { action } from "@storybook/addon-actions";
+import { action, actions } from "@storybook/addon-actions";
 import { withKnobs, boolean, object } from "@storybook/addon-knobs/angular";
 
-import { Component } from "@angular/core";
+import { Component, Testability, Input } from "@angular/core";
```
cal-smith: I don't think we need `Testability` here 🤔 

```suggestion
import { Component, Input } from "@angular/core";
```

src/notification/notification.stories.ts
```diff
@@ -1,12 +1,69 @@
 import { storiesOf, moduleMetadata } from "@storybook/angular";
-import { action } from "@storybook/addon-actions";
+import { action, actions } from "@storybook/addon-actions";
```
cal-smith: I don't see these imports being used anywhere ... that said, if there are some events we could hook up to the storybook action addon that would be awesome 👍

src/notification/notification.component.ts
Donisius: isObservable will always return false because after subscribing to the action subject it is no longer an observable and becomes of type object instead.

cal-smith: Hmm I don't think that should happen 🤔 interesting

src/notification/notification.stories.ts
```diff
@@ -1,12 +1,70 @@
...
+	@Input() lowContrast = false;
+
+	@Output() actionClicked: EventEmitter<any> = new EventEmitter();
+
+	actionSubject = new Subject<any>();
+
+	actions = [
+		{
+			text: "Action",
+			click: this.actionSubject.subscribe({
```
cal-smith: The `isObservable` check isn't working because you're not passing an observable to the component - you're passing a subscription (`.subscribe` returns a `Subscription`) ... Maybe add a `ngOnInit` and subscribe to the action there 👍

Donisius: It still seems to be type `object` when I pass in only actionSubject without subscribing.

cal-smith: Odd ... what happens if you initialize `actions` in `ngOnInit`?

src/notification/notification.component.ts
```diff
@@ -110,6 +120,17 @@ export class Notification {
...
 
+	onClick(action) {
+		if (!action) {
+			return;
+		}
+		if (isObservable(action)) {
+			(action as Subject<any>).next();
+		} else {
+			action();
+		}
```
cal-smith: So the type for this is `Subject<{event: Event, action: any}> | ((event: {event: Event, action: any}) => any);`

So we should be sure to pass the DOM event and action data back 👍

src/notification/notification.component.ts
```diff
@@ -120,14 +120,14 @@ export class Notification {
...
 	}
 
-	onClick(action) {
+	onClick(action, event) {
 		if (!action) {
 			return;
 		}
 		if (isObservable(action)) {
-			(action as Subject<any>).next();
+			(action as Subject<{event: Event, action: any}>).next({event, action: event.action});
```
cal-smith: I'm not sure that `event.action` is the right `action` ... instead of passing `action.click` in the template, you'll want to pass the whole `action` object and just call `.click` in here :+1:

src/notification/notification.stories.ts
```diff
@@ -1,12 +1,72 @@
...
+			actions: actions}">
+		</ibm-notification>
+		`,
+	providers: [NotificationService]
+})
+class NotificationActionStory implements OnInit {
+	@Input() showClose = true;
+	@Input() lowContrast = false;
+
+	@Output() actionClicked: EventEmitter<any> = new EventEmitter();
```
theiliad:
```suggestion
	@Output() actionClicked = new EventEmitter();
```

src/notification/notification.stories.ts
```diff
@@ -1,12 +1,72 @@
 import { storiesOf, moduleMetadata } from "@storybook/angular";
 import { action } from "@storybook/addon-actions";
-import { withKnobs, boolean, object } from "@storybook/addon-knobs/angular";
+import { withKnobs, boolean } from "@storybook/addon-knobs/angular";
 
-import { Component } from "@angular/core";
+import { Component, Input, EventEmitter, Output, OnInit } from "@angular/core";
```
zvonimirfras:
```suggestion
import {
	Component,
	Input,
	EventEmitter,
	Output,
	OnInit
} from "@angular/core";
```

Donisius: Done!

PR #933, opened 2019-11-25
fix(slider): Stop trapping all keys and listen only for the arrow keys.

stanislavgeorgiev:
Closes IBM/carbon-components-angular#932


stanislavgeorgiev:
My pleasure :)
Got some time in my hands now.

PR #934, opened 2019-11-26
Replace pure svg with icon directive for ui-shell stories

Donisius:


src/ui-shell/ui-shell.stories.ts
```diff
@@ -65,26 +69,10 @@ storiesOf("UI Shell", module)
...
-						<svg width="20" height="20" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32" aria-hidden="true">
-							<path
-								d="M8.24 25.14L7 26.67a14 14 0 0 0 4.18 2.44l.68-1.88a12
-								12 0 0 1-3.62-2.09zm-4.05-7.07l-2 .35A13.89 13.89 0 0 0 3.86
-								23l1.73-1a11.9 11.9 0 0 1-1.4-3.93zm7.63-13.31l-.68-1.88A14
-								14 0 0 0 7 5.33l1.24 1.53a12 12 0 0 1 3.58-2.1zM5.59
-								10L3.86 9a13.89 13.89 0 0 0-1.64 4.54l2 .35A11.9 11.9 0 0 1 5.59
-								10zM16 2v2a12 12 0 0 1 0 24v2a14 14 0 0 0 0-28z" />
-						</svg>
+						<svg icon ibmIconFade20></svg>
```
cal-smith: Is the `icon` attribute required by the carbon styles? We may want to open an issue in the carbon monorepo about that if so ...

Donisius: No, our sidenav-item component selects the icon attribute:
```
<div *ngIf="!isSubMenu" class="bx--side-nav__icon">
      <ng-content select="[icon]"></ng-content>
</div>
```

cal-smith: Ohhh, interesting 🤔 I wonder if we can add a selector for `svg` as well?

PR #935, opened 2019-11-28
feat(slider): Added option to have 2 slider handles for a ranged selection

stanislavgeorgiev:
Closes IBM/carbon-components-angular#916

What do you think about calling the second value for the range `value2`? In our implementation we have it `lowValue` and `highValue` but obviously I don't want to break anybody else's code and rename `value` to `lowValue` :)

I was initially thinking about creating a `range-slider.component` which will extend the `slider.component`. That didn't go well because I didn't want to clone the template code so I settled to do append some code to the `slider` itself.

### DIFF
--- a/src/slider/slider.component.ts
+++ b/src/slider/slider.component.ts
@@ -72,6 +72,14 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from "@angular/forms";
 						(keydown)="onKeyDown($event)">
 					</div>
 					<div
+						#thumb2
+						*ngIf="isRange()"
+						class="bx--slider__thumb"
+						tabindex="0"
+						(mousedown)="onMouseDown($event, false)"
+						(keydown)="onKeyDown($event, false)">
+					</div>
+					<div
 						#track
 						class="bx--slider__track"
 						(click)="onClick($event)">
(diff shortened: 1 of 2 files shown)


cal-smith:
`lowValue` and `highValue` would work with some getter/setter trickery ... The only issue there is that `ngModel` is only going to return one of the values ... Maybe for range sliders we could take an array of values? Internally we can use getters to pull the appropriate values from it ...

stanislavgeorgiev:
Yes, I couldn't solve the `ngModel` issue and that's why I wanted to submit the PR to get your take on how is everything looking in general and else should be updated/added.

One other potential problem is that the 2 thumbs can get almost on top of each other when the values are less than 4-5 points apart and it's visually unpleasing. Not the biggest issue of all but just worth mentioning.

cal-smith:
> Yes, I couldn't solve the ngModel issue and that's why I wanted to submit the PR to get your take on how is everything looking in general and else should be updated/added.

For sure!

>One other potential problem is that the 2 thumbs can get almost on top of each other when the values are less than 4-5 points apart and it's visually unpleasing. Not the biggest issue of all but just worth mentioning.

I suppose one option there would be to limit how close the handles can get ... So long as they can't cross each other I don't think it's a huge deal

cal-smith:
@stanislavgeorgiev any updates on the `ngModel` support?

stanislavgeorgiev:
Not yet @calsmith.

cal-smith:
@stanislavgeorgiev had a chance to work on this? No worries if not, I can take this over if needed 👍

stanislavgeorgiev:
I'm in a pre-release phase now until June 19th and cannot put any time in this item. I can come back to it after that. Sorry I'm keeping this open for 6 months!

cal-smith:
Not a problem! Just wanted to check in 😁 I'll check back after the 19th and we can go from there 👍

PR #936, opened 2019-11-28
feat(table): allow for toolbar batch text to be customizable

Donisius:


src/table/toolbar/table-toolbar.component.ts
```diff
@@ -64,6 +64,8 @@ import { I18n, Overridable } from "../../i18n/i18n.module";
 export class TableToolbar {
 	@Input() model: TableModel;
 
+	@Input() batchText = "items selected";
```
cal-smith: We'll want to add this to the translations as well, since it'll likely be a single translation unit for most cases

src/table/toolbar/table-toolbar.component.ts
```diff
@@ -63,8 +63,12 @@ import { I18n, Overridable } from "../../i18n/i18n.module";
 	`
 })
 export class TableToolbar {
-	@Input() model: TableModel;
+@Input() model: TableModel;
 
+	@Input() set batchText (value: { BATCH_TEXT: string }) {
+		console.log(value.BATCH_TEXT);
+		this._batchText.override(value.BATCH_TEXT);
+	}
```
cal-smith: Since this is just setting one value it should take a `value: string` and just apply that as the override ... I'm not sure what's going on with the `ariaLabel` taking an object ... that shouldn't be like that, but I'm assuming it's from before so there's not a lot of room to change it

PR #937, opened 2019-11-28
feat(tabs): add decoupled tab headers and tab panes

### LINKED ISSUE #354: Decouple tab headers and tab panes
Right now tabs, headers, and panes are all tied together as one concept. To increase flexibility we should decouple the components and tie them together via some form of model/service.

Donisius:
closes #354

To increase flexibility, two new components are created for a more decoupled use case of tabs.

#### Changelog

**New**

* Added tab-header and tab-header-group components.


src/tabs/tab-header-group.component.ts
```diff
@@ -0,0 +1,284 @@
...
+
+		this.tabHeaderQuery.toArray()[this.currentSelectedIndex].selectTab();
+	}
+
+	public getSelectedTab(): any {
+		const selected =  this.tabHeaderQuery.toArray()[this.currentSelectedIndex];
+		if (selected) {
+			return selected;
+		}
+		return {headingIsTemplate: false, heading: ""};
```
cal-smith: Minor nitpick

```suggestion
		return {
				headingIsTemplate: false, 
				heading: ""
		};
```

src/tabs/tab-header.component.ts
```diff
@@ -0,0 +1,103 @@
...
+			class="bx--tabs__nav-item"
+			role="presentation"
+			(click)="selectTab()">
+			<a
+				#tabItem
+				[attr.aria-selected]="active"
+				draggable="false"
+				class="bx--tabs__nav-link"
+				href="javascript:void(0)"
+				[attr.tabindex]="(active?0:-1)"
```
cal-smith:
```suggestion
				[attr.tabindex]="(active ? 0 : -1)"
```

src/tabs/tab-header.component.ts
```diff
@@ -0,0 +1,103 @@
...
+				[attr.tabindex]="(active? 0 : -1)"
+				role="tab">
+				<ng-container *ngIf="!isTemplate(heading)">
+					{{ heading }}
+				</ng-container>
+				<ng-template
+					*ngIf="isTemplate(heading)"
+					[ngTemplateOutlet]="heading"
+					[ngTemplateOutletContext]="{$implicit: context}">
+				</ng-template>
```
cal-smith: Since the header is all net new code, could we use `ng-content` instead of templates? That way we don't have to worry about passing around all sorts of context for the header

Donisius: Yeah that would make the component smaller

src/tabs/tab-header.component.ts
```diff
@@ -0,0 +1,72 @@
...
+	 */
+	@Input() disabled = false;
+	/**
+	 * Reference to the corresponsing tab pane.
+	 */
+	@Input() paneReference: Tab;
+	/**
+	 * Allows the user to pass data to the custom template for the `Tab` heading.
+	 */
+	@Input() context: any;
```
cal-smith: We don't need this anymore

src/tabs/tab-header-group.component.ts
```diff
@@ -0,0 +1,287 @@
...
+	 */
+	@Input() ariaLabel: string;
+	/**
+	 * Sets the aria labelledby on the nav element.
+	 */
+	@Input() ariaLabelledby: string;
+
+	@Input() contentAfter;
+
+	@Input() contentBefore;
```
theiliad: no types for the 2 above?

Donisius: Do you think it should be of type TemplateRef<any>?

src/tabs/tab-header.component.ts
```diff
@@ -0,0 +1,68 @@
...
+	 */
+	@Input() disabled = false;
+	/**
+	 * Reference to the corresponsing tab pane.
+	 */
+	@Input() paneReference: Tab;
+	/**
+	 * Value 'selected' to be emitted after a new `Tab` is selected.
+	 */
+	@Output() selected: EventEmitter<any> = new EventEmitter<any>();
```
theiliad:
```suggestion
	@Output() selected = new EventEmitter<any>();
```

PR #938, opened 2019-11-28
feat(table): add support for col and row spans

cal-smith:
Closes IBM/carbon-components-angular#931

### DIFF
--- a/src/table/table-adapter.class.ts
+++ b/src/table/table-adapter.class.ts
@@ -10,6 +10,10 @@ export abstract class TableCellAdapter {
 	 * The number of columns spanned by this cell
 	 */
 	colSpan: number;
+	/**
+	 * The number of rows spanned by this cell
+	 */
+	rowSpan: number;
 }
 
 /**
@@ -32,7 +36,7 @@ export abstract class TableRowAdapter {
  */
 export abstract class TableAdapter {
 	/**
-	 * The last acccessible column in the table
+	 * The last accessible column in the table
 	 */
 	public get lastColumnIndex(): number { return; }
 
@@ -48,6 +52,13 @@ export abstract class TableAdapter {
 	 * @param column index of the column
 	 */
 	getCell(row: number, column: number): TableCellAdapter { return; }
+
+	/**
+	 * Returns a row from the table
+	 *
+	 * @param column index of the column
+	 */
+	getColumn(column: number): TableCellAdapter[] { return; }
 
 	/**
 	 * Returns a row from the table
@@ -86,10 +97,10 @@ export abstract class TableAdapter {
  */
 export class TableDomAdapter implements TableAdapter {
 	/**
-	 * The last acccessible column in the table
+	 * The last accessible column in the table
 	 */
 	public get lastColumnIndex() {
-		return this.tableElement.rows[0].cells.length - 1;
+		return this.getRealRowLength(this.tableElement.rows[0]);
 	}
 
 	/**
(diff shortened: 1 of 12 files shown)


src/table/table-adapter.class.ts
```diff
@@ -116,23 +127,47 @@ export class TableDomAdapter implements TableAdapter {
...
+	/**
+	 * Returns a column from the table, using the `id` and `headers` attributes
+	 *
+	 * See here for more detail these attributes: https://www.w3.org/TR/WCAG20-TECHS/H43.html
+	 *
+	 * @param column the index of the column
+	 */
+	getColumn(column: number): HTMLTableCellElement[] {
+		const firstHeader = Array.from(this.tableElement.rows[0].cells);
+
```
theiliad: empty line?

src/table/table-adapter.class.ts
```diff
@@ -116,23 +127,47 @@ export class TableDomAdapter implements TableAdapter {
...
+		const { cell: header, realIndex: realColumnIndex } = this.findCellInRow(firstHeader, column);
+
```
theiliad: empty line?

src/table/table-adapter.class.ts
```diff
@@ -116,23 +127,47 @@ export class TableDomAdapter implements TableAdapter {
...
+		const linkedCells: HTMLTableCellElement[] = [];
+
+		for (let i = 1; i < this.tableElement.rows.length; i++) {
```
theiliad: Can we use `forEach` here?

src/table/table-adapter.class.ts
```diff
@@ -116,23 +127,47 @@ export class TableDomAdapter implements TableAdapter {
...
+
+	/**
+	 * Returns a column from the table, using the `id` and `headers` attributes
+	 *
+	 * See here for more detail these attributes: https://www.w3.org/TR/WCAG20-TECHS/H43.html
+	 *
+	 * @param column the index of the column
+	 */
+	getColumn(column: number): HTMLTableCellElement[] {
+		const firstHeader = Array.from(this.tableElement.rows[0].cells);
```
theiliad: `Array.from` has no IE support

image

src/table/table-adapter.class.ts
```diff
@@ -151,6 +186,53 @@ export class TableDomAdapter implements TableAdapter {
...
+			// sort the indexes largest to smallest to find the closest matching header index
+			const firstIndex = indexes.sort((a, b) => b - a)[0];
+
+			// search the row for cells that share the header
+			let similarCells = [];
+			for (const id of ids) {
+				// there's no selector that will match two space separated lists,
+				// so we have to iterate through the ids and query the row for each
+				const cells = Array.from(row.querySelectorAll(`[headers~='${id}']`));
+				for (const cell of cells) {
```
theiliad: Does `cell` not clash here with the `cell` parameter in the function?

src/table/table-adapter.class.ts
```diff
@@ -151,6 +186,53 @@ export class TableDomAdapter implements TableAdapter {
...
+					// only keep one set of cells
+					if (!similarCells.includes(cell)) {
```
theiliad: `includes` has no IE support

image

src/table/table-adapter.class.ts
```diff
@@ -181,4 +263,66 @@ export class TableDomAdapter implements TableAdapter {
...
+	 * Helper function that returns the "real" length of a row.
+	 * Only accurate with regard to colspans (though that's sufficient for it's uses here)
+	 *
+	 * TODO: Take rowSpan into account
+	 *
+	 * @param row the row to get the length of
+	 */
+	protected getRealRowLength(row: HTMLTableRowElement): number {
+		return Array.from(row.cells).reduce((count, cell) => {
+			return count + cell.colSpan;
```
theiliad: can we drop the `return` here?

src/table/table-adapter.class.ts
```diff
@@ -181,4 +263,66 @@ export class TableDomAdapter implements TableAdapter {
...
+	}
+
+	/**
+	 * Finds a cell and it's real index given an array of cells, a target index, and the spanning direction
+	 *
+	 * @param cells An array of cells to search
+	 * @param targetIndex The index we think the cell is located at
+	 * @param spanDirection The direction of the cell spans. Should be `"colSpan"` for a row and `"rowSpan"` for a column
+	 */
+	protected findCell(cells: HTMLTableCellElement[], targetIndex: number, spanDirection: "colSpan" | "rowSpan") {
```
theiliad: Can `spanDirection` be an enum?

src/table/table-model.class.ts
```diff
@@ -149,6 +154,64 @@ export class TableModel implements PaginationModel {
...
+		TableModel.COUNT++;
+	}
+
+	/**
+	 * Returns an id for the given column
+	 *
+	 * @param column the column to generate an id for
+	 * @param row the row of the header to generate an id for
+	 */
+	getId(column: number | "select" | "expand", row = 0): string {
```
theiliad: can `column` use enums or interfaces?

src/table/table-adapter.class.ts
cal-smith: That's fine, we have polyfills for that 👍

src/table/table-adapter.class.ts
cal-smith: Just for readability 🤷‍♂ totally can be removed if we prefer that

src/table/table-adapter.class.ts
cal-smith: same as above

src/table/table-adapter.class.ts
cal-smith: Probably 🤔

src/table/table-adapter.class.ts
cal-smith: Huh, totally missed that ... `const`/`let` will lexically scope it so it's not an issue, but for readability we should rename it 👍

src/table/table-adapter.class.ts
cal-smith: Polyfills!

src/table/table-adapter.class.ts
cal-smith: Yeah, so long as it doesn't make the line unreadable

src/table/table-adapter.class.ts
cal-smith: Maybe! It kinda is already, but an explicit enum may help

src/table/table-model.class.ts
cal-smith: not sure how I'd apply an interface to this ... Enum would be mostly the same (union with `number | HeaderTypeEnum`) so I'm not sure it's a strict advantage

theiliad: Right now we've got
`getId(column: number | "select" | "expand", row = 0): string {`
and
`getHeaderId(column: number | "select" | "expand", colSpan = 1): string {`
that repeat the inline enum.

We should define somewhere else and re-use

src/table/table-adapter.class.ts
cal-smith: So for whatever reason `.rows` is a `HTMLCollection` instead of a `NodeList` ... so we can't `forEach` it - https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection

theiliad: I was suspecting that that's the reason you use `for`

src/table/table-model.class.ts
cal-smith: There's also not a clean way to apply enum values from a template ...

theiliad: Can't we just do `getHeaderId(column: number | HeaderTypeEnum, colSpan = 1): string {`?

cal-smith: Sure, but in each component that calls that method in the template we'd have to keep a local `headerTypeEnum` on the class ... not super ideal

cal-smith: We get the same type guarantees with string unions as with enums in this case, so there's not a huge advantage (and it's not like we'll have 50 string options)

cal-smith: What we _can_ do is add a `type` definition so we can reuse it

cal-smith: so: `type HeaderType = number | "select" | "expand"`

and `getHeaderId(column: HeaderType`

PR #939, opened 2019-12-02
fix(date picker): display correct date on calendar when focused or updated

Donisius:
When date picker is in range mode:

* Opening the calendar using the calendar icon will display the date which was last opened. For example, if the first input was clicked first, then clicking the second input will display the first input's date on the calendar.

* Switching focus between the inputs with tab will always display the second input's month. If at any point the second input has received focus.

* Manually changing the input and then clicking enter did not display the correct month on the calendar.

It all behaves as expected now.

#### Changelog

**New**

* Manually update the month when input receives focus.

* Call updateCurrentMonth in the onMonthChange event hook in the month select plugin.

**Changed**

* When opening the calendar using the calendar icon, manually change the current month and year to make sure the calendar displays the correct date.

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -227,7 +227,6 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 			if (this.rangeInput.input.nativeElement === document.activeElement && this.flatpickrInstance.selectedDates[1]) {
 				const currentMonth = this.flatpickrInstance.selectedDates[1].getMonth();
 				this.flatpickrInstance.changeMonth(currentMonth, false);
-
 			} else if (this.input.input.nativeElement === document.activeElement && this.flatpickrInstance.selectedDates[0]) {
 				const currentMonth = this.flatpickrInstance.selectedDates[0].getMonth();
 				this.flatpickrInstance.changeMonth(currentMonth, false);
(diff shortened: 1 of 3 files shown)


src/datepicker/datepicker.component.ts
```diff
@@ -213,6 +221,19 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 
 	@HostListener("focusin")
 	onFocus() {
+		// Updates the month manually when calendar mode is range because month
+		// will not update properly without manually updating them on focus.
+		if (this.range) {
+			if (this.rangeInput.input.nativeElement === document.activeElement && this.flatpickrInstance.selectedDates[1]) {
+				const currentMonth = this.flatpickrInstance.selectedDates[1].getMonth();
+				this.flatpickrInstance.changeMonth(currentMonth, false);
+
```
zvonimirfras:
```suggestion
```

src/datepicker/datepicker.component.ts
```diff
@@ -273,10 +294,23 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
...
+	 * Handles opening the calendar "properly" when the calendar icon is clicked.
+	 */
+	openCalendar(datepickerInput: DatePickerInput) {
+		datepickerInput.input.nativeElement.click();
+
+		// If the first input's calendar icon is clicked when calendar is in range mode, then
+		// the month and year needs to be manually changed to the current selected month and
+		// year otherwise the calendar view will not be updated upon opening.
+		if (datepickerInput === this.input && this.range && this.flatpickrInstance.selectedDates[0]) {
+
```
zvonimirfras:
```suggestion
```

PR #943, opened 2019-12-04
fix(pagination): add sensible default values and better value checking

cal-smith:


src/pagination/pagination-model.class.ts
```diff
@@ -2,15 +2,15 @@ export class PaginationModel {
 	/**
 	 * Tracks the current page.
 	 */
-	currentPage: number;
+	currentPage = 0;
```
theiliad: Shouldn't current page be 1?

src/table/table-model.class.ts
```diff
@@ -90,22 +90,22 @@ export class TableModel implements PaginationModel {
...
 	 */
-	header: TableHeaderItem[];
+	header: TableHeaderItem[] = [];
 
 	/**
-	 * Tracks the current page of the table.
+	 * Tracks the current page.
 	 */
-	currentPage: number;
+	currentPage = 0;
```
theiliad: Shouldn't current page be 1?

src/pagination/pagination-model.class.ts
cal-smith: 🤔 we may zero index pages ... let me check ...

PR #951, opened 2019-12-10
fix(table): Make sortable optional

### LINKED ISSUE #947: ibm-table is sortable by default
## Detailed description

After upgrading carbon-components-angular from version **3.16.3** to version **3.28.0**, we found that the table headers are by default sortable and there is no way to remove sorting arrows.

This is a bug

<ibm-table> component

The table should not be sortable by default. Instead, there should be a provision to let the developer decide weather the table should be sortable or not.

Issue is reproducible on all browsers

"carbon-components": "^10.6.2"
"carbon-components-angular": "^3.28.0"

The issue can be seen here: https://angular.carbondesignsystem.com/?path=/story/components-table--basic

## Additional information

* Screenshots or code

image

* Notes

On further checking the documentation, the sortable is hardcoded to 'true'. It should be of type input, so that developer can enable or disable sortable 

`<table
	ibmTable
	sortable="true"
	[size]="size"
	[striped]="striped"
	[skeleton]="skeleton">`

https://angular.carbondesignsystem.com/?path=/story/components-table--documentation

Donisius:
Closes #947


src/table/head/table-head.component.ts
```diff
@@ -86,6 +88,8 @@ export class TableHead {
 
 	@Input() stickyHeader = false;
 
+	@Input() sortable = true;
```
cal-smith: Maybe add a comment to note that disabling this will disable sorting **for all headers, including sortable ones** and to generally prefer setting the sortable state on the header item.

src/table/table.component.ts
```diff
@@ -357,6 +359,10 @@ export class Table implements AfterViewInit, OnDestroy {
 		}
 	}
 
+	@Input() sortable = true;
```
cal-smith: Same comment about the sortable state here 👍

src/table/head/table-head-cell.component.ts
```diff
@@ -17,7 +17,7 @@ import { TableHeaderItem } from "./../table-header-item.class";
 	template: `
 		<button
 			class="bx--table-sort"
-			*ngIf="!skeleton && this.sort.observers.length > 0 && column.sortable"
+			*ngIf="sortable && !skeleton && this.sort.observers.length > 0 && column.sortable"
```
cal-smith: I _think_ we can remove `this.sort.observers.length > 0` ... I'm not sure how it's even working in the template as is ...

Donisius: Should it be removed in the other places it is present as well?

cal-smith: Basically just checks to see if there's any listeners to the observer ... it seems to be safe to remove in the head cell, but I'm not sure about the other components

cal-smith:
Bump @Donisius

src/table/head/table-head.component.ts
```diff
@@ -95,6 +97,12 @@ export class TableHead implements AfterViewInit {
 
 	@Input() stickyHeader = false;
 
+	/**
+	 * Setting sortable to false will disable all headers including headers which are sortable. Is is
+	 * possible to set the sortable state on the header item to disable/enable sorting for only some headers.
+	 */
```
natashadecoste: documentation 👍  🎉

src/table/table.component.ts
```diff
@@ -359,6 +361,14 @@ export class Table implements AfterViewInit, OnDestroy {
 		}
 	}
 
+	/**
+	 * Setting sortable to false will disable all headers including headers which are sortable. Is is
+	 * possible to set the sortable state on the header item to disable/enable sorting for only some headers.
+	 */
+	@Input() sortable = true;
+
+	@Input() noBorder = true;
```
natashadecoste: support for a table rendered without a border? Try to make sure to document new provisions

PR #952, opened 2019-12-10
feat(tiles): Add skeleton states

### LINKED ISSUE #946: Missing ibm-tile skeleton
## Missing ibm-tile skeleton - There's no skeleton for ibm-tile component

To follow the standards of most of the components, ibm-tile should have it's skeleton available.

Donisius:
Closes #946


Donisius:
> Instead of just blanking out, maybe it would be nice if it could toggle to a provided skeleton content/template?



> Instead of just blanking out, maybe it would be nice if it could toggle to a provided skeleton content/template?

Good call, then the skeleton state wouldn't allow the users to perform any actions on the component.

cal-smith:
Sorry, I meant a little more involved than that ... If I have a tile I may have arbitrary content in it, when the tile is in a skeleton state I should be able to provide a skeleton template/ng-content that allows me to represent that unloaded content

src/skeleton-placeholders/skeleton-text.component.ts
```diff
@@ -0,0 +1,38 @@
...
+	template: `
+		<p
+			*ngFor="let width of lineWidths"
+			class="bx--skeleton__text"
+			[style.width]="width">
+		</p>
+	`
+})
+export class SkeletonText implements OnChanges {
+	@Input() numberOfLines = 5;
```
cal-smith: Maybe just `lines`? ... not totally sure though

src/skeleton/skeleton-text.component.ts
```diff
@@ -0,0 +1,38 @@
...
+export class SkeletonText implements OnChanges {
+	@Input() lines = 5;
+
+	@Input() minLineWidth = 100;
+
+	@Input() maxLineWidth = 300;
+
+	lineWidths: Array<string>;
+
+	randoms = [0.973051493507435, 0.15334737213558558, 0.5671034553053769];
```
zvonimirfras: That's... unusual...
Why not generate these one needed so they are really random?

cal-smith: So the advantage here is that the lines can be more consistent, compare the react demo: http://react.carbondesignsystem.com/?path=/story/skeletontext--default to our demo: https://deploy-preview-952--c4ng.netlify.com/?path=/story/components-skeleton--skeleton-text

I'm not sure which is preferable, but the react one is far more consistent

src/skeleton/skeleton.stories.ts
```diff
@@ -0,0 +1,30 @@
...
+	.add("Skeleton Placeholder", () => ({
+		template: `<ibm-skeleton-placeholder></ibm-skeleton-placeholder>`
+	}))
+	.add("Skeleton Text", () => ({
+		template: `
+			<ibm-skeleton-text
+				[lines]="lines"
+				[minLineWidth]="minLineWidth"
+				[maxLineWidth]="maxLineWidth"
+				>
```
zvonimirfras: We don't normally keep these lonely, we keep them on the previous line :)

Donisius: no longer lonely 😃

PR #972, opened 2019-12-19
feat(Date picker): Add localization support for date picker

### LINKED ISSUE #956: Need a localization support for datapicker
We need a localization  support for datepicker 
https://angular.carbondesignsystem.com/?path=/story/components-date-picker--single

We need support for the following languages `german` `finnish` `japanese` `mandarin` `spanish` `french` `italian` `korian` `portuguese`.

Donisius:
Closes #956


src/datepicker/datepicker.component.ts
```diff
@@ -16,6 +16,7 @@ import rangePlugin from "flatpickr/dist/plugins/rangePlugin";
 import flatpickr from "flatpickr";
 import { NG_VALUE_ACCESSOR } from "@angular/forms";
 import { carbonFlatpickrMonthSelectPlugin } from "./carbon-flatpickr-month-select";
+import * as languages from "flatpickr/dist/l10n/index";
```
cal-smith: Could we export this from the datepicker module, and have the input take the full object instead?

src/datepicker/datepicker.component.ts
```diff
@@ -184,6 +193,8 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 	constructor(protected elementRef: ElementRef) { }
 
 	ngOnChanges(changes: SimpleChanges) {
+		// Global localization of flatpickr instances
+		flatpickr.localize(languages.default[this.language]);
```
cal-smith: This changes the config globally ... which is probably what we want in most cases, but slightly inconsistent with how the options are presented (and doesn't allow for per-calendar language overrides) ... this should be something handled by the `flatpickrOptions` getter

Global translations are handled by the `i18n` service ... that's out of scope for this PR though, we can look at that later

src/datepicker/datepicker.component.ts
Donisius: When I try to export from the datepicker module I get the error:

Donisius: I get the following error when I export FlatpickrLanguages from the datepicker module:
```
Type 'Record<key, CustomLocale>' is not assignable to type 'any[] | Type<any>'.
  Type 'Record<key, CustomLocale>' is missing the following properties from type 'Type<any>': apply, call, bind, prototype, and 5 more
```

cyrusfurtado:
hi. Any idea when this would be merged.

src/datepicker/datepicker.component.ts
cal-smith: `export * as languages from "flatpickr/dist/l10n/index";` or even just leaving the import and doing `export languages;` causes that?

cal-smith:
Cool! thanks for that @cyrusfurtado, there's definitely some useful features from #976 we should be able to incorporate

src/datepicker/datepicker.component.ts
cyrusfurtado: Changing the language config globally should be left as an option to be done in the DatePicker module imho as in my PR here https://github.com/IBM/carbon-components-angular/pull/976/files#diff-1e9013bc3a56c5ec0f998d5ec536e11cL19.

src/datepicker/datepicker.module.ts
```diff
@@ -3,6 +3,9 @@ import { NgModule } from "@angular/core";
 import { CommonModule } from "@angular/common";
 import { DatePicker } from "./datepicker.component";
 
+import * as languages from "flatpickr/dist/l10n/index";
+export default languages;
```
cal-smith: We don't want to default export `languages` ... does `export languages;` or `export { languages };` not work?

src/datepicker/datepicker.stories.ts
```diff
@@ -150,6 +152,18 @@ storiesOf("Components|Date Picker", module)
 			</ibm-date-picker>
 		`,
 		props: {
+			language: select("Language of the calendar", [
```
zvonimirfras:
```suggestion
			language: select("Calendar language", [
```
? :)

src/datepicker/datepicker.stories.ts
```diff
@@ -150,6 +152,18 @@ storiesOf("Components|Date Picker", module)
...
+				"en",
```
zvonimirfras: Is it useful to spread these over new lines?

src/datepicker/datepicker.module.ts
Donisius: They both give errors.

`export languages;` gives the following error:
```
Declaration or statement expected
```

`export { languages };` gives: 
```
...datepicker.module has no default export.
```

src/datepicker/datepicker.stories.ts
Donisius: I suppose it looks better in one line 😄

src/datepicker/datepicker.module.ts
cal-smith: Hmm ... `export const languages = flatpickrLanguages;` perhaps? Weird that the object export wouldn't work

Donisius: That gives me `Individual declarations in merged declaration 'languages' must be all exported or all local.`

cal-smith: Okay now it's just getting silly

cal-smith: Let's skip that for now ... ideally it should work with the i18n service down the road, but we do need _something_ working in the short term

cal-smith: Don't default export the languages since we're not using them as inputs 👍 
```suggestion
```

cal-smith:
Also there's some conflicts in the datepicker component ... let me know if you need help sorting those out

cyrusfurtado:
Any idea when this could be available in master?

cal-smith:
Just waiting on one more review then we can get it in! Do you have a product deadline we should know about?

cyrusfurtado:
Is there a way to globally set the language for the `Datepicker` Module.

cal-smith:
Not at the moment, though that's something we're still sorting out ... Ideally it should be set through the `i18n` module, but given how datepicker applies the locale I'm not sure if that's going to be possible ... Definitly open an issue for this enhancement :+1:

PR #987, opened 2020-01-08
feat(ui shell): Add model to create header nav and sidenav items, add story to show responsive use of model

### LINKED ISSUE #732: header: header items disappear at 1056px view and lower
## Detailed description

The header items should be in the header at all times

Chrome

Donisius:
Closes #732 

#### Changelog

**New**

* Created option to create header menu, item, and populate header navigation using a model.
* Sidenav is able to take the model used to create header navigation to place the header navigation items within itself when the innerWidth of the window is small enough.


src/ui-shell/header/header-item.component.ts
```diff
@@ -47,6 +52,15 @@ export class HeaderItem {
...
 	@Input() routeExtras: any;
 
+	@Input() content: string;
+
+	/**
+	 * Set breakpoint to respond to the size of the window innerWidth. ie. the window
+	 * innerWidth to send the header item to the sidenav. Currently not used as styles
+	 * direct the response to the innerWidth of the window.
+	 */
+	@Input() breakpoint: number;
```
cal-smith: Since this is unused we can just remove it ... We'd likely want a more comprehensive solution for programmatic control and use of the carbon breakpoints anyway

src/ui-shell/header/header-menu.component.ts
```diff
@@ -44,6 +57,15 @@ export class HeaderMenu {
...
 	@Input() trigger: "click" | "mouseover" = "click";
 
+	@Input() headerItems: HeaderItemInterface[];
+
+	/**
+	 * Set breakpoint to respond to the size of the window innerWidth. ie. the window
+	 * innerWidth to send the header menu to the sidenav. Currently not used as styles
+	 * direct the response to the innerWidth of the window.
+	 */
+	@Input() breakpoint: number;
```
cal-smith: Same as the other `breakpoint`

src/ui-shell/header/header-menu.component.ts
```diff
@@ -44,6 +57,15 @@ export class HeaderMenu {
 	}
 	@Input() trigger: "click" | "mouseover" = "click";
 
+	@Input() headerItems: HeaderItemInterface[];
```
cal-smith: Maybe add some docs around this :+1: especially if this is the recommended usage model going forward

src/ui-shell/header/header-navigation-items.interface.ts
```diff
@@ -0,0 +1,20 @@
+// Used for both HeaderMenus and HeaderItems in the header navigation to generate
+// items and menus through a model.
```
cal-smith: This should be a full `/**` documentation block - that way it'll show up in the generated docs

src/ui-shell/header/header-navigation.component.ts
```diff
@@ -17,4 +39,6 @@ export class HeaderNavigation {
 	@HostBinding("style.height.%") height = 100;
 
 	@Input() ariaLabel: string;
+
+	@Input() navigationItems: NavigationItem[];
```
cal-smith: Same note here as for `headerItems`

src/ui-shell/header/header-navigation-items.interface.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface NavigationItem {
+    type: "menu | item",
+    breakpoint?: number;
+    title?: string,
+    href?: string,
+    trigger?: "click" | "mouseover",
+    route?: any[],
+    routeExtras?: any[],
+    content?: string
+    headerItems?: HeaderItemInterface[]
```
cal-smith: Maybe just `menuItems`?

src/ui-shell/sidenav/sidenav-item.interface.ts
```diff
@@ -0,0 +1,8 @@
+export interface SideNavItemInterface {
```
cal-smith: It may be possible to have `HeaderItemInterface` and `SideNavItemIterface` share an interface ... or at least have one extend the other ... Really they're both "menu items" and barring `isSubMenu` (which is a bit of a weird case anyway) they're already 99% identical 🤔 

They should definitely share a base interface at least though

src/ui-shell/sidenav/sidenav.component.ts
```diff
@@ -22,6 +23,23 @@ import { I18n } from "../../i18n/i18n.module";
...
+							*ngIf="navigationItem.type === 'item'"
+							[href]="navigationItem.href"
+							[route]="navigationItem.route"
+							[routeExtras]="navigationItem.routeExtras"
+							[content]="navigationItem.content">
+						</ibm-sidenav-item>
+						<ibm-sidenav-menu
+							*ngIf="navigationItem.type === 'menu'"
+							[title]="navigationItem.title"
+							[subMenuItems]="navigationItem.headerItems">
```
cal-smith: It's already a sub-menu right? So maybe just `menuItems`?

cal-smith:
Overall looks good though! I think this is basically the right approach for this situation, we'll just need to document it well.

cal-smith:
One thing to check @Donisius ... if there's no sidenav, but there are menu items, what happens to the menu items?

Donisius:
> One thing to check @Donisius ... if there's no sidenav, but there are header menu items, what happens to the menu items?

In order for the header items to appear on the side nav, the header items also need to be passed into the side nav. The css makes the items disappear from the header and appear on the side nav when window is small enough.

cal-smith:
> In order for the header items to appear on the side nav, the header items also need to be passed into the side nav. The css makes the items disappear from the header and appear on the side nav when window is small enough.

Okay ... I think there may be some other tweaks required to make the mobile UI 💯 but for the intent of this PR I think we're good to go

src/ui-shell/sidenav/sidenav-item.component.ts
```diff
@@ -58,19 +63,21 @@ export class SideNavItem {
 	 */
 	@Input() route: any[];
 
+	@Input() content: string;
```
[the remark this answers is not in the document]
Donisius: I tried passing in the content through `ng-content` by taking the `TemplateRef` of the header items and header menus but that caused a lot of issues. So we opted for a model approach. I thought that it would be easier for the user to pass in the same model to both the sidenav and header but let me know if there is a better way

PR #991, opened 2020-01-09
feat(Notification): Add support for custom content and add story example

### LINKED ISSUE #985: Toast and notification should support template content
Both toast and notification should support passing a template for the content area. We should likely skip the header/message/etc and just allow the user to specify whatever fields and controls they need.

Example from @awackygamer

image

image

(note that those are toasts, not notifications, and they have fully custom content)

Donisius:
Closes #985

### DIFF
--- a/src/notification/notification.component.ts
+++ b/src/notification/notification.component.ts
@@ -39,7 +39,7 @@ import { of, isObservable, Subject } from "rxjs";
 			<div class="bx--inline-notification__text-wrapper">
 				<p [innerHTML]="notificationObj.title" class="bx--inline-notification__title"></p>
 				<p [innerHTML]="notificationObj.message" class="bx--inline-notification__subtitle"></p>
-				<ng-content></ng-content>
+				<ng-container *ngTemplateOutlet="notificationObj.content"></ng-container>
 			</div>
 		</div>
 		<div *ngFor="let action of notificationObj.actions">
--- a/src/notification/notification.component.ts
+++ b/src/notification/notification.component.ts
@@ -36,11 +36,11 @@ import { of, isObservable, Subject } from "rxjs";
 				*ngIf="notificationObj.type === 'success'"
 				class="bx--inline-notification__icon">
 			</ibm-icon-checkmark-filled16>
-			<div class="bx--inline-notification__text-wrapper" *ngIf="!notificationObj.template">
+			<div class="bx--inline-notification__text-wrapper">
 				<p [innerHTML]="notificationObj.title" class="bx--inline-notification__title"></p>
 				<p [innerHTML]="notificationObj.message" class="bx--inline-notification__subtitle"></p>
+				<ng-container *ngTemplateOutlet="notificationObj.template"></ng-container>
 			</div>
-			<ng-container *ngTemplateOutlet="notificationObj.template"></ng-container>
 		</div>
 		<div *ngFor="let action of notificationObj.actions">
 			<button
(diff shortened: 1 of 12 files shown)


src/notification/notification.component.ts
```diff
@@ -39,6 +39,7 @@ import { of, isObservable, Subject } from "rxjs";
 			<div class="bx--inline-notification__text-wrapper">
 				<p [innerHTML]="notificationObj.title" class="bx--inline-notification__title"></p>
 				<p [innerHTML]="notificationObj.message" class="bx--inline-notification__subtitle"></p>
+				<ng-content></ng-content>
```
cal-smith: We can't use `ng-content` with service created notifications/toasts, so this has to come in via a template as part of the `notificationObj`

Donisius: fixed!

src/notification/notification.stories.ts
```diff
@@ -233,29 +233,51 @@ storiesOf("Components|Notification", module)
...
-					Lorem ipsum dolor sit amet, consectetur adipiscing elit.
-					Ut scelerisque, risus eu tincidunt faucibus
-				</p>
+			<ibm-toast [notificationObj]="{
+				type: 'error',
+				title: 'Sample custom toast',
+				subtitle: 'Sample subtitle message',
+				caption: 'Sample caption',
+				lowContrast: lowContrast,
+				content: customToastContent,
```
cal-smith: Going to be picky - searching through the codebase we have one existing instance of `obj.content` (in the dialog) and several instances of `obj.template` ... I'm going to suggest that we go with `obj.template` to _try_ and keep things consistent

cal-smith: also, do we want to keep the default layout when a template is used? It's possible a user would want full control over the content portion of a notification ... we can pass the `notificationObj` through as template context

src/notification/notification.component.ts
```diff
@@ -39,6 +39,7 @@ import { of, isObservable, Subject } from "rxjs";
 			<div class="bx--inline-notification__text-wrapper">
 				<p [innerHTML]="notificationObj.title" class="bx--inline-notification__title"></p>
```
cal-smith: We can use the new directives here too :+1:

src/notification/toast.component.ts
```diff
@@ -22,21 +22,22 @@ import { I18n } from "./../i18n/i18n.module";
...
+		<ibm-icon-warning-filled16
+			*ngIf="notificationObj.type === 'warning'"
+			class="bx--toast-notification__icon">
+		</ibm-icon-warning-filled16>
+		<ibm-icon-checkmark-filled16
+			*ngIf="notificationObj.type === 'success'"
+			class="bx--toast-notification__icon">
+		</ibm-icon-checkmark-filled16>
 		<div class="bx--toast-notification__details">
 			<h3 class="bx--toast-notification__title" [innerHTML]="notificationObj.title"></h3>
```
cal-smith: Same here, we can use the new directives

src/table/head/table-head-checkbox.component.ts
```diff
@@ -52,7 +51,7 @@ export class TableHeadCheckbox {
 
 	@HostBinding("class.bx--table-column-checkbox") hostClass = true;
 
-	protected _ariaLabel = this.i18n.getOverridable("TABLE.CHECKBOX_HEADER");
+	protected _ariaLabel: Overridable = this.i18n.getOverridable("TABLE.CHECKBOX_HEADER");
```
cal-smith: Where did all these changes come from?

Donisius: Not entirely sure 🤔

cal-smith: 🤔 🤔 🤔 🤔

PR #1004, opened 2020-01-16
fix(code snippet): add expansion button AfterViewInit to avoid ExpressionChangedAfterItHasBeenCheckedError

Donisius:
When a multi line code snippet is created dynamically, for example in an `ngIf`, rendering of the expansion button causes `ExpressionChangedAfterItHasBeenCheckedError` since its' existence depends on the view. It now behaves as expected.


Donisius:
> This seems far more complex than it needs to be. Why not just us a `hasExpandButton: boolean` and flip it's state in `ngAfterViewInit`? We don't need dynamic component generation, we just need to delay the `ngIf` to another change detection cycle 👍

It worked!

src/code-snippet/code-snippet.component.ts
```diff
@@ -193,6 +192,14 @@ export class CodeSnippet {
 		}, this.feedbackTimeout);
 	}
 
+	ngAfterViewInit() {
+		setTimeout(() => {
+			if (this.code ? this.code.nativeElement.getBoundingClientRect().height > 255 : false && this.display === "multi") {
```
cal-smith: This is likely a touch more readable

```suggestion
			if ((this.code && this.code.nativeElement.getBoundingClientRect().height > 255) && this.display === "multi") {
```

src/code-snippet/code-snippet.component.ts
```diff
@@ -137,11 +138,19 @@ export class CodeSnippet {
...
 
 	get shouldShowExpandButton() {
-		return this.code ? this.code.nativeElement.getBoundingClientRect().height > 255 : false;
+		// Checks if `hasExpand` button has been initialized in `AfterViewInit` before detecting whether or not to
+		// show the expand button when the code displayed in the component changes during the life of the component.
+		// This is to avoid the `ExpressionChangedAfterItHasBeenCheckedError`.
+		if (this.hasExpandButton === null) {
+			return this.hasExpandButton;
+		}
+		return this.code ? this.code.nativeElement.getBoundingClientRect().height > 255 : false && this.display === "multi";
```
cal-smith: Could we wrap this up in a protected method? That way we aren't duplicating code all over the place

Donisius: Just the `this.code ? this.code.nativeElement.getBoundingClientRect().height > 255 : false && this.display === "multi"` part? Any ideas for the name of this function ? 😅

src/code-snippet/code-snippet.component.ts
```diff
@@ -137,11 +138,19 @@ export class CodeSnippet {
 	@ViewChild("code") code;
 
 	get shouldShowExpandButton() {
-		return this.code ? this.code.nativeElement.getBoundingClientRect().height > 255 : false;
+		// Checks if `hasExpand` button has been initialized in `AfterViewInit` before detecting whether or not to
+		// show the expand button when the code displayed in the component changes during the life of the component.
+		// This is to avoid the `ExpressionChangedAfterItHasBeenCheckedError`.
+		if (this.hasExpandButton === null) {
+			return this.hasExpandButton;
```
cal-smith: does `null` behave differently to `false` in angular templates?

src/code-snippet/code-snippet.component.ts
cal-smith: Actually ... couldn't we just use this (`shouldShowExpandButton`) in `ngAfterViewInit`?

src/code-snippet/code-snippet.component.ts
Donisius: That is to prevent the change in `hasExpandButton` before it gets changes in `ngAfterViewInit`. Without the check it gives the expression changed error, and if its checking for false, the expand button will not return if it goes away until the component gets instantiated again

cal-smith: Huh, so angular does treat a `null` differently than a `false` then? Interesting!

src/code-snippet/code-snippet.component.ts
Donisius: Good idea!!

Donisius: Actually this only works after `hasExpandButton` has been initialized otherwise it will just keep returning null

src/code-snippet/code-snippet.component.ts
Donisius: In the template it does treat `null` as `false` and the expand button doesn't show up, when `hasExpandButton` is `null` it means that it has not been initialized in `ngAfterViewInit` yet and it will stay null until it does to avoid the expression changed error

cal-smith: Okay, so for clarity - if it was `false`, and stayed as `false` until `hasExpandButton` initialized, would that also work or no?

Donisius: If there was a way for `shouldShowExpandButton` to detect if it has been changed after `ngAfterViewInit` then yes, I currently have it so that it detects if it has not changed if it is still null.

cal-smith: okay ... I think I understand, we should be good to go then 👍

src/code-snippet/code-snippet.component.ts
cal-smith: Right ... maybe `canExpand` or something for the method name?

PR #1005, opened 2020-01-18
fix(date picker): Open picker popup when calendar icon is clicked

yinglejia:

Open picker popup when calendar icon is clicked in a single-mode date picker. This fixes a recent regression caused by previous PR https://github.com/IBM/carbon-components-angular/pull/939.

image

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -307,17 +307,22 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 	 * Handles opening the calendar "properly" when the calendar icon is clicked.
 	 */
 	openCalendar(datepickerInput: DatePickerInput) {
-		datepickerInput.input.nativeElement.click();
+		if (this.range) {
+			datepickerInput.input.nativeElement.click();
 
-		// If the first input's calendar icon is clicked when calendar is in range mode, then
-		// the month and year needs to be manually changed to the current selected month and
-		// year otherwise the calendar view will not be updated upon opening.
-		if (datepickerInput === this.input && this.range && this.flatpickrInstance.selectedDates[0]) {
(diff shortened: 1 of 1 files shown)


yinglejia:
> Thank you so much for catching this and providing a fix! Just one nitpick: instead of opening the calendar with `click` for `range` and `flatpickrInstance.open()` for `single-mode`, opening the calendar with `flatpickrInstance.open()` for both _should_ work.

Thought about that. OK, I'll update the code to do that.

src/datepicker/datepicker.component.ts
```diff
@@ -307,7 +307,7 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked {
 	 * Handles opening the calendar "properly" when the calendar icon is clicked.
 	 */
 	openCalendar(datepickerInput: DatePickerInput) {
-		datepickerInput.input.nativeElement.click();
+		this.flatpickrInstance.open();
```
cal-smith: @Donisius I believe you changed this last ... do you remeber why we switched to `datepickerInput.input.nativeElement.click` from `this.flatpickrInstance.open`?

Donisius: The months do not get updated properly when the calendar is opened through `this.flatpickrInstance.open` when the calendar is in range mode but it needs to be opened through `this.flatpickrInstance.open` when the datepicker is in single mode otherwise it will not open when the calendar icon is clicked. @yinglejia Sorry, it was perfect the way it was originally.

yinglejia: OK, I'll revert the last commit then.

yinglejia:
@cal-smith OK to merge this?

PR #1009, opened 2020-01-21
feat(dropdown): flip position of dropdown menu if outside of right view port

### LINKED ISSUE #986: Dropdown: Allow for left/right alignment change
## Detailed description

We're using the dropdown component close to the screen edge and it ends up being cut (also adding a horizontal scrollbar on the page). We'd like to have the dropdown detect it will go show out of the page bounds and right align itself so there's no cut out content. If that doesn't work then a manual option to set it to right aligned is also good. Same idea would apply for detecting also the bottom edge of the page so it pops upward instead of downward.

image

Feature

Dropdown

Chrome

    "carbon-components": "10.7.4",
    "carbon-components-angular": "3.30.0",

IBM ADDI

Donisius:
Closes #986

### DIFF
--- a/src/dropdown/dropdown.service.ts
+++ b/src/dropdown/dropdown.service.ts
@@ -100,8 +100,29 @@ export class DropdownService implements OnDestroy {
 	}
 
 	protected positionDropdown(parentRef, menuRef) {
+		let leftOffset = 0;
+		const boxMenu = menuRef.querySelector(".bx--list-box__menu");
+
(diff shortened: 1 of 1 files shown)


Donisius:
> Looks good! Just one note - It doesn't update when I resize the window ... I'm not sure what's going on with that since it works correctly when the menu opens, so (logically) it should continuously update the position as the window resizes

Fixed...hopefully 🤞

PR #1014, opened 2020-01-23
fix(table): bubble up click event when row item is clicked

### LINKED ISSUE #944: Need to bubble up rowClick event on the click of table row
For  https://angular.carbondesignsystem.com/?path=/story/table--with-toolbar ,  
   need to bubble up separate `rowClick` event on the table row click without changing any configuration to toggle between checkbox select and row select 

## Detailed description

feature request

https://angular.carbondesignsystem.com/?path=/story/table--with-toolbar
https://github.com/IBM/carbon-components-angular/blob/master/src/table/table.component.ts

Expected : Table  should be like gmail inbox where clicking on the row opens up the email and clicking on the checkbox selects it.  Should be able to do both simultaneously
Happened :  `singleSelect` enabled and `showSelectionColumn` disabled ,   `selectRow` is emitted when the row is clicked.  checkbox select and row select should not be mutually exclusive.
Requirement : Need to be able to select checkboxes as well as rows with the same config.

Chrome

carbon-components-angular@3.28.0

ECAM.

Donisius:
Closes #944

### DIFF
--- a/src/table/body/table-row.component.ts
+++ b/src/table/body/table-row.component.ts
@@ -60,7 +60,8 @@ import { Observable } from "rxjs";
 					[skeleton]="skeleton"
 					[attr.colspan]="item.colSpan"
 					[attr.rowspan]="item.rowSpan"
-					(click)="onRowItemClick()">
+					(click)="onRowClick()"
+					(keydown.enter)="onRowClick()">
 				</td>
 				<td
 					*ngIf="item && model.getHeader(j) == null"
@@ -70,7 +71,8 @@ import { Observable } from "rxjs";
 					[skeleton]="skeleton"
 					[attr.colspan]="item.colSpan"
 					[attr.rowspan]="item.rowSpan"
-					(click)="onRowItemClick()">
+					(click)="onRowClick()"
+					(keydown.enter)="onRowClick()">
 				</td>
 			</ng-container>
 		</ng-container>
@@ -158,7 +160,7 @@ export class TableRowComponent {
 	 * SHould only get emitted when a row item is selected excluding expandButtons,
 	 * checkboxes, or radios.
 	 */
-	@Output() rowItemClick = new EventEmitter();
+	@Output() rowClick = new EventEmitter();
 
 	@HostBinding("class.bx--data-table--selected") get selectedClass() {
 		return this.selected;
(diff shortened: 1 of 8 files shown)


src/table/body/table-row.component.ts
```diff
@@ -151,6 +155,13 @@ export class TableRowComponent {
 	 */
 	@Output() expandRow = new EventEmitter();
 
+	/**
+	 * Emits when a row is not dependant on enableSingleSelect or showSelectionColumn.
+	 * SHould only get emitted when a row item is selected excluding expandButtons,
```
zvonimirfras:
```suggestion
	 * Should only get emitted when a row item is selected excluding expand buttons,
```

src/table/body/table-row.component.ts
```diff
@@ -151,6 +155,13 @@ export class TableRowComponent {
 	 */
 	@Output() expandRow = new EventEmitter();
 
+	/**
+	 * Emits when a row is not dependant on enableSingleSelect or showSelectionColumn.
```
zvonimirfras:
```suggestion
	 * Emits when a row is not dependant on `enableSingleSelect` or `showSelectionColumn`.
```

src/table/body/table-row.component.ts
```diff
@@ -59,7 +59,9 @@ import { Observable } from "rxjs";
 					[ngStyle]="model.getHeader(j).style"
 					[skeleton]="skeleton"
 					[attr.colspan]="item.colSpan"
-					[attr.rowspan]="item.rowSpan">
+					[attr.rowspan]="item.rowSpan"
+					(click)="onRowClick()"
```
zvonimirfras: So adding these once on `tr` instead of every `td` doesn't work?

Donisius: Adding them once on `tr` would also get triggered if a selection/expansion column was clicked

src/table/body/table-row.component.ts
```diff
@@ -151,6 +155,13 @@ export class TableRowComponent {
 	 */
 	@Output() expandRow = new EventEmitter();
 
+	/**
+	 * Emits when a row is not dependant on `enableSingleSelect` or `showSelectionColumn`.
```
cal-smith:
```suggestion
	 * Emits when a row is clicked regardless of `enableSingleSelect` or `showSelectionColumn`.
```

sparikh1594:
Also waiting for this to be delivered

PR #1015, opened 2020-01-23
feat(modal): add size property

### LINKED ISSUE #1002: Missing binding to size in Modal component
## Detailed description

The Modal component is missing binding to the size property.
https://www.carbondesignsystem.com/components/modal/style#sizes

This is a bug.

The modal component.

Setting the size should reflect change on modal size.

Version 3.32.1.

Donisius:
Closes #1002

### DIFF
--- a/src/modal/modal.component.ts
+++ b/src/modal/modal.component.ts
@@ -86,6 +86,11 @@ export class ModalDemo {
 		<ibm-overlay [theme]="theme" (overlaySelect)="overlaySelected.emit()">
 			<div
 				class="bx--modal-container"
+				[ngClass]="{
+					'bx--modal-container--xs': size === 'xs',
+					'bx--modal-container--sm': size === 'sm',
+					'bx--modal-container--lg': size === 'lg'
+				}"
 				[@modalState]="modalState"
 				role="dialog"
 				aria-modal="true"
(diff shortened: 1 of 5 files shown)


src/modal/modal.component.ts
```diff
@@ -112,7 +117,7 @@ export class Modal implements AfterViewInit, OnInit, OnDestroy {
 	/**
 	 * Size of the modal to display.
 	 */
-	@Input() size: "sm" | "md" | "lg" | "xl" | "xxl" = "md";
+	@Input() size: "xs" | "sm" | "md" | "lg" | "xl" | "xxl";
```
cal-smith: We don't support an `xxl` size right? We should be able to drop that

cal-smith: Or at the very least mention that we don't actually support `xxl`

patricklowry: It would be good to reach some level of parity with [Carbon](https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/modal/_modal.scss) and the [React implementation](https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Modal/Modal.js#L138) here. In other words, this might be an opportunity to remove non-functional code without introducing breaking changes while aligning with Carbon (the source of truth) and a similar component implementation.

E.g. the modal has a default size from `bx--modal-container` (what this calls "md") and three size variants ("xs", "sm", and "lg") as the _only_ options for `size`. If a developer specifies a non-Carbon size it could just default to the default size like always.

I'm also not sure it's safe to assume that "md" is the default for all components (e.g. I would expect a Carbon button size variant that sits between the default and "sm" would be "md"), but that's a topic for a different issue!

cal-smith: Great points @patricklowry ... since this code was non-functional until this point, It shouldn't be a breaking change to simplify the type 👍

Donisius:
> Did you check this with the starter app @Donisius?

Seems to build fine when setting size to anything other than `xs, sm, lg` with the `--prod` and `--aot` flags.

PR #1017, opened 2020-01-24
feat(tabs): add cacheActive support for tab header group

Donisius:


src/tabs/tab-header-group.component.ts
```diff
@@ -192,6 +198,10 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy {
 	ngAfterContentInit() {
 		this.selectedSubscriptionTracker.unsubscribe();
 
+		if (this.tabHeaderQuery) {
+			this.tabHeaderQuery.toArray().forEach(tabHeader => tabHeader.cacheActive = this.cacheActive);
```
zvonimirfras: This should be done in a way that it updates with changes in `cacheActive` during the lifetime of the component.

src/tabs/tab-header.component.ts
```diff
@@ -48,13 +49,23 @@ export class TabHeader {
...
 	/**
 	 * Value 'selected' to be emitted after a new `Tab` is selected.
 	 */
 	@Output() selected = new EventEmitter<any>();
 
 	@ViewChild("tabItem") tabItem: ElementRef;
 
+	ngAfterViewInit() {
+		if (this.paneReference) {
+			this.paneReference.cacheActive = this.cacheActive;
```
zvonimirfras: This should be done in a way that it updates with changes in cacheActive during the lifetime of the component.

src/tabs/tab-header-group.component.ts
Donisius: Fixed!

src/tabs/tab-header-group.component.ts
```diff
@@ -86,6 +87,19 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy {
 	@Input() contentAfter: TemplateRef<any>;
 
 	@Input() contentBefore: TemplateRef<any>;
+
+		/**
```
cal-smith: some weird tabbing going on here ...

```suggestion
	/**
```

src/tabs/tab-header-group.component.ts
```diff
@@ -86,6 +87,19 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy {
...
+	 * Set to 'true' to have all pane references associated with each tab header
+	 * in the tab header group cached and not reloaded on tab switching.
+	 */
+	@Input() set cacheActive(shouldCache: boolean) {
+		this._cacheActive = shouldCache;
+	}
+
+	get cacheActive() {
+		return this._cacheActive;
+	}
```
cal-smith: If this has no other effect, could it just be a normal property? i.e. not a getter/setter?

src/tabs/tab-header-group.component.ts
```diff
@@ -211,6 +231,12 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy {
 		this.tabHeaderQuery.toArray()[this.currentSelectedIndex].selectTab();
 	}
 
+	ngOnChanges() {
+		if (this.tabHeaderQuery) {
+			this.tabHeaderQuery.toArray().forEach(tabHeader => tabHeader.cacheActive = this.cacheActive);
+		}
```
cal-smith: probably want to check if `cacheActive` is actually in the changeset, otherwise this could end up running way more frequently than necessary

src/tabs/tab-headers.component.ts
```diff
@@ -250,6 +251,12 @@ export class TabHeaders implements AfterContentInit {
 		this.setFirstTab();
 	}
 
+	ngOnChanges() {
+		if (this.tabs) {
+			this.tabs.forEach(tab => tab.cacheActive = this.cacheActive);
```
cal-smith: Same note here as in `tab-header-group.component`

src/tabs/tab.component.ts
```diff
@@ -104,7 +104,9 @@ export class Tab implements OnInit {
 	/**
 	 * Set to true to have Tab items cached and not reloaded on tab switching.
 	 */
-	@Input() cacheActive = false;
+	@Input() set cacheActive(shouldCache: boolean) {
+		this._cacheActive = shouldCache;
+	}
```
cal-smith: I'm not sure this needs to be a getter/setter?

src/tabs/tabs.component.ts
```diff
@@ -119,6 +119,12 @@ export class Tabs implements AfterContentInit {
 		});
 	}
 
+	ngOnChanges() {
+		if (this.tabHeaders) {
+			this.tabHeaders.cacheActive = this.cacheActive;
```
cal-smith: Same note here as in `tab-header-group.component`

src/tabs/tab-header-group.component.ts
Donisius: True, since the input is being changed directly and that would get picked up by angular's change detection.

src/tabs/tab.component.ts
Donisius: I'm using `tab` in the `tab-header-group` method so it needs the setter to actually trigger the change detection since if `cacheActive` was just an input it would not get changed directly

cal-smith: Interesting ... there shouldn't be a difference between setter/getter and properties, but if it's something you've noticed is necessary then I'm fine with it 👍

Donisius: Yeah since its being changed like `this.paneReference.cacheActive = this.cacheActive`, I guess it doesn't count as directly changing the property and for some reason it angular doesn't detect any changes

PR #1018, opened 2020-01-27
fix(table): check/change scrollbar width in the next cycle

### LINKED ISSUE #1016: using StickyHeader attribute Getting ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'ngIf: 0'. Current value: 'ngIf: 10'.
## Detailed description

> Sticker Header is working with basic table but with table having expansion panel it throwing the error and disturbing the UI table

Yes This issue is with <ibm-table>

Wanted stickyheader for all kinds of tables with out affecting the CSS design
Chrome, mozilla,IE

        "carbon-components": "^10.9.2",
        "carbon-components-angular": "^3.36.0",

  <ibm-table [striped]="true" **class="display_block"** **[stickyHeader]="true"** [model]="model" [showSelectionColumn]="false" (sort)="simpleSort($event)">
            </ibm-table>

Adding stickyHeader getting the error, 
sticky

* Screenshots or code
* Notes

Donisius:
Closes #1016

### DIFF
--- a/src/table/head/table-head.component.ts
+++ b/src/table/head/table-head.component.ts
@@ -171,7 +171,9 @@ export class TableHead implements AfterViewInit {
 	constructor(protected i18n: I18n) {}
 
 	ngAfterViewInit() {
-		this.scrollbarWidth = getScrollbarWidth();
+		setTimeout(() => {
+			this.scrollbarWidth = getScrollbarWidth();
+		});
 	}
 
 	onSelectAllCheckboxChange() {


src/table/head/table-head.component.ts
```diff
@@ -171,7 +171,9 @@ export class TableHead implements AfterViewInit {
 	constructor(protected i18n: I18n) {}
 
 	ngAfterViewInit() {
-		this.scrollbarWidth = getScrollbarWidth();
+		setTimeout(() => {
+			this.scrollbarWidth = getScrollbarWidth();
+		});
```
theiliad: setTimeout usually should be avoided but if you think it's really necessary I'll approve

cal-smith: In this case we're using `setTimeout` to hook into the Angular change detection, and schedule an update for the next tick ... this usually implies an error in how the values are being propagated, though in this case that may not be true

PR #1025, opened 2020-01-29
fix(dropdown): Externalize clear button tooltip text.

stanislavgeorgiev:
Closes IBM/carbon-components-angular#1024


src/dropdown/dropdown.component.ts
```diff
@@ -71,7 +71,7 @@ import { ElementService } from "./../utils/utils.module";
 				(click)="clearSelected()"
 				*ngIf="type === 'multi' && getSelectedCount() > 0"
 				class="bx--list-box__selection--multi"
-				title="Clear all selected items">
+				title="clearText">
```
theiliad: doesn't that need to be in `[]`?

stanislavgeorgiev: Yea, you're right! Sorry about that!

PR #1032, opened 2020-02-03
fix(modal): tentative approach to overriding modal closing on overlay click

### LINKED ISSUE #1027: Modal always dismisses itself when clicking outside of it
Modal always dismisses itself when clicking outside of it

## Detailed description

I am creating an alert modal using ModalService.show(...) and when this modal is displayed, it will be dismissed when the user clicks on any area outside of it.
Bug
Modal
Clicking outside this modal should not dismiss the modal.(at least it should be optional to dismiss it)
3.30.0
IBM Application Delivery Intelligence
Ships early March 2020.
1) The issue can me seen using the story book in Modal-Transactional at https://angular.carbondesignsystem.com/?path=/story/components-modal--transactional
2) Click the "Open Modal" button
3) Once the modal is displayed, click somewhere outside the modal and the modal disappears.

## Additional information
I think the bullet "give an escape hatch for applications that don't want that (maybe overriding some method?)" of this previous issue comment may not have been addressed:
https://github.com/IBM/carbon-components-angular/issues/505#issuecomment-484240090

HYACCCINT:
Closes #1027

{{Tentative approach to overriding modal closing on overlay click}}

Changelog
New

{{
Alert modal:
Input field onOverlayClick : string
function handleClickOverlay
}}**
Changed

{{change thing}}
Removed

{{removed thing}}

### DIFF
--- a/src/modal/alert-modal.component.ts
+++ b/src/modal/alert-modal.component.ts
@@ -85,7 +85,7 @@ export class AlertModal extends BaseModal implements AfterViewInit {
 		@Inject("modalContent") public modalContent: string,
 		@Inject("size") public size: string,
 		@Inject("buttons") public buttons = [],
-		@Inject("close") public onClose: Function,
+		@Inject("close") public onClose: Function
 	) {
 		super();
 		for (let i = 0; i < this.buttons.length; i++) {
--- a/src/modal/alert-modal.component.ts
+++ b/src/modal/alert-modal.component.ts
@@ -51,9 +51,8 @@ import { BaseModal } from "./base-modal.class";
 			[size]="size"
 			[theme]="modalType"
 			[modalLabel]="modalTitle"
-			(overlaySelected)="handleClickOverlay()">
-
-			<ibm-modal-header (closeSelect)="dismissModal()">
+			(overlaySelected)="dismissModal('overlay')">
+			<ibm-modal-header (closeSelect)="dismissModal('close')">
 				<p class="bx--modal-header__label bx--type-delta">{{modalLabel}}</p>
       			<p class="bx--modal-header__heading bx--type-beta">{{modalTitle}}</p>
 			</ibm-modal-header>
@@ -87,7 +86,6 @@ export class AlertModal extends BaseModal implements AfterViewInit {
 		@Inject("size") public size: string,
 		@Inject("buttons") public buttons = [],
 		@Inject("close") public onClose: Function,
-		@Inject("onOverlayClick") public onOverlayClick: string
 	) {
 		super();
 		for (let i = 0; i < this.buttons.length; i++) {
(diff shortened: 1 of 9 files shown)


src/modal/modal.component.ts
```diff
@@ -128,6 +128,11 @@ export class Modal implements AfterViewInit, OnInit, OnDestroy {
 	 */
 	@Input() modalLabel = "default";
 
+	/**
+	 * Determins if the modal closes on an overlay click.
+	 */
+	@Input() onOverlayClick: "default" | "override" = "default";
```
cal-smith: Modal doesn't need this, since the user can choose how to handle `overlaySelected` outside of the component

src/modal/alert-modal.interface.ts
```diff
@@ -48,6 +48,12 @@ export interface AlertModalData {
 	 * Callback for non-specific close events. `return false;` to prevent the modal from closing
 	 */
 	close?: Function;
```
cal-smith: Perhaps changing the signature to be something like:
```suggestion
	close?: ({ trigger: "overlay" | "close" }) => boolean;
```
Would be more flexible, and perhaps a touch more explicit with regard to what we're actually handling

src/modal/alert-modal.interface.ts
```diff
@@ -48,6 +48,12 @@ export interface AlertModalData {
...
+
+	/**
+	 * Determines if modal closes on overlay click
+	 */
+
+	onOverlayClick?: string;
```
cal-smith: With the above change we can drop this

src/modal/alert-modal.component.ts
```diff
@@ -124,4 +126,11 @@ export class AlertModal extends BaseModal implements AfterViewInit {
 		}
 		this.closeModal();
 	}
+
+	handleClickOverlay() {
+		if (this.onOverlayClick === "override") {
```
cal-smith: This gets the job done, but I'm concerned that we now have two different ways to handle controlling how a modal closes that do very different things and could lead to unexpected bugs on the consumer side. We already have a `close` function that controls how/when the modal closes, so it may be clearer to take advantage of that over introducing new properties and methods

HYACCCINT: Utilizing existing functions is definitely better than introducing new methods, I should have considered this more carefully :p 
To clarify, the trigger parameter in the close signature suggested would be part of user input?

src/modal/alert-modal.component.ts
```diff
@@ -118,10 +118,11 @@ export class AlertModal extends BaseModal implements AfterViewInit {
 		this.closeModal();
 	}
 
-	dismissModal() {
+	dismissModal(trigger) {
```
Donisius: At the moment the trigger parameter doesn't appear to be doing anything. I think onClose() needs to take in trigger as an argument?

src/modal/alert-modal.interface.ts
```diff
@@ -46,8 +46,10 @@ export interface AlertModalData {
 	buttons?: Array<ModalButton>;
 	/**
 	 * Callback for non-specific close events. `return false;` to prevent the modal from closing
+	 * trigger : "overlay" | "close"
 	 */
-	close?: Function;
+	close?: (trigger: string) => {};
```
Donisius:
```suggestion
	close?: ({ trigger: "overlay" | "close" }) => boolean
```

src/modal/alert-modal.component.ts
cal-smith: In the case of `alert-modal` the overlay click and the close button click are handled the same way, so we need a way for users to differentiate 👍

PR #1034, opened 2020-02-04
feat(input): Typeahead support for Search Component

cyrusfurtado:
Partially addresses: https://github.com/IBM/carbon-components-angular/issues/997

Added a typeahead support for the `<ibm-search>` component.

#### Changelog

**New**

* Boolean Input for `typeahead`
* `TypeAheadResults` array passed in from parent component
* `debounceTime` for specifying a non standard debounce interval for the `valueChange` output event

**Changed**

* Changed the way the `valueChange` event is fired using Observables for unique values only after a default debounce time of 500ms.

**Screenshots**

pr_1034

### DIFF
--- a/src/search/search.component.html
+++ b/src/search/search.component.html
@@ -36,7 +36,7 @@
 		>
 			<li
 				(click)="updateSearchResult(option)"
-				*ngFor="let option of typeAheadResults"
+				*ngFor="let option of _typeAheadResults"
 				class="bx--list-box__menu-item"
 			>
 				<div
@@ -44,7 +44,9 @@
 					tabindex="-1"
 					class="bx--list-box__menu-item__option"
 				>
-					<ng-container *ngIf="!listTpl">{{ option }}</ng-container>
+					<ng-container *ngIf="!listTpl">
+						<span [innerHtml]="option.html"></span>
+					</ng-container>
 					<ng-template 
 						*ngIf="listTpl" 
 						[ngTemplateOutletContext]="{item: option}"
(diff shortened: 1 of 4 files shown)


cyrusfurtado:
So could it be something like a `directive` that can be added to any input of type text

cal-smith:
> So could it be something like a directive that can be added to any input of type text. Also didn't find any references to AbstractSearchPane.

`AbstractSearchPane` would be a new class that follows the precedent set by `AbstractDropdownView` - a common base class we can query for and call standardized methods on.

It's possible that a directive would also work (linking two components together that way 🤔)

So there's a few options we should explore that will give us more flexibility

cyrusfurtado:
So `AbstractSearchPane` would be a new class to be implemented for `Typeahead` support the way `AbstractDropdownView` is implemented in `DropdownList` aka `ibm-dropdown-list`.

cal-smith:
> So AbstractSearchPane would be a new class to be implemented for Typeahead support the way AbstractDropdownView is implemented in DropdownList aka ibm-dropdown-list. 

Exactly, that way whatever component or directive is used to control the search pane has a consistent reference to query for, and a consistent API to use. And that way consumers aren't locked into one or two patterns for search results.

>  If not a directive then do we add something like a typeahead-list component to the DOM.

Yep! There's a few different ways to wire it up (directives, models, components in `ng-content`) so it's really whatever works best and gives the consumer the most control

cal-smith:
Closing this for now since it's been open for a while with no movement ... Totally open to a updated PR though!

PR #1037, opened 2020-02-04
fix(table): allow span to be included in the table header in skeleton mode

### LINKED ISSUE #1036: Data Table: Animation is removed from skeleton view of the table

The loading animation is removed from the skeleton table. The only way for the loading animation to get applied is if a span exists within the table header/body elements.

Bug.

Table.

Loading animation should appear in skeleton table.

Chrome, Firefox

10.9.2

Donisius:
Closes #1036

### DIFF
--- a/src/table/head/table-head-cell.component.ts
+++ b/src/table/head/table-head-cell.component.ts
@@ -17,7 +17,7 @@ import { TableHeaderItem } from "./../table-header-item.class";
 	template: `
 		<button
 			class="bx--table-sort"
-			*ngIf="sortable && !skeleton && this.sort.observers.length > 0 && column.sortable"
+			*ngIf="sortable && this.sort.observers.length > 0 && column.sortable"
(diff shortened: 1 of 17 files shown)


src/table/head/table-head-cell.component.ts
```diff
@@ -17,7 +17,7 @@ import { TableHeaderItem } from "./../table-header-item.class";
 	template: `
 		<button
 			class="bx--table-sort"
-			*ngIf="sortable && !skeleton && this.sort.observers.length > 0 && column.sortable"
+			*ngIf="sortable && this.sort.observers.length > 0 && column.sortable"
```
cal-smith: I _think_ the `!skeleton` was added to sort out a different skeleton bug ... maybe check the commit history for this to verify

Donisius: I think it was suppose to be removed but was added back by mistake in a future commit

PR #1038, opened 2020-02-04
fix(tooltip): close all other known dialogs on hover trigger

### LINKED ISSUE #1023: Tooltip - Not hiding on hover out.
I am using tooltip on hover (trigger:hover).

**Issue:**
When i click instead of hover on the label, tooltip stays on the UI even I hover out or click outside.

**Expected behaviour:**
Must be removed on hover out.

Tooltip

Donisius:
Closes #1023

### DIFF
--- a/src/dialog/dialog.service.ts
+++ b/src/dialog/dialog.service.ts
@@ -193,17 +193,28 @@ export class DialogService {
 	}
 
 	/**
-	 * Closes all known `Dialog`s. Does not focus any previous elements, since we can't know which would be correct
+	 * Closes all known `Dialog`s. Does not focus any previous elements, since we can't know which would be correct.
+	 * If `preventClosure` parameter is provided, it closes all know `Dialog`s except the given Dialog ref.
+	 * @param preventClosure the dialogRef to keep open while all others are closed
 	 */
-	closeAll() {
-		DialogService.dialogRefs.forEach(ref => ref.destroy());
-		DialogService.dialogRefs.clear();
-		DialogService.dialogCloseSubscription.unsubscribe();
-		this.isClosed.emit(true);
+	closeAll(preventClosure?: ComponentRef<Dialog>) {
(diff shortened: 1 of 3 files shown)


src/dialog/dialog.service.ts
```diff
@@ -193,17 +193,28 @@ export class DialogService {
...
-		DialogService.dialogRefs.forEach(ref => ref.destroy());
-		DialogService.dialogRefs.clear();
-		DialogService.dialogCloseSubscription.unsubscribe();
-		this.isClosed.emit(true);
+	 * Closes all known `Dialog`s. Does not focus any previous elements, since we can't know which would be correct.
+	 * If `preventClosure` parameter is provided, it closes all know `Dialog`s except the given Dialog ref.
+	 * @param preventClosure the dialogRef to keep open while all others are closed
+	 */
+	closeAll(preventClosure?: ComponentRef<Dialog>) {
+		// DialogService.dialogRefs.forEach(ref => ref.destroy());
```
cal-smith: On one hand we could probably remove this, on the other it's a pretty handy description of whats going on below, so 🤷‍♂

Donisius: I meant to remove it before I committed and pushed 🤦‍♂ but you're right it can be a pretty good description!

Donisius: On a side note... if there are two tooltips with a click trigger and there is no element within the dialog to focus it automatically focuses on the other element when you change which is opened. Not sure if that is intended but if you click one and press enter on the other while the other tooltip is focused it keeps both of them open and unable to close.

cal-smith: Can you demo this to me? It seems like a bug ... but it could also just be an awkward interaction pattern

zvonimirfras:
Looks like it's unsubscribing from the last open dialog subscription, not the one actually being closed, so it closes the dialog, and kills the subscription for the one that's left open.

Donisius:
Closing in favour of #1048

PR #1040, opened 2020-02-05
fix(notification-toast): sets showClose property to a default of true

HYACCCINT:
Closes IBM/carbon-components-angular#1039

{{Sets showClose property to a default of true}}

#### Changelog
Adds setter for showClose, sets it to a default of true
**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}


theiliad:
> We typically want to create the notification and toast components through the `notificationObj` input so it would be enough just to check if `showClose` was provided in the `notificationObj` and default it to true otherwise 👍

@Donisius are you asking for changes here then?

src/notification/notification.component.ts
```diff
@@ -104,6 +104,8 @@ export class Notification {
 		return this._notificationObj.showClose;
 	}
 
+
+
```
Donisius:
```suggestion
```

src/notification/notification.component.ts
```diff
@@ -53,7 +53,7 @@ import { of, isObservable, Subject } from "rxjs";
 			</button>
 		</div>
 		<button
-			*ngIf="showClose"
+			*ngIf="notificationObj.showClose"
```
Donisius: This will default to false, _maybe_ something like `notificationObj.showClose === null ? true : notificationObj.showClose`?

PR #1042, opened 2020-02-05
fix(button): add field size

cal-smith:
Closes IBM/carbon-components-angular#1041


cal-smith:
the "full bleed" buttons aren't supported outside of specific contexts - react doesn't seem to provide them as a normal button style ... I _think_ they may just be grid linked buttons, but I'm not totally sure

PR #1043, opened 2020-02-05
fix(modal): allow user to insert custom close function based on close event location

HYACCCINT:
Closes IBM/carbon-components-angular#1027

{{short description}}

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}


src/modal/modal.service.ts
```diff
@@ -9,6 +9,7 @@ import { Injectable } from "@angular/core";
 import { AlertModal } from "./alert-modal.component";
 import { AlertModalData } from "./alert-modal.interface";
 import { PlaceholderService } from "./../placeholder/placeholder.module";
+import { flattenStyles } from "@angular/platform-browser/src/dom/dom_renderer";
```
cal-smith: I don't think we need this ...
```suggestion
```

src/modal/alert-modal.interface.ts
```diff
@@ -46,8 +46,9 @@ export interface AlertModalData {
 	buttons?: Array<ModalButton>;
 	/**
 	 * Callback for non-specific close events. `return false;` to prevent the modal from closing
+	 * Property trigger has a value of `overlay` or `close`
 	 */
-	close?: Function;
+	close?: (trigger) => boolean;
```
cal-smith: We'll want to add the possible types here

```suggestion
	close?: (trigger: "overlay" | "close") => boolean;
```

HYACCCINT:
> Couple suggests, but otherwise looks good!

Interesting, I did not add the flattened styles... wonder where it came from 0.0

src/modal/alert-modal.component.ts
```diff
@@ -118,8 +118,8 @@ export class AlertModal extends BaseModal implements AfterViewInit {
 		this.closeModal();
 	}
 
-	dismissModal() {
-		if (this.onClose && this.onClose() === false) {
+	dismissModal(trigger) {
+		if (this.onClose && this.onClose(trigger) === false) {
 			return;
 		}
 		this.closeModal();
```
Donisius: Would
```
if (this.onClose && this.onClose(trigger)) {
	this.closeModal()
}
```
work?

HYACCCINT: Should work so long as onClose is always defined

Donisius: Okay, and since we don't want onClose to always be defined it wouldn't 👍

cal-smith:
> It looks like this might be a breaking change to some people?
> 
> Maybe we need to default to something or have an option (a default option perhaps) that does the old behavior?

I'm not sure how this changes any behaviour? If you don't do anything with the value passed in it will still always close the modal. We could make the `trigger` argument optional (`trigger?:`), since that may typecheck differently now, but otherwise there's be no actual behaviour changes

This is in `AlertModal`, not `Modal` if that clears things up?

src/modal/alert-modal.interface.ts
```diff
@@ -46,8 +46,9 @@ export interface AlertModalData {
 	buttons?: Array<ModalButton>;
 	/**
 	 * Callback for non-specific close events. `return false;` to prevent the modal from closing
+	 * Property trigger has a value of `overlay` or `close`
 	 */
-	close?: Function;
+	close?: (trigger: "overlay" | "close") => boolean;
```
cal-smith: Totally my bad for not catching this @HYACCCINT, and thanks @zvonimirfras for indirectly pointing it out ... this should be:
```suggestion
	close?: (trigger?: "overlay" | "close") => boolean;
```

so that consumers aren't required to accept the `trigger` argument

HYACCCINT: oops I definitely should have noticed this. sorry!

PR #1046, opened 2020-02-06
Add optional input to not display the close icon when header action is active

fionamurphy:
Closes IBM/carbon-components-angular#1044

Updates HeaderAction component to add an Input `showCloseIcon` to allow close icon not to be shown when in the active state - defaulted to true

### DIFF
--- a/src/ui-shell/header/header-action.component.ts
+++ b/src/ui-shell/header/header-action.component.ts
@@ -19,8 +19,8 @@ import {
 			[attr.aria-label]="title"
 			[title]="title"
 			(click)="onClick()">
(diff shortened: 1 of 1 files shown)


cal-smith:
Hey @fionamurphy! Thanks for digging into this - There's already another PR open that covers the same ground: #1031 and brings the shell more in line with the Carbon specs

Definitly check that out to make sure it meets your requirements!

I'm going to close this - we can continue the discussion over in the other PR 👍

fionamurphy:
Thanks @cal-smith Hadn't seen that one - that covers our use case

PR #1047, opened 2020-02-06
fix(file-uploader): add length check to prevent multiple input from being accepted when multiselect is disabled

HYACCCINT:
Closes IBM/carbon-components-angular#1045

{{adds a check for length and multi-select property in function onDrop}}

### DIFF
--- a/src/file-uploader/file-uploader.component.ts
+++ b/src/file-uploader/file-uploader.component.ts
@@ -11,7 +11,7 @@ import { NG_VALUE_ACCESSOR } from "@angular/forms";
 import { I18n } from "../i18n/i18n.module";
 import { FileItem } from "./file-item.interface";
 
-const noop = () => {};
+const noop = () => { };
 
 /**
  * [See demo](../../?path=/story/file-uploader--basic)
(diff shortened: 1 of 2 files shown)


src/file-uploader/file-uploader.component.ts
```diff
@@ -231,6 +231,9 @@ export class FileUploader {
 
 		if (!this.accept.length) {
 			transferredFiles.forEach(file => {
+				if ((this.files.size >= 1) && !this.multiple) {
```
cal-smith: Perhaps we could set up the `transferredFiles` correctly for this case, instead of bailing out in the `forEach` ... That way we only have on conditional to track, instead of two 🤔

HYACCCINT: good point!

src/file-uploader/file-uploader.component.ts
```diff
@@ -229,31 +229,29 @@ export class FileUploader {
...
+				this.files.add(fileItem);
+
+				this.filesChange.emit(this.files);
+			}
 		});
 
 		this.value = this.files;
+
 		this.dragOver = false;
+
```
zvonimirfras:
```suggestion
```

src/file-uploader/file-uploader.component.ts
```diff
@@ -229,31 +229,29 @@ export class FileUploader {
...
+				const fileItem = this.createFileItem(file);
+
+				this.files.add(fileItem);
+
+				this.filesChange.emit(this.files);
+			}
 		});
 
 		this.value = this.files;
+
```
zvonimirfras:
```suggestion
```

src/file-uploader/file-uploader.component.ts
```diff
@@ -229,31 +229,29 @@ export class FileUploader {
...
 			// Check if the accept array contains the mime type or extension of the file.
 			return this.accept.includes(type) || this.accept.includes(fileExtension);
 		}).forEach(file => {
-			const fileItem = this.createFileItem(file);
-			this.files.add(fileItem);
-			this.filesChange.emit(this.files);
+
+			if ((this.files.size < 1) || this.multiple) {
+				const fileItem = this.createFileItem(file);
+
```
zvonimirfras: There's no need to space everything out like this. All three lines essentially work together to accomplish a joint thing, they are contextually related, which is why they were grouped in the first place.
Readability doesn't always mean adding more space :)
```suggestion
```

src/file-uploader/file-uploader.component.ts
```diff
@@ -229,31 +229,29 @@ export class FileUploader {
...
+			if (this.accept.length === 0) {
+				return true;
+			}
 			// Check if the accept array contains the mime type or extension of the file.
 			return this.accept.includes(type) || this.accept.includes(fileExtension);
 		}).forEach(file => {
-			const fileItem = this.createFileItem(file);
-			this.files.add(fileItem);
-			this.filesChange.emit(this.files);
+
```
zvonimirfras:
```suggestion
```

src/file-uploader/file-uploader.component.ts
```diff
@@ -229,31 +229,29 @@ export class FileUploader {
...
+			if ((this.files.size < 1) || this.multiple) {
+				const fileItem = this.createFileItem(file);
+
+				this.files.add(fileItem);
+
```
zvonimirfras:
```suggestion
```

src/file-uploader/file-uploader.component.ts
```diff
@@ -233,11 +233,13 @@ export class FileUploader {
 			// Get the file extension and add a "." to the beginning.
 			const fileExtension = name.split(".").pop().replace(/^/, ".");
 			// Check if the accept array contains the mime type or extension of the file.
-			return this.accept.includes(type) || this.accept.includes(fileExtension) || this.accept.length === 0;
+			return this.accept.includes(type) || this.accept.includes(fileExtension) || !this.accept.length;
 		}).forEach(file => {
-			const fileItem = this.createFileItem(file);
-			this.files.add(fileItem);
-			this.filesChange.emit(this.files);
+			if ((this.files.size < 1) || this.multiple) {
```
Donisius:
```suggestion
			if (!this.files.size || this.multiple) {
```

PR #1048, opened 2020-02-06
fix(dialog): unsubscribe to local close subscriptions in OnDestroy

### LINKED ISSUE #1023: Tooltip - Not hiding on hover out.

### LINKED ISSUE #1057: [Progress indicator] Highlighting the tabs
## What packages are you using?

- [x] `carbon-components: "^10.9.3"`
- [x] `carbon-components-angular: "^3.39.1"`

## Detailed description
If I select a tab in the 'second step' to get additional information it will get highlighted. But if I now select the next one, still the first one stays highlighted. Please highlight only the tab for which the popup is displayed.

progrInd

I checked [react design carbon system](https://react.carbondesignsystem.com/?path=/story/progressindicator--default) and [carbon design system](https://www.carbondesignsystem.com/components/progress-indicator/code) and not saw same problem. So, I think this problem specific only to the [angular design system](https://angular.carbondesignsystem.com/?path=/story/components-progress-indicator--basic).

Donisius:
Closes #1023
Closes #1057

### DIFF
--- a/src/dialog/dialog.directive.ts
+++ b/src/dialog/dialog.directive.ts
@@ -43,7 +43,8 @@ export class DialogDirective implements OnInit, OnDestroy, OnChanges {
 	 */
 	@Input() ibmDialog: string | TemplateRef<any>;
 	/**
-	 * Defines how the Dialog is triggered.(Hover and click behave the same on mobile - both respond to a single tap)
+	 * Defines how the Dialog is triggered.(Hover and click behave the same on mobile - both respond to a single tap).
+	 * Do not add focusable elements if trigger is `hover` or `mouseenter`.
 	 */
 	@Input() trigger: "click" | "hover" | "mouseenter" = "click";
 	/**
--- a/src/dialog/dialog.directive.ts
+++ b/src/dialog/dialog.directive.ts
@@ -167,7 +167,7 @@ export class DialogDirective implements OnInit, OnDestroy, OnChanges {
 
 		// bind events for hovering or clicking the host
 		if (this.trigger === "hover" || this.trigger === "mouseenter") {
-			fromEvent(this.elementRef.nativeElement, "mouseenter").subscribe(() => this.toggle());
+			fromEvent(this.elementRef.nativeElement, "mouseenter").subscribe(() => this.open());
 			fromEvent(this.elementRef.nativeElement, this.closeTrigger).subscribe(() => this.close());
 			fromEvent(this.elementRef.nativeElement, "focus").subscribe(() => this.open());
 			fromEvent(this.elementRef.nativeElement, "blur").subscribe(() => this.close());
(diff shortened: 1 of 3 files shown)


src/dialog/dialog.service.ts
```diff
@@ -146,17 +148,18 @@ export class DialogService {
...
 		});
 
-		// keep track of all dialog subscriptions
-		DialogService.dialogCloseSubscription.add(closeSubscription);
+		// Adds current close subscription to the reference of all close subscriptions for
+		// local dialog service.
+		this.dialogSubscription.add(closeSubscription);
+
+		// keep track of all dialog subscriptions globally.
+		DialogService.dialogCloseSubscription.add(this.dialogSubscription);
```
cal-smith: One note, this should be called in the service constructor, since we don't want to accidentally add the subscription again (I don't think it will, but that's an implementation detail of rxjs we shouldn't rely on)

src/dialog/dialog.service.ts
```diff
@@ -77,7 +81,14 @@ export class DialogService {
 		protected componentFactoryResolver: ComponentFactoryResolver,
 		protected injector: Injector,
 		protected placeholderService: PlaceholderService
-	) {}
+	) {
+		// Adds current close subscription to the reference of all close subscriptions for
+		// local dialog service.
+		this.dialogSubscription.add(this.closeSubscription);
```
cal-smith: Sorry, my bad for not being clearer, `DialogService.dialogCloseSubscription.add` should absolutely be in the constructor, but `this.dialogSubscription.add` still needs to be in `open` 👍 otherwise it won't actually track it

src/dialog/dialog.directive.ts
```diff
@@ -222,10 +223,12 @@ export class DialogDirective implements OnInit, OnDestroy, OnChanges {
...
 	toggle() {
-		this.dialogService.toggle(this.viewContainerRef, this.dialogConfig);
-		this.expanded = this.dialogService.isOpen;
+		this.expanded = !this.expanded;
 		if (this.expanded) {
 			this.onOpen.emit();
+			this.open();
+		} else {
+			this.close();
 		}
```
theiliad: can we do?

```js
if (this.expanded) {
		this.onOpen.emit();
		return this.open();
}

this.close();
```

theiliad: This is a rare case where I think doing the `return` might not be so readable

src/dialog/dialog.service.ts
```diff
@@ -185,13 +188,24 @@ export class DialogService {
...
 
-		if (this.dialogSubscription) {
-			this.dialogSubscription.unsubscribe();
+		// Keeps the focus on the dialog trigger if there are no focusable elements. Change focus to previously focused element
+		// if there are focusable elements in the dialog or if trigger is set to `hover` or `mouseenter`.
+		if (
+			!dialogRef.location.nativeElement.querySelectorAll(tabbableSelector) ||
+			dialogRef.instance.dialogConfig.trigger === "hover" ||
+			dialogRef.instance.dialogConfig.trigger === "mouseenter") {
+			elementToFocus.focus();
```
theiliad: can we ENUM `mouseenter`?

Donisius: Why just `mouseenter`?

theiliad: well everything

PR #1056, opened 2020-02-11
fix(AnimationFrameService): use runOutsideAngular to prevent RAF triggering change detection

cal-smith:
Closes IBM/carbon-components-angular#1029

### DIFF
--- a/src/utils/animation-frame.service.ts
+++ b/src/utils/animation-frame.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, OnDestroy } from "@angular/core";
+import { Injectable, OnDestroy, NgZone } from "@angular/core";
 import { Observable, Subject, from } from "rxjs";
 
 @Injectable()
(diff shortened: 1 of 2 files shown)


src/utils/animation-frame.service.ts
```diff
@@ -9,9 +9,11 @@ export class AnimationFrameServiceSingleton implements OnDestroy {
 
 	protected animationFrameId: number;
 
-	constructor() {
+	constructor(protected ngZone: NgZone) {
```
Donisius: Maybe this needs to be optional? It's expecting an `ngZone` argument in `utils.module`

cal-smith: This is what I get for not running my changes locally 😂

src/utils/utils.module.ts
```diff
@@ -1,18 +1,18 @@
-import { NgModule, Optional, SkipSelf } from "@angular/core";
+import { NgModule, Optional, SkipSelf, NgZone } from "@angular/core";
```
zvonimirfras:
```suggestion
import {
	NgModule,
	Optional,
	SkipSelf,
	NgZone
} from "@angular/core";
```

PR #1061, opened 2020-02-18
fix(dropdown): a11y interactions

cal-smith:
This fixes the broken the a11y interactions.


src/dropdown/dropdown.component.ts
```diff
@@ -57,16 +57,16 @@ import { ElementService } from "./../utils/utils.module";
...
 			#dropdownButton
 			class="bx--list-box__field"
 			[ngClass]="{'a': !menuIsClosed}"
 			[attr.aria-expanded]="!menuIsClosed"
 			[attr.aria-disabled]="disabled"
-			(click)="disabled ? $event.stopPropagation() : toggleMenu()"
+			(click)="toggleMenu()"
 			(blur)="onBlur()"
-			[attr.disabled]="disabled ? true : null">
+			[disabled]="disabled ? true : null">
```
theiliad: A bit weird this line but I guess there aren't many things you can do when dealing with the `disabled` attr

PR #1062, opened 2020-02-18
feat(table): add onCancelSelected event

### LINKED ISSUE #1033: Need an event emitter for on cancel in toolbar actions
## Detailed description

we have another array used to track more detailed info of our orders, however, we cannot empty it when we click cancel button in the toolbar section, please see this line. we need a oncancel event:

https://github.com/IBM/carbon-components-angular/blob/master/src/table/toolbar/table-toolbar.component.ts#L94

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

* Screenshots or code
* Notes

### LINKED ISSUE #1065: Need event emitter for table toolbar Cancel action
We are using ibm-table-toolbar component. We have cancel action out of the box provided by this component which deselects all the rows. We would need one event emitter for this cancel action? As, at our end, we want to do something when user clicks on cancel button. I checked, deselectAll event also doesn't get emitted on click of this cancel button as table model is directly being updated in this action => https://angular.carbondesignsystem.com/documentation/components/TableToolbar.html#source.

https://ibm-cognitive-engage.slack.com/archives/CBZC0LM2N/p1582029510308300 for reference.

Thanks in advance!

Donisius:
Closes #1033 
closes #1065

Add `onCancelSelected` event which gets emitted when tool bar cancel button is clicked. Also remove duplicate story from storybook.

### DIFF
--- a/src/table/table-model.class.ts
+++ b/src/table/table-model.class.ts
@@ -52,6 +52,11 @@ export class TableModel implements PaginationModel {
 	dataChange = new EventEmitter();
 	rowsSelectedChange = new EventEmitter<number>();
 	rowsExpandedChange = new EventEmitter<number>();
+	/**
+	 * Gets emitted when `selectAll` is called. Emits false if all rows are deselected and true if
+	 * all rows are selected.
+	 */
+	selectAllChange = new EventEmitter<boolean>();
 
 	/**
 	 * Gets the full data.
@@ -588,6 +593,7 @@ export class TableModel implements PaginationModel {
 	 * @param value state to set all rows to. Defaults to `true`
 	 */
 	selectAll(value = true) {
+		this.selectAllChange.emit(value);
 		if (this.data.length >= 1) {
 			for (let i = 0; i < this.rowsSelected.length; i++) {
 				this.selectRow(i, value);
--- a/src/table/table-model.class.ts
+++ b/src/table/table-model.class.ts
@@ -5,6 +5,7 @@ import {
 import { PaginationModel } from "./../pagination/pagination-model.class";
 import { TableHeaderItem } from "./table-header-item.class";
 import { TableItem } from "./table-item.class";
+import { Subject } from "rxjs";
 
 export type HeaderType = number | "select" | "expand";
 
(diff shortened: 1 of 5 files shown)


src/table/toolbar/table-toolbar.component.ts
```diff
@@ -92,6 +101,7 @@ export class TableToolbar {
 	}
 
 	onCancel() {
+		this.onCancelSelected.emit();
 		for (let i = 0; i < this.model.rowsSelected.length; i++) {
 			this.model.selectRow(i, false);
```
Donisius: @Cal Do you think it would break anything if the whole loop is changed into `this.model.selectAll(false)`?

cal-smith: Nope, that's definitely how we should be handling it now

src/table/table-model.class.ts
```diff
@@ -52,6 +52,11 @@ export class TableModel implements PaginationModel {
 	dataChange = new EventEmitter();
 	rowsSelectedChange = new EventEmitter<number>();
 	rowsExpandedChange = new EventEmitter<number>();
+	/**
+	 * Gets emitted when `selectAll` is called. Emits false if all rows are deselected and true if
+	 * all rows are selected.
+	 */
+	selectAllChange = new EventEmitter<boolean>();
```
cal-smith: Should be `Subject<boolean>()` - `EventEmitter` isn't intended to be used outside of component classes

src/table/table-model.class.ts
```diff
@@ -588,6 +593,7 @@ export class TableModel implements PaginationModel {
 	 * @param value state to set all rows to. Defaults to `true`
 	 */
 	selectAll(value = true) {
+		this.selectAllChange.emit(value);
```
cal-smith: We may want to emit this after we've applied the changes to the model 🤔

src/table/toolbar/table-toolbar.component.ts
```diff
@@ -74,6 +79,10 @@ export class TableToolbar {
 	@Input() set cancelText(value: { CANCEL: string }) {
 		this._cancelText.override(value.CANCEL);
 	}
+	/**
+	 * Will be emitted when cancel button is clicked and all rows are deselected.
+	 */
+	@Output() onCancelSelected = new EventEmitter();
```
cal-smith: we probably want to remove this if we're running everything through the model ... this doesn't seem to be hooked up to anything anyway

PR #1068, opened 2020-02-19
fix: make compare function more robust

virkt25:
in a striped table in dark theme there is an empty row added at the start of the table. That row shows as `undefined` in compare when a column is sorted. This ensures the empty row stays at the top. Also using localeCompare to compare a string


virkt25:
@cal-smith Lint issue fixed. PR rebased.

PR #1090, opened 2020-02-28
fix(combobox): update view with the correct selected item

### LINKED ISSUE #1087: Combo-box Clear button defect
## Detailed description
Combo-box clear does not let you add same selection, once you have cleared it.

1. Select any element in combo-box
image
image

2. Clear selection
image

3. Select the same option again
image

4. See that you can see the cross icon but no element can be seen
image

## Additional information

* Screenshots or code
* Notes

Donisius:
Closes #1087


src/combobox/combobox.component.ts
```diff
@@ -469,9 +475,14 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit {
...
 
-		this.input.nativeElement.value = "";
-
 		this.clearSelected();
+
+		// This makes sure there are no selected items before clearing the input manually.
+		// The view needs to be updated manually because the input will not clear automatically
+		// if we clear the selected item for a single selection combobox.
+		const selected = this.view.getSelected();
+		this.input.nativeElement.value = selected ? selected : "";
```
cal-smith: This seems redundant since two lines above we call `clearSelected` which ... clears the selection

src/combobox/combobox.component.ts
```diff
@@ -285,6 +285,12 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit {
 						this.showClearButton = true;
 						this.selectedValue = event.item.content;
 						this.propagateChangeCallback(event.item);
+						// Under some circumstances the view does not get updated with the selected
+						// item for a single selection combobox. This manually updates the input with the
+						// correct selected item.
+						if (this.view.getSelected()[0]) {
+							this.input.nativeElement.value = this.view.getSelected()[0].content;
+						}
```
cal-smith: Which circumstances? Why are we bypassing `selectedValue` and the angular value update in just this one spot?

src/combobox/combobox.component.ts
Donisius: It clears the selection and the content does update correctly but the view does not update, it keeps the same value in the input (even if the model does not reflect this).

src/combobox/combobox.component.ts
Donisius: I think it's better if I show you 😅

cal-smith: Not really ... we shouldn't be updating the state of the input in two places ... either this needs to be run on the next change detection, or all the updates should go through the native element in a setter

src/combobox/combobox.component.ts
cal-smith: But that's not what this code is doing ... this code is checking if there's a selection and then not clearing the input value

cal-smith: There should be no way that there is a selection at this point (barring disabled items which may not even be part of the `getSelected` result set)

PR #1091, opened 2020-02-28
Add/improve docs for boolean Inputs

cal-smith:
Works on IBM/carbon-components-angular#673

WIP since there's a ton of area to cover


cal-smith:
this probably needs a fresh start

PR #1111, opened 2020-03-09
Pass data to notification/toast template

jdcai:
Pass data to notification/toast template

Closes IBM/carbon-components-angular#1110

#### Changelog

**New**

Added templateData for NotificationContent interface to pass it to template

**Changed**

Added context: notificationObj.templateData to pass data to the template in the components
Changed storybook to show the data being passed


jdcai:
Hi @Donisius can you review this since you've worked on this before?

src/notification/notification.stories.ts
```diff
@@ -236,19 +236,21 @@ storiesOf("Components|Notification", module)
 			<ibm-toast [notificationObj]="{
 				type: 'error',
 				template: customToastContent,
+				templateData: {data: {title: 'Sample custom toast', subtitle: 'Sample custom subtitle', caption: 'Sample custom caption' }},
```
Donisius:
```suggestion
				templateData: {
					data: {
						title: 'Sample custom toast', subtitle: 'Sample custom subtitle', caption: 'Sample custom caption'
					}
				},
```

src/notification/notification.stories.ts
```diff
@@ -236,19 +236,21 @@ storiesOf("Components|Notification", module)
...
 				showClose: showClose
 			}">
 			</ibm-toast>
 			<ibm-notification [notificationObj]="{
 				type: 'warning',
 				template: customNotificationContent,
+				templateData: {data: {title: 'Sample custom notification', subtitle: 'Sample custom subtitle' }},
```
Donisius:
```suggestion
				templateData: {
					data: {
						title: 'Sample custom notification', subtitle: 'Sample custom subtitle'
					}
				},
```

src/notification/notification.stories.ts
jdcai: Added commit for it

src/notification/notification-content.interface.ts
```diff
@@ -12,6 +12,7 @@ export interface NotificationContent {
 	showClose?: boolean;
 	lowContrast?: boolean;
 	template?: TemplateRef<any>;
+	templateData?: any;
```
cal-smith: It might be more useful if we add an index type (https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types-and-index-signatures - `[key: any]: any`) and just passed the whole `NotificationContent` through to the template ... that way for simple cases like adjusting titles or captions we don't need to duplicate the values

jdcai: added a commit, is it what you were thinking of?

PR #1112, opened 2020-03-10
fix(table): create a new subscription instance when a new table model is set

### LINKED ISSUE #1103: Data table: Checkbox for all broken when used with skeleton
## Detailed description

BUG REPORT:
When changing model of the table, main checkbox misbehaves. On (selectAll) and (deselectAll) always triggers (selectAll).
Code example: https://stackblitz.com/edit/angular-7wrzxr 

Bug

Data table

Expected function (deselectAll) to unmark all checkboxes

Chrome & Firefox

"carbon-components": "^10.9.3",
"carbon-components-angular": "^3.40.3",

Example code here: https://stackblitz.com/edit/angular-7wrzxr

1. try to use checkbox for all

## Additional information

skeleton_checkbox_bug

Donisius:
Closes #1103

### DIFF
--- a/src/table/table.component.ts
+++ b/src/table/table.component.ts
@@ -305,6 +305,9 @@ export class Table implements AfterViewInit, OnDestroy {
 	set model(m: TableModel) {
 		if (this._model) {
 			this.subscriptions.unsubscribe();
+			// Need to create a new subscription instance here because unsubscribing prevents any new subscriptions
+			// from being added for some reason. When a new model is set, none of the subscriptions would exist.
+			this.subscriptions = new Subscription();
 		}
 
 		this._model = m;


src/table/table.component.ts
```diff
@@ -305,6 +305,9 @@ export class Table implements AfterViewInit, OnDestroy {
 	set model(m: TableModel) {
 		if (this._model) {
 			this.subscriptions.unsubscribe();
+			// Need to create a new subscription instance here because unsubscribing prevents any new subscriptions
+			// from being added for some reason. When a new model is set, none of the subscriptions would exist.
+			this.subscriptions = new Subscription();
```
cal-smith: Interesting 🤔 this also has some potential implications for other places we use this pattern if true

cal-smith: May be worth double checking a few other components and opening an issue/pr to sort those out too

PR #1120, opened 2020-03-12
fix(dropdown): make clear button accessible via keyboard

### LINKED ISSUE #1102: Dropdown component - clear filter button is not accessible via keyboard

Our AVT team flagged that the clear filter button cannot be accessed by keyboard, pressing Tab doesn't take the control over to the button.

Bug

Dropdown 

Chrome

V3.40.4

IBM ADDI

Donisius:
Closes #1102

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -58,8 +58,10 @@ import { DocumentService } from "../utils/utils.module";
 				<div
 					*ngIf="type === 'multi' && pills.length > 0"
 					(click)="clearSelected()"
+					(keydown.enter)="clearSelected()"
 					role="button"
-					class="bx--list-box__selection bx--list-box__selection--multi"
+					class="bx--list-box__selection bx--tag--filter bx--list-box__selection--multi"
+					tabindex="0"
 					title="Clear all selected items">
 					{{ pills.length }}
 					<svg
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -60,7 +60,7 @@ import { DocumentService } from "../utils/utils.module";
 					(click)="clearSelected()"
 					(keydown.enter)="clearSelected()"
 					role="button"
-					class="bx--list-box__selection bx--tag--filter bx--list-box__selection--multi"
+					class="bx--tag--filter bx--list-box__selection--multi"
 					tabindex="0"
 					title="Clear all selected items">
 					{{ pills.length }}
--- a/src/dropdown/dropdown.component.ts
+++ b/src/dropdown/dropdown.component.ts
@@ -25,6 +25,8 @@ import {
 
 import { AbstractDropdownView } from "./abstract-dropdown-view.class";
 import { I18n } from "./../i18n/i18n.module";
+// Import needed to avoid compiler issue.
+import { ListItem } from "./list-item.interface";
 import { DropdownService } from "./dropdown.service";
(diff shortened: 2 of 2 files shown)


src/combobox/combobox.component.ts
```diff
@@ -58,8 +58,10 @@ import { DocumentService } from "../utils/utils.module";
 				<div
 					*ngIf="type === 'multi' && pills.length > 0"
 					(click)="clearSelected()"
+					(keydown.enter)="clearSelected()"
```
theiliad: nice! didn't know angular can do that

Donisius: I didn't either :D

cal-smith: Such a handy feature for things like this! There's a few other modifiers you can stack IIRC

src/combobox/combobox.component.ts
```diff
@@ -58,8 +58,10 @@ import { DocumentService } from "../utils/utils.module";
...
 					role="button"
-					class="bx--list-box__selection bx--list-box__selection--multi"
+					class="bx--list-box__selection bx--tag--filter bx--list-box__selection--multi"
```
cal-smith: Do we need `bx--tag--filter` as well? I know it was working without it before

src/dropdown/dropdown.component.ts
```diff
@@ -25,7 +25,6 @@ import {
 
 import { AbstractDropdownView } from "./abstract-dropdown-view.class";
 import { I18n } from "./../i18n/i18n.module";
-import { ListItem } from "./list-item.interface";
```
cal-smith: This may be needed to solve a compiler issue ... I'd leave it just in case 😬 maybe add a comment though

src/dropdown/dropdown.component.ts
```diff
@@ -70,12 +69,13 @@ import { ElementService } from "./../utils/utils.module";
 			[tabindex]="disabled ? -1 : 0">
 			<div
 				(click)="clearSelected()"
+				(keydown.enter)="clearSelected()"
 				*ngIf="type === 'multi' && getSelectedCount() > 0"
-				class="bx--list-box__selection--multi"
+				class="bx--list-box__selection bx--tag--filter bx--list-box__selection--multi"
```
cal-smith: Same note here about `tag--filter` ... also do we want `selection` and `selection--multi` at the same time? Or is that something we should `ngClass`?

src/dropdown/dropdown.component.ts
```diff
@@ -70,12 +69,13 @@ import { ElementService } from "./../utils/utils.module";
...
+				tabindex="0"
 				[title]="clearText">
 				{{getSelectedCount()}}
 				<svg
-					focusable="false"
```
cal-smith: SVGs shouldn't be focusable 👍

src/combobox/combobox.component.ts
Donisius: It doesn't allow the close button to be focused without it

src/dropdown/dropdown.component.ts
Donisius: I'm not sure if we need both since this clear selected button only pops up for multi-select. I was following the structure in react's implementation and they had both, however it seems to work without the `selection` class

src/combobox/combobox.component.ts
cal-smith: Huh, odd

PR #1121, opened 2020-03-12
fix(dialog): account for partial visibility in visibility calculation

### LINKED ISSUE #1101: Overflow menu not opening if menu button is below the fold of scroll parent
## Detailed description

See description, we hit the issue in our Carbon table but can be reproduced in the Storybook example

Bug

Overflow menu

Chrome, FF

Seen in 3.39.6 and latest on Storybook (should be 3.43.0)

IBM Watson Health - Study Advance

1. Go to the overflow menu [Storybook example](https://angular.carbondesignsystem.com/?path=/story/components-overflow-menu--basic).
2. Shrink the application height of the application so that the "Top Placement" overflow menu is below the fold, like this:
image
3. Attempt to open the "Top Placement" menu and the menu will not open, and the scroll position may jump to the top.

## Additional information

* Happens with both top and bottom placement options
* I couldn't reproduce with this Code Sandbox example (might not have the right combination of styles and parent elements): https://codesandbox.io/s/carbon-components-angular-gjsxc but I see it in our app and in the storybook.

Donisius:
Closes #1101

### DIFF
--- a/src/utils/scroll.ts
+++ b/src/utils/scroll.ts
@@ -27,7 +27,11 @@ export const isScrollableElement = (element: HTMLElement) => {
 export const isVisibleInContainer = (element: HTMLElement, container: HTMLElement) => {
 	const elementRect = element.getBoundingClientRect();
 	const containerRect = container.getBoundingClientRect();
-	return elementRect.bottom <= containerRect.bottom && elementRect.top >= containerRect.top;
+	return (
+		// This also accounts for partial visibility. It will still return true if the element is partially visible inside the container.
+		(elementRect.bottom - element.clientHeight) <= (containerRect.bottom + (container.offsetHeight - container.clientHeight) / 2) &&
+		elementRect.top >= (- element.clientHeight)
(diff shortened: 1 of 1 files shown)


Donisius:
> LGTM 👍

From slack: just wanted to add that the fix doesn’t seem to work when accounting for the window scroll position

cal-smith:
That's odd ... is it any scrolling container, or just the window? If it's just the window you can probably add the x/y scroll position in your visibility calculation to sort that out

Donisius:
> That's odd ... is it any scrolling container, or just the window? If it's just the window you can probably add the x/y scroll position in your visibility calculation to sort that out

Yeah I'll add that as well, maybe the component is in a container that doesn't fully wrap it

cal-smith:
> Yeah I'll add that as well, maybe the component is in a container that doesn't fully wrap it

Definitly double check, should be pretty easy to wrap the component in a `overflow: scroll` div to test

PR #1141, opened 2020-03-20
The v4 update PR part two: deprecations and breaking things

cal-smith:
Each major piece of work is a commit, so it should be easy to follow

TODO: details


src/modal/modal.stories.ts
```diff
@@ -164,13 +165,13 @@ class AlertModalStory {
 	@Input() modalTitle: string;
 	@Input() modalContent: string;
 	@Input() buttons: Array<ModalButton>;
-	@Input() size: "xs" | "sm" | "default" | "lg";
```
Donisius: I don't think this should be taken out, because it does show the medium sized modal that's different than the other sizes

src/modal/modal.stories.ts
```diff
@@ -210,7 +212,7 @@ storiesOf("Components|Modal", module)
 		`,
 		props: {
 			modalText: text("modalText", "Hello, World!"),
-			size: select("size", ["xs", "sm", "default", "lg"], "default")
```
Donisius: Same with this

src/modal/modal.stories.ts
cal-smith: Huh that doesn't seem to be removed on my end?

cal-smith: If there's a specific `"md"` size modal we can add that in

Donisius: I think it just defaults to medium if you provide it with anything other than `xs, sm`, or `lg`. There's no medium class that can be applied

src/checkbox/checkbox.component.ts
```diff
@@ -126,26 +134,47 @@ export class Checkbox implements ControlValueAccessor, AfterViewInit {
...
+		this._ariaLabelledby = value;
+	}
+
+	get ariaLabelledby() {
+		return this._ariaLabelledby;
 	}
 
+	// TODO: drop the `_`
+	// tslint:disable-next-line:no-input-rename
+	@Input("ariaLabelledby") _ariaLabelledby: string;
```
Donisius: Does this need to be an input, or can it be a private/non-input property?

src/dialog/tooltip/tooltip.stories.ts
```diff
@@ -3,7 +3,9 @@ import { action } from "@storybook/addon-actions";
 import { withKnobs, text, select } from "@storybook/addon-knobs/angular";
 
 import { InformationFilledModule } from "@carbon/icons-angular";
-import { DialogModule, PlaceholderModule, DocumentationModule } from "../../";
+import { DialogModule, PlaceholderModule } from "../../";
+import { DocumentationModule } from "../../documentation-component/documentation.module";
```
Donisius: Some of the other stories are also failing because of the the wrong import location for this module

src/modal/modal.stories.ts
cal-smith: Okay, so we _should_ be good then ... We may need to update the story to send a `null` by default though

src/table/table.component.ts
```diff
@@ -30,6 +30,8 @@ export interface TableTranslations {
 	CHECKBOX_ROW: string;
 }
 
+export type TableRowSize = "sm" | "sh" | "md" | "lg";
```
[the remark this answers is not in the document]
cal-smith: Enums are tricky since to use them properly you have to first assign them to a class property ... Hence why I've declared types like that (plus, if they do become viable, it's not a breaking change to swap them over)

src/checkbox/checkbox.component.ts
cal-smith: I'm not sure what you mean ... Why would we want the label to not be an input?

cal-smith: The setter above is an input remapped to `aria-label` ... this change maps this input to `ariaLabel`, but because we already have an `ariaLabel` property (the setter) we need to rename it to `_ariaLabel` ... We're deprecating `aria-label` and going with `ariaLabel`

Donisius: Oh okay I understand, thanks for clarifying

src/dialog/dialog.service.ts
```diff
@@ -142,12 +98,10 @@ export class DialogService implements OnDestroy {
 		// initialize some extra options
 		dialogConfig["previouslyFocusedElement"] = document.activeElement;
 		dialogRef.instance.dialogConfig = dialogConfig;
-		this.onClose = dialogRef.instance.close;
-		this.isOpen = true;
 
-		const closeSubscription = this.onClose.subscribe(() => {
+		const closeSubscription = dialogRef.instance.close.subscribe(() => {
```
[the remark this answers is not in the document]
cal-smith: It was mostly working for me ... turns out I forgot a call to `afterDialogViewInit` and could simplify the re-focus logic ... seems to be working correctly now 👍

src/modal/modal.component.ts
```diff
@@ -106,8 +109,28 @@ export class Modal implements AfterViewInit {
...
+	get modalLabel() {
+		return this.ariaLabel;
+	}
+
+	@Input() ariaLabel = "default";
+
+	/**
+	 * Controls the visibility of the modal when used directly in a template
+	 */
+	@Input() open = false;
```
Donisius: The `In modal` story for `datepicker` needs to have the open property be set to true

src/accordion/accordion.stories.ts
```diff
@@ -2,7 +2,8 @@ import { storiesOf, moduleMetadata } from "@storybook/angular";
 import { action } from "@storybook/addon-actions";
 import { withKnobs, boolean, object } from "@storybook/addon-knobs/angular";
 
-import { AccordionModule, DocumentationModule } from "../";
+import { AccordionModule } from "../";
+import { DocumentationModule } from "../documentation-component/documentation.module";
```
Donisius: Documentation module import in file-uploader and table also needs to be updated

src/dialog/overflow-menu/overflow-menu.component.ts
```diff
@@ -77,9 +75,16 @@ export class OverflowMenu {
...
+	@Input() open = false;
+
+	@Output() openChange = new EventEmitter<boolean>();
 
-	open = false;
+	@ContentChild(OverflowMenuDirective) overflowMenuDirective: OverflowMenuDirective;
 
 	constructor(protected elementRef: ElementRef, protected i18n: I18n) {}
+
+	handleOpenChange(event: boolean) {
```
Donisius: Not sure where it's coming from but when I open an overflow menu I get the following error:

```
TypeError: Cannot set property 'tabIndex' of undefined
    at OverflowMenuPane.push../src/dialog/overflow-menu/overflow-menu-pane.component.ts.OverflowMenuPane.afterDialogViewInit (overflow-menu-pane.component.ts:128)
    at dialog.component.ts:154
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:421)
    at Object.onInvokeTask (core.js:17290)
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:420)
    at Zone.push../node_modules/zone.js/dist/zone.js.Zone.runTask (zone.js:188)
    at push../node_modules/zone.js/dist/zone.js.ZoneTask.invokeTask (zone.js:496)
    at ZoneTask.invoke (zone.js:485)
    at timer (zone.js:2054)
```

And opening a partially visible overflow menu doesn't work, maybe that change wasn't pulled yet?

src/modal/modal.component.ts
cal-smith: Not specifically to true, but I see what you mean 👍

src/toggle/toggle.component.ts
```diff
@@ -167,12 +169,15 @@ export class Toggle extends Checkbox {
...
 	 */
 	emitChangeEvent() {
+		/* begin deprecation */
 		let event = new ToggleChange();
 		event.source = this;
 		event.checked = this.checked;
+		this.change.emit(event);
+		/* end deprecation */
 
+		this.checkedChange.emit(this.checked);
```
zvonimirfras: Same emit/propagate order question

src/dialog/overflow-menu/overflow-menu.component.ts
cal-smith: This lead me down a whole rabbit hole of refactoring the dialog service ... but, everything works like 1000% better now, so that's good 🎉

src/modal/modal.component.ts
```diff
@@ -132,20 +155,24 @@ export class Modal implements AfterViewInit {
...
+		if (open && open.currentValue) {
+			// `100` is just enough time to allow the modal
+			// to become visible, so that we can set focus
+			setTimeout(() => this.focusInitialElement(), 100);
+		} else if (open && !open.currentValue) {
+			if (this.trigger) {
+				this.trigger.focus();
+			}
+		}
+	}
```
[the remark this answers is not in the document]
cal-smith: I don't totally want to early return in this case, if we end up needing more handlers in `ngOnChanges`

src/dialog/dialog.service.ts
```diff
@@ -85,27 +56,12 @@ export class DialogService implements OnDestroy {
...
-	 *
-	 * @deprecated components should implement their own `toggle` with the `close` and `open` methods
-	 */
-	toggle(viewContainer: ViewContainerRef, dialogConfig: DialogConfig) {
-		if (this.isOpen) {
-			this.close(viewContainer);
-		} else {
-			this.open(viewContainer, dialogConfig);
-		}
+	context(options: { component: any }) {
```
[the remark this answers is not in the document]
cal-smith: Sure, I'm not totally set one way or the other

src/checkbox/checkbox.component.ts
```diff
@@ -300,12 +357,15 @@ export class Checkbox implements ControlValueAccessor, AfterViewInit {
...
 	 */
 	emitChangeEvent() {
+		/* begin deprecation */
 		let event = new CheckboxChange();
 		event.source = this;
 		event.checked = this.checked;
+		this.change.emit(event);
+		/* end deprecation */
 
+		this.checkedChange.emit(this.checked);
```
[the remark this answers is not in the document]
cal-smith: Nope, just re-grouping for readability, but I can swap it back if there's strong feelings

src/toggle/toggle.component.ts
cal-smith: Same answer 👍

src/checkbox/checkbox.component.ts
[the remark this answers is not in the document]
zvonimirfras: Not if it doesn't change the functionality.

src/dialog/dialog.service.ts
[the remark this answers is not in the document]
zvonimirfras: Cool, let's do that then :)

src/modal/modal.component.ts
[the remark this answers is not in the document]
zvonimirfras: Sure, in that case, how about
```suggestion
		if (open) {
			if (open.currentValue) {
				// `100` is just enough time to allow the modal
				// to become visible, so that we can set focus
				setTimeout(() => this.focusInitialElement(), 100);
			} else if (this.trigger) {
				this.trigger.focus();
			}
		}
	}
```

src/checkbox/checkbox.component.ts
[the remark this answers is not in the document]
cal-smith: No reason it would :+1:

src/modal/modal.component.ts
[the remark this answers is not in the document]
cal-smith: Yeah that looks like it should work 👍

src/dialog/dialog.service.ts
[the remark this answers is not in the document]
cal-smith: sounds good

PR #1145, opened 2020-03-23
 fix(table-container): remove position relative from container

### LINKED ISSUE #1144: Expansion table in a page with a scrollbar causes scrollbar to move to table on row expansion

### LINKED ISSUE #1148: Table component vertical scroll issue
Table component - Expandable table (https://angular.carbondesignsystem.com/?path=/story/components-table--with-expansion)

To replicate - Expand any table row. Now scroll up and down.

Issue - Once user tries to scroll through page, vertical scrollbar shows inside table because of which user is not able to scroll through page.

This issue got induced in carbon version - 3.45.1. As earlier, this wasn't there.

Kindly look into this. Thanks in advance!

Donisius:
Closes #1144 
Closes #1148

### DIFF
--- a/src/table/table-container.component.ts
+++ b/src/table/table-container.component.ts
@@ -4,9 +4,7 @@ import { Component, HostBinding } from "@angular/core";
 	selector: "ibm-table-container",
 	template: `<ng-content></ng-content>`,
 	styles: [`
-		:host {
-			display: block;
-		}
+		:host { display: block }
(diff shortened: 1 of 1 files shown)


Donisius:
Not sure if there are potential repercussions from removing the position: relative but it seems to have solved the issue and kept the container as a visible scrollable parent to the visibility calculation

PR #1147, opened 2020-03-23
fix(timepicker-select): add input wrapper to prevent icon from displaying incorrectly in IE11

### LINKED ISSUE #1143: Time picker select's arrow displays incorrectly on IE11
## Detailed description

Issue details can be found at https://github.com/carbon-design-system/carbon/issues/5651

There is a pull request https://github.com/carbon-design-system/carbon/pull/5672 that will fix the issue for vanilla and react since the class "bx--select_arrow" is on the SVG but for angular it will make it display incorrectly in all browsers since the class will be on the container instead of the SVG.
image

Bug

Time picker select

Arrow should display correctly on IE11. Maybe if the arrow icon uses carbon icons angular's SVG directive instead of the component. This way we can put the class `class="bx--select__arrow"` on the SVG and it should display correctly.

IE11

carbon-components: 10.10.2
carbon-components-angular: 3.46.6

Donisius:
Closes #1143 

Still not perfect, but it looks a lot better with the wrapper. It seems like a styling issue since the problem exists in the react implementation as well.

before:
Screen Shot 2020-03-23 at 4 32 56 PM

after:
Screen Shot 2020-03-23 at 4 54 18 PM

### DIFF
--- a/src/timepicker-select/timepicker-select.component.ts
+++ b/src/timepicker-select/timepicker-select.component.ts
@@ -17,7 +17,8 @@ import { NG_VALUE_ACCESSOR } from "@angular/forms";
 @Component({
 	selector: "ibm-timepicker-select",
 	template: `
-			<label *ngIf="!skeleton && label" [attr.for]="id" class="bx--label bx--visually-hidden">{{label}}</label>
+		<label *ngIf="!skeleton && label" [attr.for]="id" class="bx--label bx--visually-hidden">{{label}}</label>
+		<div class="bx--select-input__wrapper">
 			<select
 				#select
 				[attr.id]="id"
@@ -27,6 +28,7 @@ import { NG_VALUE_ACCESSOR } from "@angular/forms";
 				<ng-content></ng-content>
 			</select>
 			<ibm-icon-chevron-down16 *ngIf="!skeleton" class="bx--select__arrow"></ibm-icon-chevron-down16>
+		</div>
 	`,
 	providers: [
 		{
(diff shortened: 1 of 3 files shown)


jdcai:
The pull request https://github.com/carbon-design-system/carbon/pull/5672 hasn't been merged yet but when it is merged it will fix it for vanilla and react. In angular it will cause the arrow to display incorrectly. More details in issue.

src/timepicker-select/timepicker-select.component.ts
```diff
@@ -27,6 +28,7 @@ import { NG_VALUE_ACCESSOR } from "@angular/forms";
 				<ng-content></ng-content>
 			</select>
 			<ibm-icon-chevron-down16 *ngIf="!skeleton" class="bx--select__arrow"></ibm-icon-chevron-down16>
```
cal-smith: Does changing to the directive form `<svg ibmIconChevronDown16></svg>` improve things at all?

Donisius: It doesn't change anything

jdcai: Did you try applying the styles from https://github.com/carbon-design-system/carbon/pull/5672/files after changing to the directive?

Donisius: I haven't! I will try that out

cal-smith: We should switch to the directive form anyway 👍

PR #1149, opened 2020-03-25
docs(combobox): add storybook example with ngModel

Donisius:


Donisius:
> Overall looks good ... though this really isn't a `feat` so maybe amend your commit to be `docs():` or similar

true 😂

PR #1150, opened 2020-03-26
Add a very basic checkbox exported test

### LINKED ISSUE #1117: Create an exported checkbox test using the Watson Health exported-tests tool

Donisius:
Closes #1117


src/checkbox/checkbox.component.spec.ts
```diff
@@ -25,13 +27,32 @@ class CheckboxTest {
 	onIndeterminateChange() {}
 }
 
-describe("Checkbox", async() => {
+const defaults = {
```
elizabethsjudd, line 30: Typically we try to [provide defaults within the tests](https://github.ibm.com/Watson-Health/design-library/blob/carbon-x/packages/pattern-library/src/components/modal/tests/requirements_exp.js#L42) themselves so that developers using them only have to overwrite them if needed. In which most cases, developers will only have to overwrite the root as the rest of the elements are always searched for from the root element. (e.g: https://github.ibm.com/Watson-Health/design-library/blob/carbon-x/packages/pattern-library/src/components/modal/tests/modal.spec.js#L66-L68)

src/exported-tests/component-tests.ts
```diff
@@ -0,0 +1,39 @@
+import merge from "lodash-es/merge";
```
elizabethsjudd, line 1: We knew you weren't going to have this file via PAL at this stage. I'm curious if there was anything (other than formatting for TypeScript) that was needed in order to meet your needs.  I'm excited to get more of these files in a shared location for everyone to easily use.

src/checkbox/checkbox-exported-tests.ts
```diff
@@ -0,0 +1,45 @@
...
+						const initialState = component.checked;
+						component.click();
+						const afterClickState = component.checked;
+						resolve({
+							initialState,
+							afterClickState
+						});
+					}),
+				runComparison: results => {
+					/* tslint:disable-next-line */
```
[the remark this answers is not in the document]
Donisius: It seems to think the two expect statements are unused expressions, I come across the error: `unused expression, expected an assignment or function call`

src/checkbox/checkbox-exported-tests.ts
```diff
@@ -0,0 +1,45 @@
...
+		};
+		return [{
+			name: "Basic checkbox tests",
+			tests: this.getTests(tests)
+		}];
+	}
+
+	constructor(configs) {
+		super(configs);
+		(this as any).tests = (this as any).tests.concat((this.constructor as typeof CheckboxExportedTests).structure((this as any).settings));
```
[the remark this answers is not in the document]
Donisius: Not including the `as any` and `as typeof CheckboxExportedTests` causes type check issues in typescript

src/exported-tests/component-tests.ts
Donisius: I don't think there aren't any other features we would need (besides the typescript formatting) from the base Component Tests file. Once that is available in a shared location we will definitely switch over to utilizing that 👍

Donisius: Thank you so much for reviewing!!

src/checkbox/checkbox.component.spec.ts
Donisius: That makes sense 👍

src/checkbox/checkbox-exported-tests.ts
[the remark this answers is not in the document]
elizabethsjudd: That was the same error we were getting too I believe... linters must not like assertion libraries in general than lol.  We have separate linting rules for our tests vs our components and we turned that one off globally for tests.  It'd be good to put our heads together and figure out what we'd like to do long-term with these tests.

src/checkbox/checkbox-exported-tests.ts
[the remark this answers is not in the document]
asudoh: To add some to @Donisius's point: `this.constructor as typeof CheckboxExportedTests` is used basically when you want to access to static property/method of `CheckboxExportedTests`. I guess `this as any` is used for assigning an arbitrary property.

src/checkbox/checkbox-exported-tests.ts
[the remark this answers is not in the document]
asudoh: Good discussion, those lint errors came from Chai's pattern using the getter of `.true`/`.false`, etc. as assertion test calls, basically:

```javascript
get false() {
  if (this.theValue !== false) {
    throw new AssertionError(errorMessage);
  }
}
```

That said, global ESLint rule for test definitely makes sense. I have done that for some projects, too. ([Example](https://github.com/carbon-design-system/carbon-custom-elements/blob/v1.0.0-beta.7/tests/spec/.eslintrc))

Just FYI BTW: https://medium.com/palantir/tslint-in-2019-1a144c2317a9

package.json
```diff
@@ -98,12 +98,14 @@
...
     "babel-loader": "8.0.5",
     "carbon-components": "10.9.2",
+    "chai": "^4.2.0",
     "codelyzer": "5.0.0",
     "commitizen": "3.0.7",
     "core-js": "3.1.3",
     "css-loader": "2.1.1",
     "cz-conventional-changelog-components": "1.0.0",
     "element-closest-polyfill": "1.0.0",
+    "exported-tests": "https://github.com/IBM/exported-tests.git",
```
cal-smith, line 108: We'll want to switch over to the npm version :+1:

package.json
```diff
@@ -121,6 +123,7 @@
     "karma-coverage-istanbul-reporter": "2.0.5",
     "karma-jasmine": "2.0.1",
     "karma-jasmine-html-reporter": "1.4.2",
+    "lodash-es": "^4.17.15",
```
cal-smith, line 126: If we still need this, this should be pinned

package.json
```diff
@@ -98,12 +98,14 @@
     "angular2-template-loader": "0.6.2",
     "babel-loader": "8.0.5",
     "carbon-components": "10.9.2",
+    "chai": "^4.2.0",
```
cal-smith, line 101: Ditto for this

cal-smith:
I'm going to mark this ready for review, because it looks great

src/exported-tests/component-tests.ts
```diff
@@ -0,0 +1,39 @@
...
+
+class ComponentTests {
```
cal-smith, line 3: Just want to double check if this class is still something we need to write out like this, or if there's an implementation we can use (or extend) from the `exported-tests` package?

cal-smith: If not then I think we're good to go with this PR 😁

Donisius: If I'm not mistaken @elizabethsjudd, I don't think the file is available in the `exported-tests` package. If it is we definitely want to extend that one.

elizabethsjudd: This file is not yet opened source.  We were planning on putting it wherever we put the shared UI unit tests which would be separate from the Architecture (stuff to talk about at our meeting on Tuesday)

src/checkbox/checkbox.component.spec.ts
```diff
@@ -81,4 +95,13 @@ describe("Checkbox", async() => {
...
 		expect(wrapper.onIndeterminateChange).toHaveBeenCalled();
 	});
+
+	describe("PAL exported tests", () => {
+		// Get checkbox from the fixture
+		element = setupFixture.debugElement.query(By.css("ibm-checkbox"));
+		setupFixture.detectChanges();
+		const instance = new CheckboxExportedTest({});
+		// tslint:disable-next-line
+		new BDDTestParser((instance as any).tests, setupFixture.debugElement.nativeElement);
```
cal-smith, line 105: It seems to me, that as long as the tests are reasonably self contained we shouldn't run into serious perf issues ... I'm having trouble finding any nodejs memory benchmarking tools, so I'm not sure we can actually investigate this as much as we would like 😞 In general I think we should make sure to keep the tests as flat as possible (minimize nesting) and take advantage of the parallel nature of most test tools to keep our perf in check. If we have functions that can be pulled out of the classes and just used when needed that can also help things.

cal-smith:
Once the checks pass I'm going to go ahead and merge this one 👍

PR #1167, opened 2020-03-31
fix(combobox): allow selected values to be updated through setValue

### LINKED ISSUE #1166: SetValue() of FormControl does not update displayed content of Combo box

Donisius:
Closes #1166

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -377,7 +377,6 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit {
 				}
 				this.selected.emit(event);
 			});
-			this.view.items = this.items;
 			// update the rest of combobox with any pre-selected items
 			// setTimeout just defers the call to the next check cycle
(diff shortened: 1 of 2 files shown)


src/combobox/combobox.component.ts
```diff
@@ -377,7 +377,6 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit {
 				}
 				this.selected.emit(event);
 			});
-			this.view.items = this.items;
```
cal-smith: How does removing this allow items to be updated via `setValue`/`writeValue`?

cal-smith:
Everything else looks good BTW, I'm just curious what effect removing that one line has?

src/combobox/combobox.component.ts
Donisius: It already was updating the values but this line has been overwriting the values that are set through set/write value

cal-smith: Interesting 🤔

PR #1170, opened 2020-04-01
feat(visiblity calculation): improve calculation of element's visibility within html/body element

Donisius:


Donisius:
> I'm guessing you've tested this already, but this definitely sorts the common case we've been seeing with partial vis?

The problem is not partial visibility of the element since that was already sorted out from a previous fix, but that the boundaries of the html/body tags do not reach overflow content if there exists `height: 100%` on those tags. This means that if there is enough content on the page to make an overflow appear, the elements that you need to scroll down for are considered invisible by the calculation since the boundaries do not go past the bottom of the viewport before scrolling. You can see this in our stories as well, just inspect the html tag and scroll down. I took that into account in the calculation and just see if the element is visible in the viewport just for the html/body tags

cal-smith:
Cool, that should be good then 👍

PR #1181, opened 2020-04-07
refactor(accordion-item): adds context input

babieste:
Closes IBM/carbon-components-angular#1180

#### Changelog

**Changed**

* Added a `context` input to `ibm-accordion-item`, which is a object that is passed to `[ngTemplateOutletContext]`

### DIFF
--- a/src/accordion/accordion-item.component.ts
+++ b/src/accordion/accordion-item.component.ts
@@ -23,7 +23,7 @@ import {
 				}">
 				{{!skeleton ? title : null}}
 			</p>
-			<ng-template *ngIf="isTemplate(title)" [ngTemplateOutlet]="title"></ng-template>
+			<ng-template *ngIf="isTemplate(title)" [ngTemplateOutlet]="title" [ngTemplateOutletContext]="context"></ng-template>
 		</button>
 		<div [id]="id" class="bx--accordion__content">
 			<ng-content *ngIf="!skeleton; else skeletonTemplate"></ng-content>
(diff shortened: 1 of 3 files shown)


src/accordion/accordion-item.component.ts
```diff
@@ -23,7 +23,7 @@ import {
 				}">
 				{{!skeleton ? title : null}}
 			</p>
-			<ng-template *ngIf="isTemplate(title)" [ngTemplateOutlet]="title"></ng-template>
+			<ng-template *ngIf="isTemplate(title)" [ngTemplateOutlet]="title" [ngTemplateOutletContext]="context"></ng-template>
```
Donisius, line 26:
```suggestion
			<ng-template
				*ngIf="isTemplate(title)"
				[ngTemplateOutlet]="title"
				[ngTemplateOutletContext]="context">
			</ng-template>
```

src/accordion/accordion-item.component.ts
```diff
@@ -38,6 +38,7 @@ import {
 export class AccordionItem {
 	static accordionItemCount = 0;
 	@Input() title: string | TemplateRef<any>;
+	@Input() context?: Object;
```
Donisius, line 41: I don't think this needs to be made into an optional input, or maybe we can have it something like `@Input() context: Object | null = null` since it seems that it takes both type Object or null https://angular.io/api/common/NgTemplateOutlet

babieste: I changed it, but I was thinking... If this input is not optional and the user doesn't write it, wouldn't it be `undefined` instead of `null`?

PR #1184, opened 2020-04-08
fix(modal): add automatic detection of scrollable elements

### LINKED ISSUE #1172: Modal overflow indicator is permanently enabled
It appears a new element has been recently added to the modal component which shows a gradient fade effect to the bottom of the modal content. This should be conditional, and allow the developer to make a choice of when and where they would enable this. 

https://github.com/IBM/carbon-components-angular/blob/94a3c9f1e18686b3652d218a96b389f850fd2ac5/src/modal/modal.component.ts#L101

If you see the React implementation, they have wrapped this particular element in a conditional rendering block - 
https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Modal/Modal.js#L399

I would make the suggestion to of appending a `*ngIf` to this element and a new input that sets a boolean value that controls the visibility of this element.

Donisius:
Closes #1172

### DIFF
--- a/src/modal/alert-modal.component.ts
+++ b/src/modal/alert-modal.component.ts
@@ -85,7 +85,7 @@ export class AlertModal extends BaseModal implements AfterViewInit {
 		@Inject("modalTitle") public modalTitle: string,
 		@Inject("modalContent") public modalContent: string,
 		@Inject("size") public size: string,
-		@Inject("hasScrollingContent") public hasScrollingContent: boolean | null = null,
+		@Inject("hasScrollingContent") public hasScrollingContent: boolean = null,
 		@Inject("buttons") public buttons = [],
 		@Inject("close") public onClose: Function
 	) {
--- a/src/modal/alert-modal.component.ts
+++ b/src/modal/alert-modal.component.ts
@@ -50,11 +50,12 @@ import { BaseModal } from "./base-modal.class";
 		<ibm-modal
 			[size]="size"
 			[theme]="modalType"
+			[hasScrollingContent]="hasScrollingContent"
 			[modalLabel]="modalTitle"
 			(overlaySelected)="dismissModal('overlay')">
 			<ibm-modal-header (closeSelect)="dismissModal('close')">
 				<p class="bx--modal-header__label bx--type-delta">{{modalLabel}}</p>
-      			<p class="bx--modal-header__heading bx--type-beta">{{modalTitle}}</p>
+				<p class="bx--modal-header__heading bx--type-beta">{{modalTitle}}</p>
 			</ibm-modal-header>
 			<div #content class="bx--modal-content">
 				<p [innerHTML]="modalContent"></p>
@@ -84,6 +85,7 @@ export class AlertModal extends BaseModal implements AfterViewInit {
 		@Inject("modalTitle") public modalTitle: string,
 		@Inject("modalContent") public modalContent: string,
 		@Inject("size") public size: string,
+		@Inject("hasScrollingContent") public hasScrollingContent: boolean | null = null,
 		@Inject("buttons") public buttons = [],
 		@Inject("close") public onClose: Function
 	) {
--- a/src/modal/alert-modal.interface.ts
+++ b/src/modal/alert-modal.interface.ts
@@ -45,6 +45,12 @@ export interface AlertModalData {
 	 */
 	buttons?: Array<ModalButton>;
 	/**
+	 * Specify whether the modal contains scrolling content. This property overrides the automcatic detection of the existance of scrolling
+	 * content. Set this property to `true` to force overflow indicator to show up or to `false` to force overflow indicator to disappear.
+	 * It is set to `null` by default which indicates not to override automatic detection.
+	 */
+	hasScrollingContent?: boolean | null;
+	/**
 	 * Callback for non-specific close events. `return false;` to prevent the modal from closing
 	 * Property trigger has a value of `overlay` or `close`
 	 */
--- a/src/modal/alert-modal.interface.ts
+++ b/src/modal/alert-modal.interface.ts
@@ -50,7 +50,7 @@ export interface AlertModalData {
 	 * show up or to `false` to force overflow indicator to disappear.
 	 * It is set to `null` by default which indicates not to override automatic detection.
 	 */
-	hasScrollingContent?: boolean | null;
+	hasScrollingContent?: boolean;
 	/**
 	 * Callback for non-specific close events. `return false;` to prevent the modal from closing
 	 * Property trigger has a value of `overlay` or `close`
--- a/src/modal/modal.component.ts
+++ b/src/modal/modal.component.ts
@@ -9,7 +9,8 @@ import {
 	OnInit,
 	Output,
 	ElementRef,
-	ViewChild
+	ViewChild,
+	AfterContentChecked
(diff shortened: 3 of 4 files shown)


src/modal/modal.component.ts
```diff
@@ -129,6 +129,11 @@ export class Modal implements AfterViewInit, OnInit, OnDestroy {
 	 */
 	@Input() modalLabel = "default";
 
+	/**
+	 * Specify whether the modal contains scrolling content
+	 */
+	@Input() hasScrollingContent = false;
```
cal-smith, line 135: We'll need to expose this to the alert modal ... also we'll need to consider how this affects users that are now taking advantage of the new default (hasScrollingContent = true) ...

Donisius: Do you think that it should be defaulted to true then?

cal-smith:
> Is there a way we can detect this instead of being explicit?

That would be cool too ... ideally we should provide both - default auto-detection, and explicit overrides for when the auto-detection fails

src/modal/alert-modal.interface.ts
```diff
@@ -44,6 +44,12 @@ export interface AlertModalData {
 	 * Array of `ModalButton`s
 	 */
 	buttons?: Array<ModalButton>;
+	/**
+	 * Specify whether the modal contains scrolling content. This property overrides the automcatic detection of the existance of scrolling
+	 * content. Set this property to `true` to force overflow indicator to show up or to `false` to force overflow indicator to disappear.
+	 * It is set to `null` by default which indicates not to override automatic detection.
+	 */
```
zvonimirfras, line 51: Typos and line length.

```suggestion
	/**
	 * Specify whether the modal contains scrolling content. This property overrides the automatic detection
	 * of the existence of scrolling content. Set this property to `true` to force overflow indicator to
	 * show up or to `false` to force overflow indicator to disappear.
	 * It is set to `null` by default which indicates not to override automatic detection.
	 */
```

src/modal/modal.component.ts
```diff
@@ -129,6 +133,13 @@ export class Modal implements AfterViewInit, OnInit, OnDestroy {
 	 */
 	@Input() modalLabel = "default";
 
+	/**
+	 * Specify whether the modal contains scrolling content. This property overrides the automcatic detection of the existance of scrolling
+	 * content. Set this property to `true` to force overflow indicator to show up or to `false` to force overflow indicator to disappear.
+	 * It is set to `null` by default which indicates not to override automatic detection.
+	 */
```
zvonimirfras, line 140:
```suggestion
	/**
	 * Specify whether the modal contains scrolling content. This property overrides the automatic
	 * detection of the existence of scrolling content. Set this property to `true` to force
	 * overflow indicator to show up or to `false` to force overflow indicator to disappear.
	 * It is set to `null` by default which indicates not to override automatic detection.
	 */
```

src/modal/modal.component.ts
```diff
@@ -152,6 +163,12 @@ export class Modal implements AfterViewInit, OnInit, OnDestroy {
 	 */
 	selectorPrimaryFocus = "[modal-primary-focus]";
 
+	/**
+	 * This is used within the component for automatic scrolling content detection. This can be overwritten
+	 * through public property `hasScrollingContent`.
+	 */
+	protected shouldShowScrollbar = false;
```
zvonimirfras, line 170: This could be a `get`ter or a function so there's no need for `ngAfterContentChecked`, or `onResize`.

Donisius: Changing it to a getter and removing the`ngAfterContentChecked`, or `onResize` makes automatic detection on the alert modal not work

src/modal/modal.component.ts
```diff
@@ -152,6 +163,11 @@ export class Modal implements AfterViewInit, OnInit, OnDestroy {
 	 */
 	selectorPrimaryFocus = "[modal-primary-focus]";
 
+	/**
+	 * This is used within the component for automatic scrolling content detection. This can be overwritten
+	 * through public property `hasScrollingContent`.
+	 */
+
```
zvonimirfras, line 170:
```suggestion
```

src/modal/modal.component.ts
```diff
@@ -205,4 +221,21 @@ export class Modal implements AfterViewInit, OnInit, OnDestroy {
 			}
 		}
 	}
+	/**
+	 * This determines whether or not the modal contains scrolling content automatically. If it fails at doing so,
+	 * you can use the `hasScrollingContent` property and manually set the overflow indicator on or off.
+	 *
+	 * If you want to force trigger a detection (ie. on window resize) you can reset the value of the modal
+	 * content.
+	 */
```
zvonimirfras, line 230:
```suggestion
	/**
	 * This detects whether or not the modal contains scrolling content.
	 *
	 * To force trigger a detection (ie. on window resize), change or reset the value of the modal content.
	 *
	 * Use the `hasScrollingContent` input to manually override the overflow indicator.
	 */
```

src/modal/alert-modal.interface.ts
```diff
@@ -44,6 +44,13 @@ export interface AlertModalData {
 	 * Array of `ModalButton`s
 	 */
 	buttons?: Array<ModalButton>;
+	/**
+	 * Specify whether the modal contains scrolling content. This property overrides the automatic detection
+	 * of the existence of scrolling content. Set this property to `true` to force overflow indicator to
+	 * show up or to `false` to force overflow indicator to disappear.
+	 * It is set to `null` by default which indicates not to override automatic detection.
+	 */
+	hasScrollingContent?: boolean | null;
```
cal-smith, line 53: Just `boolean` _should_ be sufficient

src/modal/modal.component.ts
```diff
@@ -129,6 +132,14 @@ export class Modal implements AfterViewInit, OnInit, OnDestroy {
 	 */
 	@Input() modalLabel = "default";
 
+	/**
+	 * Specify whether the modal contains scrolling content. This property overrides the automatic
+	 * detection of the existence of scrolling content. Set this property to `true` to force
+	 * overflow indicator to show up or to `false` to force overflow indicator to disappear.
+	 * It is set to `null` by default which indicates not to override automatic detection.
+	 */
+	@Input() hasScrollingContent: boolean | null = null;
```
cal-smith, line 141: ditto here, just `null` should be fine

src/modal/alert-modal.component.ts
```diff
@@ -84,6 +85,7 @@ export class AlertModal extends BaseModal implements AfterViewInit {
 		@Inject("modalTitle") public modalTitle: string,
 		@Inject("modalContent") public modalContent: string,
 		@Inject("size") public size: string,
+		@Inject("hasScrollingContent") public hasScrollingContent: boolean | null = null,
```
cal-smith, line 88: just `boolean` here too I think

src/modal/modal.service.ts
```diff
@@ -99,6 +99,7 @@ export class ModalService {
 				modalLabel: data.label || data.modalLabel,
 				modalTitle: data.title || data.modalTitle,
 				modalContent: data.content || data.modalContent,
+				hasScrollingContent: data.hasScrollingContent,
```
cal-smith, line 102: If we need explicit `null`s we should be sure to use `|| null` here as a fallback

PR #1185, opened 2020-04-08
`search.component`: Remove `role="search"` from `input` element

sjbeatle:
The search role is added to the **container element** that encompasses the items and objects that, as a whole, combine to create search functionality.
https://www.w3.org/TR/wai-aria-practices/examples/landmarks/search.html

#### Changelog

**Removed**

* `role="search"` from `input` element


sjbeatle:
> Cool, thanks for catching this!

My pleasure!
My first time contributing to a popular library that I wasn't paid to do so, lol.

Thanks for the great work! It's a very powerful Design System, with an excellent eye for accessibility. Cheers!

PR #1188, opened 2020-04-09
improvement(overflow-menu): adds custom trigger handler

damianpia:
Closes IBM/carbon-components-angular#1187

Adds custom trigger handler for overflow menu.

#### Changelog

**New**

Custom trigger handler for overflow menu support


src/dialog/overflow-menu/overflow-menu.component.ts
```diff
@@ -78,6 +83,11 @@ export class OverflowMenu {
 
 	@Input() placement: "bottom" | "top" = "bottom";
 
+	/**
+	 * This specifies is custom trigger template was provided
+	 */
+	@Input() customTrigger = false;
```
cal-smith, line 89: Maybe make this take a full template? Something like `@Inupt() customIcon: TemplateRef<any>` ... then the default icon can just be the default value to `customIcon` ... then we're also not running into potential conflicts with any content children selectors

cal-smith: Alternatively we could use `ngComponentOutlet` (https://angular.io/api/common/NgComponentOutlet) and have `customIcon` just take in a component class ...

damianpia: Thanks for the review! Changing the `customTrigger` type to `TemplateRef<any>` is a great idea, however setting it's default value is not a common practice - I was searching for solutions done in this repo, and looks like `*ngIf` has to stay.

src/dialog/overflow-menu/overflow-menu.component.ts
```diff
@@ -40,13 +41,14 @@ import { OverflowMenuDirective } from "./overflow-menu.directive";
 			class="bx--overflow-menu"
 			[placement]="placement"
 			tabindex="0">
-			<svg focusable="false" class="bx--overflow-menu__icon" width="3" height="15" viewBox="0 0 3 15">
+			<svg *ngIf="!customIcon" focusable="false" class="bx--overflow-menu__icon" width="3" height="15" viewBox="0 0 3 15">
```
zvonimirfras, line 44: Seems like `*ngIf="condition; else template` combo (like in accordion-item.component.ts for example) should work here.
That would make it clearer that it's a on-or-the-other situation here.

src/dialog/overflow-menu/overflow-menu.component.ts
```diff
@@ -78,6 +80,11 @@ export class OverflowMenu {
 
 	@Input() placement: "bottom" | "top" = "bottom";
 
+	/**
+	 * Sets the icon of trigger button
+	 */
+	@Input() customIcon: TemplateRef<any>;
```
zvonimirfras, line 86: Looks like `customTrigger` might be a more accurate name since users can pass anything in a template.

src/dialog/overflow-menu/overflow-menu.component.ts
damianpia: Added 👍

src/dialog/overflow-menu/overflow-menu.component.ts
damianpia: That's right, thanks :)

cal-smith: The joys of naming ahaha ... `customTrigger` totally works, sorry for jerking you around with `cutomIcon` 😆

PR #1190, opened 2020-04-10
feature: add input to set max length to the input inside combobox

Pablodotnet:
New input was added to combobox component to limit the characters on the input when the developer requires it, in this case the input is called `maxLength`

#### Changelog

**New**

* Sets maxlength attribute to input tag when `maxLength` input is defined

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -64,7 +64,7 @@ import { Observable } from "rxjs";
 					(keydown.enter)="clearSelected()"
 					role="button"
 					class="bx--tag--filter bx--list-box__selection--multi"
-          tabindex="0"
+					tabindex="0"
 					[title]="clearSelectionsTitle"
 					[attr.aria-label]="clearSelectionAria">
 					{{ pills.length }}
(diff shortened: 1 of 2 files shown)


src/combobox/combobox.component.ts
```diff
@@ -379,6 +383,9 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit {
 				}
 				this.selected.emit(event);
 			});
+			if (!!this.maxLength && !!this.elementRef.nativeElement.querySelector("input")) {
+				this.elementRef.nativeElement.querySelector("input").setAttribute("maxlength", this.maxLength);
```
Donisius, line 387: Is there a reason this couldn't be put directly on the `input` in the template? You can set the value to `null` and it won't have a restriction on the `maxlength`. You will probably have to change `[maxlength]` in to `[attr.maxlength]` however and set the type of the input to be something like `@Input() maxLength: number | null = null;`

Pablodotnet: Oh ok, great, I'll make that change now

cal-smith: `maxLength: number` should already include `null` ... or we should make it an optional `:?` ... agree that this should be the native `maxlength` - https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength

PR #1191, opened 2020-04-12
fix(table-toolbar): Improve batch text handling

yinglejia:
Closes IBM/carbon-components-angular#1186

Improve batch text handling to properly handle single/multiple items and also translations.

Example:

image

### DIFF
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -142,8 +142,9 @@ export default {
 		"ROW": "row"
 	},
 	"TABLE_TOOLBAR": {
-		"ACTION_BAR": "Table Action Bar",
-		"BATCH_TEXT": "items selected",
+		"ACTION_BAR": "Table action bar",
+		"BATCH_TEXT_SINGLE": "1 item selected",
+		"BATCH_TEXT_MULTIPLE": "{{count}} items selected",
 		"CANCEL": "Cancel"
 	},
 	"TABS": {
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -143,6 +143,7 @@ export default {
 	},
 	"TABLE_TOOLBAR": {
 		"ACTION_BAR": "Table action bar",
+		"BATCH_TEXT": "",
 		"BATCH_TEXT_SINGLE": "1 item selected",
 		"BATCH_TEXT_MULTIPLE": "{{count}} items selected",
 		"CANCEL": "Cancel"
--- a/src/table/toolbar/table-toolbar.component.ts
+++ b/src/table/toolbar/table-toolbar.component.ts
@@ -59,8 +59,13 @@ import { I18n, Overridable } from "../../i18n/i18n.module";
 			</div>
 			<div class="bx--batch-summary">
 				<p class="bx--batch-summary__para" *ngIf="count as n">
-					<span *ngIf="n === 1">{{_batchTextSingle.subject | async}}</span>
-					<span *ngIf="n !== 1">{{_batchTextMultiple.subject | i18nReplace: {count: n} | async}}</span>
+					<ng-container *ngIf="_batchTextLegacy.subject | async as legacyText; else batchTextBlock">
+						<span>{{n}}</span> {{legacyText}}
+					</ng-container>
+					<ng-template #batchTextBlock>
+						<span *ngIf="n === 1">{{_batchTextSingle.subject | async}}</span>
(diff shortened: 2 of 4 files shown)


src/i18n/en.ts
```diff
@@ -142,8 +142,9 @@ export default {
 		"ROW": "row"
 	},
 	"TABLE_TOOLBAR": {
-		"ACTION_BAR": "Table Action Bar",
-		"BATCH_TEXT": "items selected",
+		"ACTION_BAR": "Table action bar",
+		"BATCH_TEXT_SINGLE": "1 item selected",
```
cal-smith, line 146: So we don't break anyone this should just be `BATCH_TEXT` 👍

yinglejia: To make sure I understand correctly, you mean to use `"BATCH_TEXT": "1 item selected"` here? I think it would still break since the string has changed? Or should we keep the old string `"BATCH_TEXT": "items selected"` as-is, but don't reference it in code?

cal-smith: The string can change, my primary concern is users that have customized the text suddenly not seeing it ... We may actually be best to leave `BATCH_TEXT_SINGLE` as it was, and if we don't have `BATCH_TEXT_SINGLE` and `BATCH_TEXT_MULTIPLE` we just switch back to `BATCH_TEXT` ... that way people can adopt the new translation keys, and the old way doesn't break anyone

yinglejia: OK, given how `I18n` class works, it means `BATCH_TEXT_SINGLE` and `BATCH_TEXT_MULTIPLE` will always be available and the English version will be the fallback. So we can't do the check you mentioned.

Alternatively, maybe we can check `BATCH_TEXT`, and it is NOT `items selected` then we use it, otherwise we'll use `BATCH_TEXT_SINGLE` and `BATCH_TEXT_MULTIPLE`? This would preserve `BATCH_TEXT` in case someone customized it...

cal-smith: Hmm ... What we _could_ do, is remove/null `BATCH_TEXT` in the shipped `EN.ts`, and check for it's existence - anyone consuming the library without modifications will get the new (better) translations, and if `BATCH_TEXT` is set we can assume that was intentional

yinglejia: OK, sounds good

yinglejia: @cal-smith so I set `BATCH_TEXT` to an empty string in en.ts, and found that I had to fix `getValueFromPath` since it will throw an exception for empty string (which doesn't make sense).

src/table/toolbar/table-toolbar.component.ts
```diff
@@ -58,8 +58,14 @@ import { I18n, Overridable } from "../../i18n/i18n.module";
 				<button ibmButton="primary" class="bx--batch-summary__cancel" (click)="onCancel()">{{_cancelText.subject | async}}</button>
 			</div>
 			<div class="bx--batch-summary">
-				<p class="bx--batch-summary__para">
-					<span>{{count}}</span> {{_batchText.subject | async}}
+				<p class="bx--batch-summary__para" *ngIf="count as n">
+					<ng-container *ngIf="_batchTextLegacy.subject | async as legacyText; else batchTextBlock">
```
cal-smith, line 62: Nice!

src/i18n/en.ts
cal-smith: Interesting 🤔 but great work getting that fixed!

yinglejia:
@zvonimirfras @theiliad could you please review this PR?

src/i18n/en.ts
```diff
@@ -142,8 +142,10 @@ export default {
 		"ROW": "row"
 	},
 	"TABLE_TOOLBAR": {
-		"ACTION_BAR": "Table Action Bar",
-		"BATCH_TEXT": "items selected",
+		"ACTION_BAR": "Table action bar",
+		"BATCH_TEXT": "",
```
theiliad, line 146: do we still need this?

yinglejia: @theiliad yes it is needed for the `_batchTextLegacy: Overridable = this.i18n.getOverridable("TABLE_TOOLBAR.BATCH_TEXT")` call for compatibility reason (in case someone customized that message in their project).

We have to leave an empty string in en.ts like this, otherwise `i18n.getValueFromPath` will thrown an exception when `i18n.getOverridable("TABLE_TOOLBAR.BATCH_TEXT")` is called.

PR #1206, opened 2020-04-19
feat(data-table): Support disabled table rows

yinglejia:
Closes IBM/carbon-components-angular#1075

Add support for disabled table rows, which can't be selected in multi-select mode. The same functionality is available in Carbon React, refer:

https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/DataTable/README.md#rows

https://react.carbondesignsystem.com/?path=/story/datatable--with-options

image

#### Changelog

**New**

- Added `TableRow` and `disabled` property.
- Added 'click' event to `Checkbox` component to allow parent component to control `checked/indeterminate` properties.

**Changed**

- Fixed `checked/indeterminate` state handling in `Checkbox` component.


src/table/table.stories.ts
```diff
@@ -71,6 +72,26 @@ const getProps = (more = {}) => {
 	}, more);
 };
 
+function getModelWithDisabledRows() {
+	const disabledModel = new TableModel();
+	const row1 = new TableRow();
+	row1.disabled = true;
+	row1.push(new TableItem({data: "Name 1"}));
```
cal-smith, line 79: Since you're extending the `Array` constructor, I believe you can pass the items in when you create the row 

```typescript
const row1 = new TableRow(
	row1.push(new TableItem({data: "Name 1"}));
	row1.push(new TableItem({data: "Disabled 1"}));
);
row1.disabled = true;
```

PR #1210, opened 2020-04-20
fix(combobox): allow value to be reset through ngModel

### LINKED ISSUE #1203: Combobox with ngModel does not propagate change properly when resetting model
Steps to reproduce:
1) Set a selected value to ngModel
2) Try to clear the selected value by setting ngModel to an empty string

### LINKED ISSUE #1228: When using the Combobox component with ngModel, updating the model does not update the combobox input/placeholder
When using ngModel with the combobox, the dropdown list is updated correctly when the model is changed, but the placeholder in the combobox input is not updated with the new value

Update the ngModel for a combobox (single selection) with a new value - the item is selected in the dropdown but the new content is not updated in the combobox input

Donisius:
Closes #1203
Closes #1228


cal-smith:
Somehow we've ended up with two almost identical PRs fixing similar problems ... I think I'm going to use this one, since it will end up fixing a superset of the two issues

cal-smith:
Other PR https://github.com/IBM/carbon-components-angular/pull/1229

PR #1217, opened 2020-04-21
Merge next to master

cal-smith:


cal-smith:
Waiting on https://travis-ci.com/github/IBM/carbon-components-angular/builds/161324090

cal-smith:
Once this is in I'll reset the `next` tag to be `@next` 👍 

Our whole tag situation is a bit messier than it should be right now ... at some point I'll dig into it and see if I can figure out a simpler way to handle everything

PR #1229, opened 2020-04-24
Fix issue with combobox ngModel

fionamurphy:
Closes IBM/carbon-components-angular# #1228

Fixes issue with the combobox input not updating when ngModel for a combobox is changed


cal-smith:
Huh, it seems like two almost identical PRs to fix two different issues were created (see this one: https://github.com/IBM/carbon-components-angular/pull/1210) I'm going to go with #1210, since it seems like it actually fixes a superset of the two issues

cal-smith:
Closing in favour of #1210

PR #1232, opened 2020-04-26
fix(datepicker): fix weekday inconsistencies

### LINKED ISSUE #1231: weekday format inconsistencies in DatePicker calendar
## Detailed description

there are some inconsistencies between angular version of calendar and the official documentation. 

bug

Date Picker

Chrome

> version

3.28.0

## Additional information

* Screenshots or code
angular:
image

official docs:
image

Donisius:
Closes #1231

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -12,7 +12,8 @@ import {
 	SimpleChanges,
 	AfterViewChecked,
 	AfterViewInit,
-	ViewChild
+	ViewChild,
+	AfterContentInit
 } from "@angular/core";
 import rangePlugin from "flatpickr/dist/plugins/rangePlugin";
 import flatpickr from "flatpickr";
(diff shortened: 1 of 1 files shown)


src/datepicker/datepicker.component.ts
```diff
@@ -238,6 +239,11 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked, After
 		}
 	}
 
+	ngAfterContentInit() {
+		(languages.default.en.weekdays.shorthand as string[])
+			= languages.default.en.weekdays.shorthand.map(day => day === "Thu" || day ===  "T" ? "Th" : day.charAt(0));
```
cal-smith, line 244: Just curious how `carbon-react` handles this? This works, but I wonder if there's a simpler way? Also maybe make the map function a bit more explicit - that's a pretty dense oneliner, so it's kinda tricky to parse.

Donisius: They handle it in a similar way https://github.com/carbon-design-system/carbon/blob/v10.11.0/packages/components/src/components/date-picker/date-picker.js#L32-L39

Donisius: I'll make the map a little less dense though 😅

PR #1235, opened 2020-04-27
fix: use factories to ensure services are provided globally

cal-smith:


cal-smith:
> Too many internal NG-related changes in this PR, would be worth to get someone else with more recent angular experience to take a look but LGTM

No worries, as long as there isn't anything obviously wonky with the code ... This is basically just adding back code that was live in v3

PR #1237, opened 2020-04-27
fix(list): check that parentElement exists before trying to manipulate properties on it

BitForger:
Closes IBM/carbon-components-angular#1236

Fixes lists needing native HTML element wrappers when being used inside components like `ibm-accordion`

#### Changelog

**Changed**

* Check that the element has a parent before trying to check the tag name of the parent

### DIFF
--- a/src/list/list.directive.ts
+++ b/src/list/list.directive.ts
@@ -26,7 +26,7 @@ export class List {
 	}
 
 	@HostBinding("class.bx--list--nested") get nested() {
-		return this.elementRef.nativeElement.parentElement.tagName === "LI";
+		return (this.elementRef.nativeElement.parentElement && this.elementRef.nativeElement.parentElement.tagName === "LI");
 	}
 
 	constructor(protected elementRef: ElementRef) {}


src/list/list.directive.ts
```diff
@@ -26,7 +26,7 @@ export class List {
 	}
 
 	@HostBinding("class.bx--list--nested") get nested() {
-		return this.elementRef.nativeElement.parentElement.tagName === "LI";
+		return (this.elementRef.nativeElement.parentElement && this.elementRef.nativeElement.parentElement.tagName === "LI");
```
cal-smith, line 29: Might be a bit tidier to just wrap this in an `if` statement 🤔

BitForger: I broke it out to see how it would look, like so: 
```typescript
if (this.elementRef.nativeElement.parentElement) {
	return this.elementRef.nativeElement.parentElement.tagName === "LI";
} else {
	return false;
}
``` 
The correct way of doing that would be to move all conditional checks into the `if` line:
```typescript
if (this.elementRef.nativeElement.parentElement && this.elementRef.nativeElement.parentElement.tagName === "LI") {
	return true;
} else {
	return false;
}
```
I feel like there's not much benefit as most js/ts devs are used to this pattern of checking that a property exists and is truthy.

Once you move to TypeScript@>3.7 it would look better with optional chaining since you could just do
```typescript
return this.elementRef.nativeElement.parentElement?.tagName === "LI";
```

cal-smith: >Once you move to TypeScript@>3.7 it would look better with optional chaining since you could just do
> 
> ```typescript
> return this.elementRef.nativeElement.parentElement?.tagName === "LI";
> ```

That's going to be so nice 🤩 

Makes sense to me otherwise 👍

PR #1239, opened 2020-04-28
fix(datepicker): add missing icon in the datepicker

juliusstoerrle:
Closes IBM/carbon-components-angular#1238

I did not do any local tests (I did the change online without a local dev env), but checked the netlify preview and it now shows the icon as expected.


juliusstoerrle:
I did not do any local tests (I did the change online without a local dev env), but checked the netlify preview and it now shows the icon as expected.

PR #1241, opened 2020-04-30
fix(data-table): Fix table focus/tabindex handling

yinglejia:
Closes IBM/carbon-components-angular#1242

Fix data table focus/tabindex handling.

- `Table` component incorrectly sets `tabindex` to `-1` when a checkbox cell is clicked, which makes the cell unfocusable. This PR fixes that.

- For `TableToolbar` component, the buttons inside `batch-actions` and the buttons inside `toolbar-content` need to have `tabindex` set to `0` or `-1` properly depending on selection (since only one group of buttons will be shown). Otherwise the buttons may disappear when tabbing into the toolbar (!). Refer below screen shot.

image

Carbon React does the same thing: https://github.com/carbon-design-system/carbon/blob/db99425fc9164a8868a006e791fa7d1f116423a1/packages/react/src/components/DataTable/TableBatchActions.js#L53

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}

### DIFF
--- a/src/table/table.stories.ts
+++ b/src/table/table.stories.ts
@@ -156,27 +156,27 @@ storiesOf("Components|Table", module).addDecorator(
 				<h4 ibmTableHeaderTitle>{{title}}</h4>
 				<p ibmTableHeaderDescription>{{description}}</p>
 			</ibm-table-header>
-			<ibm-table-toolbar [model]="model" [batchText]="batchText">
+			<ibm-table-toolbar [model]="model" [batchText]="batchText" #toolbar>
 				<ibm-table-toolbar-actions>
-					<button ibmButton="primary">
+					<button ibmButton="primary" [tabindex]="toolbar.selected ? 0 : -1">
 						Delete
 						<ibm-icon-delete size="16" class="bx--btn__icon"></ibm-icon-delete>
 					</button>
-					<button ibmButton="primary">
+					<button ibmButton="primary" [tabindex]="toolbar.selected ? 0 : -1">
 						Save
 						<ibm-icon-save size="16" class="bx--btn__icon"></ibm-icon-save>
 					</button>
-					<button ibmButton="primary">
+					<button ibmButton="primary" [tabindex]="toolbar.selected ? 0 : -1">
 						Download
 						<ibm-icon-download size="16" class="bx--btn__icon"></ibm-icon-download>
 					</button>
 				</ibm-table-toolbar-actions>
 				<ibm-table-toolbar-content>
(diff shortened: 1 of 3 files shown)


src/table/table.stories.ts
```diff
@@ -156,27 +156,27 @@ storiesOf("Components|Table", module).addDecorator(
 				<h4 ibmTableHeaderTitle>{{title}}</h4>
 				<p ibmTableHeaderDescription>{{description}}</p>
 			</ibm-table-header>
-			<ibm-table-toolbar [model]="model" [batchText]="batchText">
+			<ibm-table-toolbar [model]="model" [batchText]="batchText" #toolbar>
```
cal-smith, line 159: It would be ideal if we could handle this switching (semi) automatically ... This is still a viable pattern for other components that might be passed in, but for `ibmButton`s we could just set the `tabindex`s automatically

src/table/toolbar/table-toolbar.component.ts
```diff
@@ -55,7 +55,14 @@ import { I18n, Overridable } from "../../i18n/index";
 			[attr.aria-label]="actionBarLabel.subject | async">
 			<div class="bx--action-list">
 				<ng-content select="ibm-table-toolbar-actions"></ng-content>
-				<button ibmButton="primary" class="bx--batch-summary__cancel" (click)="onCancel()">{{_cancelText.subject | async}}</button>
+				<button
+					ibmButton="primary"
+					class="bx--batch-summary__cancel"
+					[tabindex]="selected ? 0 : -1"
+					(click)="onCancel()"
+				>
```
cal-smith, line 63: Generally we leave the closing `>` on the same line as the last attribute :+1:

src/table/table.stories.ts
yinglejia: I like the idea, but how/where do we set the `tabindex`s automatically for `ibmButton`s? Inside the `ibmButton` directive? Hopefully the consumers don't want to control `tabindex` themselves, otherwise we'll have conflict.

On a side note, I just checked the Carbon React stories, and it seems they did the same thing in story code and not in the data table code:

https://github.com/carbon-design-system/carbon/blob/1a4cdb1051fb587f406aa082c99717c2ba6f3c75/packages/react/src/components/DataTable/stories/with-batch-actions.js#L63

yinglejia: @cal-smith any thoughts on this?

cal-smith: Sorry! This works for now 👍 we can investigate if there's an easier solution later!

PR #1251, opened 2020-05-06
Check view is initialised in ngOnChanges

fionamurphy:
Closes IBM/carbon-components-angular#1250

Avoid errors for this.view being undefined if it has yet to be initialised


fionamurphy:
@theiliad  I've force pushed the changes again to include the keyword for the semantic build release

PR #1254, opened 2020-05-07
refactor(ellipsis tooltip): create dialog content on hover

Donisius:
This allows the ellipsis tooltip to generate the dialog content on `hover`, previously it only worked when trigger is `click`. Also prevents an empty tooltip from being generated when there is no overflow.


Donisius:
Closing in favour of #1293

PR #1258, opened 2020-05-11
feat(ellipsis tooltip): toggle dialog on hover - v3

Donisius:

### DIFF
--- a/src/dialog/tooltip/ellipsis-tooltip.directive.ts
+++ b/src/dialog/tooltip/ellipsis-tooltip.directive.ts
@@ -5,7 +5,8 @@ import {
 	ElementRef,
 	Injector,
 	ComponentFactoryResolver,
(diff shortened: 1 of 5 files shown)


src/dialog/tooltip/ellipsis-tooltip.directive.ts
```diff
@@ -35,12 +36,32 @@ export class EllipsisTooltip extends TooltipDirective {
 	 * @returns null
 	 * @memberof EllipsisTooltip
 	 */
-	toggle() {
+	generateTooltipContent() {
```
zvonimirfras, line 39: perhaps `updateTooltipContent` might be more accurate?

Donisius: true!

PR #1260, opened 2020-05-11
feat(datepicker): add readOnly input

Donisius:


Donisius:
> I can see a few potential issues (especially with regards to a11y and UI consistency - "why can't I type in this one, but I can type in this other one?")

I am very inclined to agree with that... It wouldn't make sense for an input to not allow a user to type something without the input being disabled since that's how every other input is within the library.

cal-smith:
If we can make the whole component read only and non-interactive then we can totally do that ... We may still want to open an issue on the Carbon side anyway, to make sure we're not going too far into the weeds

Donisius:
> If we can make the whole component read only and non-interactive then we can totally do that ... We may still want to open an issue on the Carbon side anyway, to make sure we're not going too far into the weeds

The original request was to make it so that manually entering a date wasn't possible and the only way to change the date would be through the calendar. I'm not sure what benefits/use cases that can be gained from making it completely non-interactive.

cal-smith:
> The original request was to make it so that manually entering a date wasn't possible and the only way to change the date would be through the calendar. 

That seems unnecessarily user hostile 😛 I can't imagine a good reason for wanting to do that, especially since it doesn't seem to be supported by Carbon.

> I'm not sure what benefits/use cases that can be gained from making it completely non-interactive.

Me neither! But that's what `readonly` is 😆

Donisius:
> That seems unnecessarily user hostile 😛 I can't imagine a good reason for wanting to do that, especially since it doesn't seem to be supported by Carbon.

Yeah 😆 although it seems to be supported on vue http://vue.carbondesignsystem.com/?path=/story/components-cvdatepicker--single

cal-smith:
> Yeah 😆 although it seems to be supported on vue http://vue.carbondesignsystem.com/?path=/story/components-cvdatepicker--single

I think that's a bug on Vue's side though 😉 Vanilla/React is the reference implementation, so it's safe to assume deviations by anyone else is a bug 😛

cal-smith:
I'm going to close this since we don't have clear support from Carbon or a well defined usecase. This is something that really needs to be discussed with the Carbon design team.

PR #1272, opened 2020-05-18
feat(header): add option to use angular router attributes instead of events

BitForger:
This PR is to address a conversation I had with @cal-smith over on slack #carbon-ng. Per that conversation I tried using the `href` and `route` attributes to allow a user to right click and open in new tab. The primary issue we ran into with that was the active status of the link was no longer being tracked. 

TLDR; this adds a flag on the `ibm-header` and `ibm-header-item` that uses the `@angular/router` `routerLink` and `routerLinkActive` attributes on a standard `<a>` tag.

I don't know if this falls under the direction you all have for the project or if it's the desired way to implement it.

#### Changelog

**New**

* add `useRouter` input to `ibm-header` and `ibm-header-item` that changes the template to use angular router's attributes

### DIFF
--- a/src/ui-shell/header/header-item.component.ts
+++ b/src/ui-shell/header/header-item.component.ts
@@ -15,14 +15,24 @@ import { Router } from "@angular/router";
 	selector: "ibm-header-item",
 	template: `
 		<li style="height: 100%">
-			<a
-				class="bx--header__menu-item"
-				role="menuitem"
-				tabindex="0"
-				[href]="href"
-				(click)="navigate($event)">
-				<ng-content></ng-content>
-			</a>
+			<ng-container [ngSwitch]="useRouter">
+				<a *ngSwitchCase="false"
+				   class="bx--header__menu-item"
+				   role="menuitem"
+				   tabindex="0"
+				   [href]="href"
+				   (click)="navigate($event)">
+					<ng-content></ng-content>
+				</a>
+				<a *ngSwitchCase="true"
+				   class="bx--header__menu-item"
+				   role="menuitem"
+				   tabindex="0"
+				   [routerLink]="route"
+				   [routerLinkActive]="activeLinkClass">
+					<ng-content></ng-content>
+				</a>
+			</ng-container>
 		</li>
 	`
 })
@@ -38,6 +48,16 @@ export class HeaderItem {
 	get href() {
 		return this.domSanitizer.bypassSecurityTrustUrl(this._href) as string;
 	}
+
+	/**
+	 * Use the routerLink attribute on <a> tag for navigation instead of using event handlers
+	 */
+	@Input() useRouter: boolean = false;
+
+	/**
+	 * String or array of string class names to apply when active
+	 */
+	@Input() activeLinkClass: string | string[];
 
 	/**
 	 * Array of commands to send to the router when the link is activated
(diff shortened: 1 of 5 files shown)


BitForger:
Created as draft because I'd like to get some input on it before submitting it fully.

src/ui-shell/header/header-item.component.ts
```diff
@@ -15,14 +15,24 @@ import { Router } from "@angular/router";
...
+				   [href]="href"
+				   (click)="navigate($event)">
+					<ng-content></ng-content>
+				</a>
+				<a *ngSwitchCase="true"
+				   class="bx--header__menu-item"
+				   role="menuitem"
+				   tabindex="0"
+				   [routerLink]="route"
+				   [routerLinkActive]="activeLinkClass">
```
cal-smith, line 32: What are the differences between this approach and directly invoking `this.router.navigate`? If they're effectively the same, or if this way is the same but better, I would say we just switch over to this approach entirely

BitForger: The biggest difference is that it supports right click and open in new tab and doesn't send users to a blank void page when doing it while maintaining router navigation and adds active route class support.

cal-smith: Gotcha 👍 I'm assuming `routerLink` lets us use raw `href`s as well?

If so it sounds like both approaches are functionally identical, but using `routerLink` directly gives us better UX and more features ... in that case - as long as it's possible to keep the `navigation` event for compat reasons - I would say we just switch over to directly using `routerLink`

BitForger: > Gotcha 👍 I'm assuming `routerLink` lets us use raw `href`s as well?

`routerLink` doesn't accept urls you'd normally put in `href`s but you can use one or the other on the `a` tag.
I would use something like this
```ts
ngOnInit() {
	if (this.route && typeof this._href !== 'undefined') {
		throw new Error('You must only pass one of "route" or "href"')
	}
}
```
to restrict the use to one or the other. In the end though, this requires template switching as having the `routerLink` directive on an anchor tag prevents the `href` attribute from working normally. Unfortunately passing null to the directive doesn't keep it from binding to the element.

I see now why it's currently written the way it is since it's a bit more of an elegant approach, code-wise, albeit rigid in it's behavior.

> If so it sounds like both approaches are functionally identical, but using `routerLink` gives us better UX and more features ... in that case - as long as it's possible to keep the `navigation` event for compat reasons - I would say we just switch over to `routerLink` and drop the other method as much as possible

Keeping the `navigation` event is kind of tricky. I couldn't really emulate the same results as it was doing. I had eventually drilled down to a router events subscription like so
```ts
const subscription = this.router.events
	.pipe(filter(value => value instanceof NavigationEnd))
	.pipe(throttle((value: NavigationEnd) => interval(10000)))
	.pipe(filter((value: NavigationEnd) => value.url.replace(/^\//, '') === this.route[0]))
	.pipe(tap(event => console.log(event)))
	.pipe(mapTo(true))
	.toPromise();
this.navigation.emit(subscription);
```
but it was sending back multiples instead of just a single event.

I'm not sure of the use cases behind it since it is pretty limited and only responds true or false to the route you go to. If I needed to hook into the router events I'm more likely to subscribe to the `router.events` observable as I would have more data to work with and more flexibility.

cal-smith: Ah, so yeah, we've definitely coded ourselves into a bit of a box here then ... We can work with this for now, and figure out which method to promote and which to deprecate down the line ... There's a lot of API overlap so that does make things simpler, and `useRouter` keeps this from breaking anyone

BitForger:
So there's a bit of a bug here with how it's implemented.. apparently you can only have one `ng-content` for whatever reason. I need to make a quick tweak before this is safe to pull in

PR #1275, opened 2020-05-19
feat(modal): add form field support for modal

### LINKED ISSUE #1156: Modal doesn't support style examples without doing CSS override
Carbon design shows form fields going to within 16px of right edge but Modal right padding 20% stops this from happening.  I can override with CSS but would like a better way.
Maybe a parameter to pass to modal to control if it uses 20% right side padding or 16px padding?

https://www.carbondesignsystem.com/components/modal/style shows this example but there doesn't seem to be a way to achieve this without overriding CSS.
image

Donisius:
Closes #1156

### DIFF
--- a/src/modal/alert-modal.interface.ts
+++ b/src/modal/alert-modal.interface.ts
@@ -29,6 +29,10 @@ export interface AlertModalData {
 	 */
 	buttons?: Array<ModalButton>;
 	/**
+	 * Provide whether the modal content has a form element.
+	 */
+	hasForm?: boolean;
+	/**
 	 * Specify whether the modal contains scrolling content. This property overrides the automatic detection
 	 * of the existence of scrolling content. Set this property to `true` to force overflow indicator to
 	 * show up or to `false` to force overflow indicator to disappear.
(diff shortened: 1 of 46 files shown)


src/modal/alert-modal.interface.ts
```diff
@@ -28,6 +28,10 @@ export interface AlertModalData {
 	 * Array of `ModalButton`s
 	 */
 	buttons?: Array<ModalButton>;
+	/**
+	 * Provide whether the modal content has a form element.
+	 */
+	hasForm?: boolean;
```
cal-smith, line 34: Considering we don't allow form elements in the alert modal I'm not sure how useful this is 🤔

cal-smith: Really we'll want a separate `FormModal` to handle that situation

src/modal/modal.component.ts
```diff
@@ -180,14 +185,20 @@ export class Modal implements AfterViewInit, OnChanges {
 		}
 	}
 
-
 	/**
 	 * Set document focus to be on the modal component after it is initialized.
 	 */
 	ngAfterViewInit() {
+		const modalContent = this.modal.nativeElement.querySelector(".bx--modal-content");
```
cal-smith, line 192: modal content should be a directive (`ibmModalContent`), and `hasForm` should be an input on it 👍 That way we're not messing around with classnames directly (kinda fragile)

PR #1280, opened 2020-05-19
fix(table): fix accessibility issues - v3

### LINKED ISSUE #1106: Table Component - Table header th ids and checkbox names are duplicated when there are more than one table in the page
## Detailed description

- Our AVT team has found that the default table `<th>` ids are duplicated when there were more than one table, especially with same structure, in the page.
Also, if selection is enabled for table, all checkboxes inside `<ibm-checkbox>` have the same`"name"` attributes as `"undefined"`, which also causes checkbox grouping issue.

- **Bug**

- Table

- Chrome

- V3.40.4

- IBM ADDI

Donisius:
Closes #1106

### DIFF
--- a/src/radio/radio.component.ts
+++ b/src/radio/radio.component.ts
@@ -39,6 +39,7 @@ import { RadioChange } from "./radio-change.class";
 		<div *ngIf="skeleton" class="bx--radio-button bx--skeleton"></div>
 		<label
 			class="bx--radio-button__label"
+			[attr.aria-label]="ariaLabel"
 			[ngClass]="{
 				'bx--skeleton': skeleton
 			}"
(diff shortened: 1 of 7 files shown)


src/radio/radio.component.ts
```diff
@@ -96,6 +97,11 @@ export class Radio {
 	 * The id for the `Radio`.
 	 */
 	@Input() id = `radio-${Radio.radioCount++}`;
+	/**
+	 * Used to set the `aria-label` attribute on the input label.
+	 */
+	// tslint:disable-next-line:no-input-rename
```
cal-smith, line 103: I don't think we need the tslint line here?

PR #1281, opened 2020-05-20
fix(dropdown): moved expanded class as HostBind; added guard for boxMenu in dropdown service

Licen-it:
Closes IBM/carbon-components-angular#1276

Moved the `bx--list-box--expanded` class as HostBind of the DropdownList so that it will cover in one place both append inline and append to body cases. Also added `null` guard for boxMenu in the DropdownService to avoid console errors.

#### Changelog

**Changed**

* `bx--list-box--expanded` is now added as HostBind of the DropdownList component
* added guard for boxMenu in DropdownService

### DIFF
--- a/src/dropdown/dropdown.service.ts
+++ b/src/dropdown/dropdown.service.ts
@@ -111,22 +111,24 @@ export class DropdownService implements OnDestroy {
 
 		const boxMenu = menuRef.querySelector(".bx--list-box__menu");
 
-		// If the parentRef and boxMenu are in a different left position relative to the
-		// window, the the boxMenu position has already been flipped and a check needs to be done
-		// to see if it needs to stay flipped.
-		if (parentRef.getBoundingClientRect().left !== boxMenu.getBoundingClientRect().left) {
-			// The getBoundingClientRect().right of the boxMenu if it were hypothetically flipped
-			// back into the original position before the flip.
-			const testBoxMenuRightEdgePos =
-				parentRef.getBoundingClientRect().left - boxMenu.getBoundingClientRect().left + boxMenu.getBoundingClientRect().right;
+		if (boxMenu) {
+			// If the parentRef and boxMenu are in a different left position relative to the
+			// window, the the boxMenu position has already been flipped and a check needs to be done
+			// to see if it needs to stay flipped.
+			if (parentRef.getBoundingClientRect().left !== boxMenu.getBoundingClientRect().left) {
+				// The getBoundingClientRect().right of the boxMenu if it were hypothetically flipped
+				// back into the original position before the flip.
+				const testBoxMenuRightEdgePos =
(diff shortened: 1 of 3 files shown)


src/dropdown/dropdown.service.ts
```diff
@@ -111,22 +111,24 @@ export class DropdownService implements OnDestroy {
...
+				// back into the original position before the flip.
+				const testBoxMenuRightEdgePos =
+					parentRef.getBoundingClientRect().left - boxMenu.getBoundingClientRect().left + boxMenu.getBoundingClientRect().right;
+
+				if (testBoxMenuRightEdgePos > (window.innerWidth || document.documentElement.clientWidth)) {
+					leftOffset = parentRef.offsetWidth - boxMenu.offsetWidth;
+				}
+			// If it has not already been flipped, check if it is necessary to flip, ie. if the
+			// boxMenu is outside of the right viewPort.
+			} else if (boxMenu.getBoundingClientRect().right > (window.innerWidth || document.documentElement.clientWidth)) {
```
theiliad, line 129: Not sure if there are Gatsby-like frameworks that build angular in an SSR environment. Would the direct use of `window` reference cause an issue there?

Licen-it: I'm not familiar with what you describe, but i guess it would provide a global window object to mock browser behaviour. Anyway this line was already there...my change in this file is just putting everything using `boxMenu` under an `if` checking it to be defined.

cc @Donisius who introduced the line you commented...he might better answer your question

theiliad: In the case of Gatsby for instance, it wouldn't mock `window` by default, so you'd write something like `const isBrowser = typeof window !== "undefined"`

cal-smith: > Not sure if there are Gatsby-like frameworks that build angular in an SSR environment. Would the direct use of `window` reference cause an issue there?

Yeah, Angular Universal is Angular's SSR framework ... We have a number of areas with similar issues, eventually we'll have a service to mock out `window` which will let us solve all of these issues at once ... I just haven't gotten around to it yet 😛

src/dropdown/list/dropdown-list.component.ts
```diff
@@ -163,6 +164,8 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 	 * item selection.
 	 */
 	@Input() type: "single" | "multi" = "single";
+
+	@HostBinding("class.bx--list-box--expanded") hostClass = true;
```
cal-smith, line 168: I'm not sure we want this to be a `HostBinding` in this case ... Eventually we'll want to apply the `expanded` class conditionally, without removing the dropdown list from the DOM, so we can pick up on the opening animation Carbon supplies. Right now we have a pretty clear separation of concerns (the dropdown list doesn't need to know when it's open, just how to render the list) and I'd like to keep it that way if possible.

Everything else looks good, I'm just not sure this is the right approach for `expanded`!

PR #1286, opened 2020-05-20
feature(inline-loading): Added `state` property and error state style.

stanislavgeorgiev:

Streamlined component's state management to a single `state` property as opposed to the 2 flags `isActive` and `success`.

#### Changelog

**New**

* Added a `@Input() state: InlineLoadingState` property to better handle the different states of the inline loading component.
* Added `errorText: string` for the error text styling.

**Changed**

* Changed `isActive` and `success` to derive from `state` for backward compatibility.

**Removed**

* Removed `_success` in favor of `state`

### DIFF
--- a/src/inline-loading/inline-loading.component.ts
+++ b/src/inline-loading/inline-loading.component.ts
@@ -6,6 +6,19 @@ import {
 	HostBinding
 } from "@angular/core";
 
+export enum InlineLoadingState {
+	/** It hides the whole component. */
+	Hidden = "hidden",
+	/** It shows the `loadingText` but no loading animation. */
+	Inactive = "inactive",
+	/** It shows the `loadingText` with loading animation. */
+	Active = "active",
+	/** It shows the `successText` with a success state. */
+	Finished = "finished",
+	/** It shows the `errorText` with an error state. */
+	Error = "error"
+}
+
 /**
  * [See demo](../../?path=/story/inline-loading--basic)
  *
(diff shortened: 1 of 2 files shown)


src/inline-loading/inline-loading.component.ts
```diff
@@ -14,60 +27,88 @@ import {
...
 	`
 })
 export class InlineLoading {
+	InlineLoadingState = InlineLoadingState;
+
 	/**
 	 * Specify the text description for the loading state.
 	 */
-	@Input() loadingText;
+	@Input() state: InlineLoadingState = InlineLoadingState.Active;
```
cal-smith, line 73: I'm good with this as long as `[state]="'active'"` doesn't cause Angular to complain when it typechecks the templates 👍

stanislavgeorgiev: I haven't tested that scenario but why would somebody use the `string` literal instead of the `enum`?

cal-smith: In the case that they don't bind the `enum` to the component class ... It's an extra step that can trip people up, so we've generally avoided it in favour of string type unions

stanislavgeorgiev:
Thanks for merging this guys!
I think I didn't use the right commit message - used `feature` instead of `feat`. Can that be why the merge didn't trigger a new release?

cal-smith:
That would probably do it :+1: no worries!

PR #1288, opened 2020-05-20
Tag Filter close icon requires a class selector for latest Carbon

sjbeatle:
Added `bx--tag__close-icon` class to `svg` element to align w/latest from Carbon: https://github.com/carbon-design-system/carbon/blob/09348e65d26e32a92dca4e60a09caff87843575d/packages/components/src/components/tag/_tag.scss#L156

The selector `bx--tag--filter>svg` is no longer present in the latest from Carbon. They opted for using a class selector on the `svg` element.

#### Note
The latest docs from Carbon don't seem to reflect this change... I'm not sure if that is because the docs haven't yet been updated, but they should. For example, the tag filter section here no longer appears to be correct (unless I'm missing something): https://the-carbon-components.netlify.app/?nav=tag

I'm using:
    "carbon-components": "^10.10.3",
    "carbon-components-angular": "^4.1.6",

### DIFF
--- a/src/tag/tag-filter.component.ts
+++ b/src/tag/tag-filter.component.ts
@@ -14,6 +14,7 @@ import { Tag } from "./tag.component";
 			(click)="close.emit()"
 			focusable="false"
 			preserveAspectRatio="xMidYMid meet"
+			class="bx--tag__close-icon"
 			style="will-change: transform;"
 			xmlns="http://www.w3.org/2000/svg"
 			aria-label="Clear filter"


sjbeatle:
Selector in action: https://5ec5a613859eaf0008c3752b--carbon-components-angular.netlify.app/?path=/story/components-tag--filter
Proper hover states invoked now too :)
image

sjbeatle:
> Thanks for catching this! This is 100% a regression in `carbon-components` but we can work with this!

Very cool. Please let me know if there's more I could've done or should've done differently. I want to make sure I contribute in the appropriate manner :)

cal-smith:
> > Thanks for catching this! This is 100% a regression in `carbon-components` but we can work with this!
> 
> Very cool. Please let me know if there's more I could've done or should've done differently. I want to make sure I contribute in the appropriate manner :)

This is perfect 👍 If you wanted you could _also_ open an issue on the `carbon-components` side, but we'd want the fix here anyway since this would still break for most consumers

PR #1289, opened 2020-05-21
feat(inline-loading): Added `state` property and error state style.

stanislavgeorgiev:
Added handling for the `state` property as a string, following the comment in https://github.com/IBM/carbon-components-angular/pull/1286
Added unit tests.


stanislavgeorgiev:
I think you were on point with merging the previous one @cal-smith, I was impressed by you and @Donisius, accepting the item, reviewing and merging it in just a few short hours. I saw you published quite a lot yesterday which makes your quick response even more admirable. :)

stanislavgeorgiev:
You're awesome guys ;)

PR #1291, opened 2020-05-25
fix: change lint command and fix lint errors

Donisius:
#### Changelog

**New**

* Added a `lint:fix` command

**Changed**

* Changed the `lint` command so it can pick up on all lint errors


cal-smith:
Looks like there's some non-compliant files again @Donisius

PR #1294, opened 2020-05-26
fix(file-uploader): add disabling feature

### LINKED ISSUE #1279: File uploader needs an option to disable

khizar-s:
Closes #1279


Donisius:
> We should probably make sure this supports disabling drag and drop as well ... If there's no styles or designs for that on the Carbon side we can take this as is, and open an issue there to add that

Does adding the disabled property on the input achieves that @khizar-s ?

khizar-s:
> > We should probably make sure this supports disabling drag and drop as well ... If there's no styles or designs for that on the Carbon side we can take this as is, and open an issue there to add that
> 
> Does adding the disabled property on the input achieve that @khizar-s ?

If we're looking at just the basic button, adding the disabled property to the input is pointless since we can duplicate this disabled behavior by adding the property to the button. I left the property in input in case there were future changes for drag and drop that would allow us to disable it too

PR #1299, opened 2020-06-01
fix(modal): Modal close icon should be 20x20

yinglejia:

Modal close icon should be 20x20 according to Carbon design.

https://github.com/carbon-design-system/carbon/blob/6db63e027eeccc30d8531bf76d1983ae035767c3/packages/components/src/components/modal/_modal.scss#L342

#### Changelog

**New**

* {{new thing}}

**Changed**

* Change icon size to 20 and use svg directive instead of component.

**Removed**

* {{removed thing}}


cal-smith:
Thanks for catching and fixing this!

Issue #1312, opened 2020-06-05
DatePicker: Not disabling dates out of range after a selection is made


virkt25:
## Detailed description

> Describe in detail the issue you're having.

- No dates selected / set in the date picker:
  - You can see which dates are allowed to be selected as expected and user can select them
  
![image](https://user-images.githubusercontent.com/3311536/83923979-6b250800-a751-11ea-90fc-25103b55912d.png)

  - The moment a single selection is made the `disabled` date options look enabled even though the component doesn't let you select them. This is very confusing for a user who no longer knows what the valid range for selection is anymore.

![image](https://user-images.githubusercontent.com/3311536/83924070-a7f0ff00-a751-11ea-8eb1-d0403280ff0a.png)

> Is this a feature request (new component, new icon), a bug, or a general issue?

- Bug

> Is this issue related to a specific component?

- Date Picker

> What did you expect to happen? What happened instead? What would you like to see changed?

- Expect the disabled dates to remain disabled when the user is making a selection, has made a selection or is attempting to change their selection.

> What browser are you working in?

- Chrome

> What version of the Carbon Design System are you using?

- Latest. See Reproduction

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

- CS360 + others. We are opting to not use this component for the time being but would've liked to.

## Steps to reproduce the issue

- https://github.com/virkt25/carbon-ng-demo

virkt25:
Also for the date range picker I feel like the behaviour needs to be modified to be more intuitive for users. 

If a user is trying to make a new selection they should be forced to select both options again. A good example of this can be seen in this demo (which is also using flatpickr). https://mattlewis92.github.io/angularx-flatpickr/

cal-smith:
There may be some changes in the carbon flatpickr plugins that we haven't picked up yet ... It would be ideal if those were exposed as a package 🤔

virkt25:
Anything preventing the update to the latest version of flatpickr / associated plugins?

cal-smith:
Unsure ... That's something we definitely should look into though. The Carbon plugins aren't exposed as a package yet (IIRC), so I may take a stab at that first and try integrating them here.

Donisius:
I think this may been resolved in the latest version, does the error still come up on v4.10.1 @virkt25?

zvonimirfras:
Per Doni's comment, it looks like it's been resolved in one of the later versions. Feel free to reopen with more info if needed.

PR #1316, opened 2020-06-08
fix(datepicker): ensure bx classes are added on date picker creation

virkt25:
Closes IBM/carbon-components-angular #1312 

Using the `onDayCreate` hook to add the class fixes dates not being disabled as they should be after the date picker is opened for a re-selection.

For future I think the existing code adding the class can be removed. I wasn't sure of the difference between `daysContainer` and `dayContainer` to be able to remove it myself. 

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -202,7 +202,7 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked, After
 			}
 		},
 		onDayCreate: (_dObj, _dStr, _fp, dayElem) => {
-			dayElem.className += " bx--date-picker__day";
+			dayElem.classList.add("bx--date-picker__day");
 		},
 		value: this.value
 	};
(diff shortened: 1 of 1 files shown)


src/datepicker/datepicker.component.ts
```diff
@@ -201,6 +201,9 @@ export class DatePicker implements OnDestroy, OnChanges, AfterViewChecked, After
 				}
 			}
 		},
+		onDayCreate: (_dObj, _dStr, _fp, dayElem) => {
+			dayElem.className += " bx--date-picker__day";
```
cal-smith, line 205: We should probably use `classList` with a `contains` check to make sure we don't end up adding the class too many times - https://developer.mozilla.org/en-US/docs/Web/API/Element/classList ... I know flatpickr does some weird things with reusing elements, I'm not sure if this would be the same.

theiliad: you can just do `classList.add()` and it will skip the add if the className already exists

cal-smith: Oh sweet, that definitely simplifies that

virkt25:
Thanks @cal-smith @theiliad for the suggestion! Made the change :)

PR #1319, opened 2020-06-09
[Enhancement]: Tag Filter refactor

sjbeatle:
All changes are non-breaking enhancements to make Tag Filters more robust and inline with the React version and keyboard accessible.

#### Changelog

**New**

* HostBinding `id` attribute (setting to a unique value if one is not passed)
* HostBinding `aria-label` dynamically set by `ng-content` and `title`
* Input `tagId` to associate icon label with host element
* Input `title` for close icon label
* Input `id` for host/close icon
* Input `disabled` for close icon
* icon is now a `button` for proper keyboard access and `focus`/`hover` styles
  * `button` now owns the `(click)` event emitter
  * `aria-labelledby` bound to new input `tagId`
  *  `disabled` bound to new input
  *  `title` bound to new input

**Changed**

* `ng-content` is now wrapped in a `span` with the `bx--tag__label` class and a `title` attribute
* `svg` is now a Carbon Icon wrapped in a `button` with all relevant attributes moved to the `button`

**Removed**
* `svg`, using Carbon Icon instead


src/tag/tag-filter.component.ts
```diff
@@ -2,37 +2,59 @@ import {
...
 
 @Component({
 	selector: "ibm-tag-filter",
 	template: `
-		<ng-content></ng-content>
-		<svg
-			(click)="close.emit()"
-			focusable="false"
-			preserveAspectRatio="xMidYMid meet"
+		<span #label class="bx--tag__label" [title]="label.innerHTML">
```
cal-smith, line 15: We may want to use `textContent` for the title, since `innerHTML` may end up returning some weird HTML/SVG content ... We may also want to expose that as an Input for better control

src/tag/tag-filter.component.ts
```diff
@@ -2,37 +2,59 @@ import {
...
+			<span class="bx--visually-hidden">{{title}}</span>
+			<svg ibmIconClose size="16"></svg>
+		</button>
 	`
 })
 export class TagFilter extends Tag {
+	static tagFilterCount = 0;
+	@Input() tagId = `tag-filter-${TagFilter.tagFilterCount}`;
+	@Input() title = "Clear Filter";
+	@Input() id: string;
```
cal-smith, line 33: We should be able to use one input for all our ID requirements ... If we need a separate input to control the `aria-labelledby` it should be `ariaLabelledBy`, and in this case point to the default tag id ... If we're supporting labelledby we may also want to support `aria-label` as an alternative

src/tag/tag-filter.component.ts
```diff
@@ -2,37 +2,59 @@ import {
...
+			[title]="title">
+			<span class="bx--visually-hidden">{{title}}</span>
+			<svg ibmIconClose size="16"></svg>
+		</button>
 	`
 })
 export class TagFilter extends Tag {
+	static tagFilterCount = 0;
+	@Input() tagId = `tag-filter-${TagFilter.tagFilterCount}`;
+	@Input() title = "Clear Filter";
```
cal-smith, line 32: Is this the title for the whole tag or for the tag close button?

src/tag/tag-filter.component.ts
```diff
@@ -2,37 +2,59 @@ import {
...
 		return `bx--tag bx--tag--filter bx--tag--${this.type} ${this.class}`;
 	}
+
+	@HostBinding("attr.id") get attrId() {
+		return this.id ? this.id : this.tagId;
+	}
+
+	@HostBinding("attr.aria-label") get attrAriaLabel() {
+		const label = this.label.nativeElement.innerHTML;
+		return `${this.title} ${label}`;
```
cal-smith, line 53: This may end up adding a lot of HTML junk to the `aria-label` ... I do wonder if this is something we should provide? Maybe it'll be clearer when we sort out the title and label situation

src/tag/tag-filter.component.ts
```diff
@@ -2,37 +2,59 @@ import {
...
 	@Output() close = new EventEmitter<any>();
 
+	@ViewChild("label") label: ElementRef;
+
 	@HostBinding("attr.class") get attrClass() {
 		return `bx--tag bx--tag--filter bx--tag--${this.type} ${this.class}`;
 	}
+
+	@HostBinding("attr.id") get attrId() {
+		return this.id ? this.id : this.tagId;
```
cal-smith, line 48: This may be redundant if we can move to a single `id` input

src/tag/tag-filter.component.ts
sjbeatle: Hey! Yes, I'm open to your preference/suggestion. I was trying to find an analog for how the React library is doing it and this was the best I came up with, lol.

My preference is for it to be an Input. Would we keep `<ng-content />` in that case, to allow for complex content, or would we replace it with the Input's value?

src/tag/tag-filter.component.ts
sjbeatle: Tag close button... I'll make it more descriptive.

src/tag/tag-filter.component.ts
sjbeatle: Precisely, if we change the title and label situation to an Input, this will be much cleaner.

src/tag/tag-filter.component.ts
sjbeatle: > Thanks for taking this on! It's definitely much needed ... Just a few comments/questions 😄

I appreciate the feedback! I was _trying_, to some extent, to keep in line with a similar approach they're using in the React library, BUT I think I'll take a stab at updating this to be more in line w/what I'd want as a developer and that'll probably address most, if not all, of the concerns raised here. 

Thanks for the review!! 🎉

sjbeatle:
> Thanks for taking this on! It's definitely much needed ... Just a few comments/questions 😄

@cal-smith I appreciate the feedback! I was _trying_, to some extent, to keep in line with a similar approach they're using in the React library, BUT I think I'll take a stab at updating this to be more in line w/what I'd want as a developer and that'll probably address most, if not all, of the concerns raised here. 

Thanks for the review!! 🎉

src/tag/tag-filter.component.ts
sjbeatle: I'm just realizing that for now we have to keep `<ng-content />` or else it's a breaking change... BUT, I would recommend discussing w/the team whether in future versions you want to allow more complex tag-filter content than a simple title/string.

sjbeatle: Now exposed as an input: `title`

src/tag/tag-filter.component.ts
sjbeatle: I actually felt the aria labelling and referencing was unnecessarily complicated. I've reduced it to what I believe is minimally needed for A11y considerations.

sjbeatle:
@cal-smith Updates have been made. It is much simpler now and meets A11y/Keyboard requirements. DAP throws no violations and Voiceover reads the component very nicely. 👍🏼

sjbeatle:
> Nice work 🎉 API is definitely the biggest area where we differ from react - having a great developer experience for each framework is super important!
> 
> Thanks again for taking this on 😁

Let's get cracking on an HTML Web Component library and get rid of the React/Vue/Angular variants 😎😉

sjbeatle:
> Nice!! Thanks for taking this on! 🎉

My pleasure. I really enjoy this type of work. :)

cal-smith:
> > Nice work 🎉 API is definitely the biggest area where we differ from react - having a great developer experience for each framework is super important!
> > Thanks again for taking this on 😁
> 
> Let's get cracking on an HTML Web Component library and get rid of the React/Vue/Angular variants 😎😉

Funny you should mention that ... https://github.com/carbon-design-system/carbon-custom-elements

sjbeatle:
> Funny you should mention that ... https://github.com/carbon-design-system/carbon-custom-elements
😍

PR #1330, opened 2020-06-10
feat(tabs): add title to tab headers

### LINKED ISSUE #1125: Tabs should include title attribute
The anchor links in the tabs component should include title attribute.

This will allow the user to see full text when the text of the tab overflows and has ellipses.

Example: LongTabTex...

```
<a class="bx--tabs__nav-link" title="ANCHOR_TEXT" other-attributes>ANCHOR_TEXT</a>
```

Donisius:
Closes #1125

### DIFF
--- a/src/tabs/tab-header.component.ts
+++ b/src/tabs/tab-header.component.ts
@@ -3,7 +3,8 @@ import {
 	Input,
 	ViewChild,
 	ElementRef,
-	Output
+	Output,
+	AfterViewInit
 } from "@angular/core";
 
 import { Tab } from "./tab.component";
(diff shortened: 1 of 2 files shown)


src/tabs/tab-header.component.ts
```diff
@@ -80,6 +82,11 @@ export class TabHeader {
 	@ViewChild("tabItem", { static: false }) tabItem: ElementRef;
 
 	protected _cacheActive = false;
+	protected title = "Tab header";
```
cal-smith, line 85: We should probably expose this as an `Input`, since the actual `title` may not always be 1-1 with the `textContent`

PR #1331, opened 2020-06-10
fix(select): add guard before setting select value

### LINKED ISSUE #1318: <ibm-select> is not working in angular reactive forms
I am using <ibm-select> with Angular reactive forms.
When page is loaded, unless click on the select box, other fields are not loading and getting the following error

image
image
image
Only on click of the select box, other elements are populated. Also select box option value is coming as null. When I tried with html select I am able to get the selected value

<ibm-select
	[theme]="dark"
    formControlName="sectorName">
        <option value="" disabled selected hidden>Select Sector name</option>
        <option *ngFor="let item of items" value="item">{{ item }}</option>
</ibm-select>

Donisius:
Closes #1318

### DIFF
--- a/src/select/select.stories.ts
+++ b/src/select/select.stories.ts
@@ -9,10 +9,55 @@ import {
 
 import { SelectModule } from "../";
 import { DocumentationModule } from "../documentation-component/documentation.module";
+import { Component, OnInit } from "@angular/core";
+import {
+	FormGroup,
+	FormBuilder,
+	FormControl,
+	ReactiveFormsModule
+} from "@angular/forms";
+
+@Component({
+	selector: "app-reactive-form",
+	template: `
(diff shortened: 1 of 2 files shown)


src/select/select.stories.ts
```diff
@@ -9,10 +9,55 @@ import {
...
+	FormBuilder,
+	FormControl,
+	ReactiveFormsModule
+} from "@angular/forms";
+
+@Component({
+	selector: "app-reactive-form",
+	template: `
+		<form [formGroup]="formGroup">
+			<ibm-select formControlName="selecterino">
```
theiliad, line 24: `selecterino`?

cal-smith: I dig it

PR #1337, opened 2020-06-11
fix(checkbox): prevent setting id from disabling checkbox input

### LINKED ISSUE #1169: Adding an id on checkbox prevents it from being checked/unchecked

Donisius:
Closes #1169


src/checkbox/checkbox.stories.ts
```diff
@@ -17,6 +17,7 @@ storiesOf("Components|Checkbox", module).addDecorator(
 			<legend class="bx--label">{{label}}</legend>
 			<ibm-checkbox
 				checked="true"
+				id="This didn't work before"
```
theiliad, line 20: are we keeping this in?

Donisius:
```suggestion
```

PR #1339, opened 2020-06-11
fix(combobox): prevent search from emitting when items are replaced

### LINKED ISSUE #1338: Combo box closes and triggers search event with empty string when items changed
## Detailed description

When typing into the search field of a combo box, we have a listener that responds to the search event emitted by the combo box. In response we send a query to Infohub to get back a list of items and set the `items` property of the combo box to this returned list. Doing so causes the combo box to close and issue another search event with an empty string. This triggers another Infohub query without any filters applied, which gets back the original list of items, effectively negating the initial search. In general, we've observed that the search with empty string is triggered whenever the combo box's `items` property changes, which triggers an infinite loop in our scenario.

bug

combo box

Expected the combo box would remain open, displaying the updated list of items. Also, did not expect search event would be fired whenever the  `items` property of the combo box was updated.

angular carbon components 3.49.7

IBM UIHub Inventory Control Tower. Deadline is end of June 2020.

1. Create an angular component with a combo box defined in the *.component.html:
```
<ibm-combo-box
    appendInline="true"
    (selected)="selectedFilter($event)"
    [placeholder]="placeholder"
    [label]="label"
    [items]="filterItems"
    [invalid]="invalid"
    [invalidText]="invalidText"
    (search)="onSearch($event)"
  >
    <ibm-dropdown-list></ibm-dropdown-list>
  </ibm-combo-box>
```
2. Define an input in the *.component.ts for each of the input variables in the *.component.html file:
```
@Input()
  filterItems: ListItem[];

  @Input()
  placeholder: string;

  @Input()
  label: string;

  @Input()
  identifier: string;
```
3. Define an onSearch function in the *.component.ts file to update the `items` property of the combo box based on the entered text:
```
onSearch(event: string): void {
    console.log(`Search triggered with text: ${event}`);
    this.filterItems = getItemsBySearchText(event);
  }
```
4. Observe that even not entering any text will trigger the onSearch method to be called when the items in the combo box are first loaded and on every time they are set to a new value.

## Additional information

* Screenshots or code
* Notes

Donisius:
Closes #1338 

This PR adds an optional second parameter to the `onSearch` function of the combobox `shouldEmitSearch` which controls the emitting of the search event. `shouldEmitSearch` only needs to be set to false when onChanges repeats the search when a new item is introduced to the combobox to prevent an infinite loop if a function bound to the `search` event introduces a new item.

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -354,7 +354,7 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit {
 			this.view.items = changes.items.currentValue;
 			// If new items are added into the combobox while there is search input,
 			// repeat the search.
-			this.onSearch(this.input.nativeElement.value);
+			this.onSearch(this.input.nativeElement.value, false);
 			this.updateSelected();
 		}
 	}
@@ -509,8 +509,10 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit {
 	/**
 	 * Sets the list group filter, and manages single select item selection.
 	 */
-	public onSearch(searchString) {
-		this.search.emit(searchString);
+	public onSearch(searchString, shouldEmitSearch = true) {
+		if (shouldEmitSearch) {
+			this.search.emit(searchString);
+		}
 		if (searchString && this.type === "single") {
 			this.showClearButton = true;
 		} else {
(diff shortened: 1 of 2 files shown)


cal-smith:
Don't worry, just trying something

Donisius:
> This is definitely a good short term solution ... Once we get around to refactoring combobox we can make this whole situation a lot better
> 
> I noticed that clicking the clear button doesn't clear the search? There's a PR open for that right?

I can include it here too

PR #1365, opened 2020-06-23
WIP: Convert components to individual entry points 

cal-smith:


cal-smith:
guess I gotta open it again, since travis isn't picking it up 🤦

PR #1375, opened 2020-06-26
fix(build): Update datepicker I18nModule import

yinglejia:

Update datepicker I18nModule import to include index, which may solve a build issue for Angular 8 projects (so far reproduced in 3 Angular 8 projects).

`ERROR in Unexpected value 'undefined' imported by the module 'DatePickerModule in .../node_modules/carbon-components-angular/carbon-components-angular.d.ts'`

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}


cal-smith:
> the index shouldn't be needed, does it not resolve without it? also `./../` doesn't make sense

`index` is needed due to how the angular compiler resolves modules (afaik). `./../` had been our convention for a while, it's functionally equivalent to `../`

PR #1380, opened 2020-06-29
fix(ComboBox): implemented missing theme property

Licen-it:
Closes IBM/carbon-components-angular#1379

Implemented `theme` property for ComboBox in the same way it is implemented for Dropdown

#### Changelog

**New**

* `theme` property is now available and can assume values of `dark` or `light`

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -95,7 +95,7 @@ import { Observable } from "rxjs";
 				</div>
 				<input
 					#input
-					type="text"
+					role="combobox"
 					[disabled]="disabled"
 					(input)="onSearch($event.target.value)"
 					(keydown.enter)="onSubmit($event)"
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -95,6 +95,7 @@ import { Observable } from "rxjs";
 				</div>
 				<input
 					#input
+					type="text"
 					role="combobox"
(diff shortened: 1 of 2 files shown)


src/combobox/combobox.component.ts
```diff
@@ -94,6 +95,7 @@ import { Observable } from "rxjs";
 				</div>
 				<input
 					#input
+					type="text"
```
Donisius, line 98: Looking at the `carbon-components` css it seems that the background color of the input is targeted by `role='combobox'`:
```
.bx--list-box input[role='combobox'] {
  background-color: inherit;
  min-width: 0;
}
```
Not sure if that was from an earlier version or it was changed in a later version?

Licen-it: actually this is what i see using latest version:
image

Licen-it: and in react version of the combobox they use `type='text'`

Donisius: So I guess it targets both but the `type='text'` selector is only available in the latest version, do you think it would be better to use `role='combobox'` instead so that earlier versions can still take advantage of the light theme @cal-smith ?

cal-smith: We _should_ be able to use both I would think. `role="combobox"` is still valid (and needed) as a aria attribute, and we _should_ be specifying `type="text"` for all generic inputs

PR #1384, opened 2020-06-30
feat(notification): add support for links

cal-smith:
Closes IBM/carbon-components-angular#1240

Adds support for links in notifications as part of the `NotificationContent` interface


cal-smith:
@Donisius everything should be good now!

PR #1388, opened 2020-07-06
Ported to V4 - feat(inline-loading): Added `state` property and error state style. 

stanislavgeorgiev:

Cherry picked `InlineLoadingState` from v3 to v4

### DIFF
--- a/src/inline-loading/inline-loading.component.ts
+++ b/src/inline-loading/inline-loading.component.ts
@@ -42,20 +42,16 @@ export enum InlineLoadingState {
 			</div>
 			<svg
 				*ngIf="state === InlineLoadingState.Finished"
-				class="bx--inline-loading__checkmark-container bx--inline-loading__svg"
-				xmlns="http://www.w3.org/2000/svg"
-				viewBox="0 0 10 10">
-				<polyline class="bx--inline-loading__checkmark" points="0.74 3.4 3.67 6.34 9.24 0.74"></polyline>
+				ibmIconCheckmarkFilled
+				size="16"
+				class="bx--inline-loading__checkmark-container">
 			</svg>
 			<svg
 				*ngIf="state === InlineLoadingState.Error"
-				class="bx--inline-loading--error"
-				xmlns="http://www.w3.org/2000/svg"
-				viewBox="0 0 16 16">
-				<path d="M8,1C4.1,1,1,4.1,1,8s3.1,7,7,7s7-3.1,7-7S11.9,1,8,1z M10.7,11.5L4.5,5.3l0.8-0.8l6.2,6.2L10.7,11.5z"></path>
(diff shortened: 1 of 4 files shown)


src/inline-loading/inline-loading.component.ts
```diff
@@ -14,60 +27,88 @@ import {
...
 				<polyline class="bx--inline-loading__checkmark" points="0.74 3.4 3.67 6.34 9.24 0.74"></polyline>
 			</svg>
+			<svg
+				*ngIf="state === InlineLoadingState.Error"
+				class="bx--inline-loading--error"
+				xmlns="http://www.w3.org/2000/svg"
+				viewBox="0 0 16 16">
+				<path d="M8,1C4.1,1,1,4.1,1,8s3.1,7,7,7s7-3.1,7-7S11.9,1,8,1z M10.7,11.5L4.5,5.3l0.8-0.8l6.2,6.2L10.7,11.5z"></path>
+				<path fill="none" d="M10.7,11.5L4.5,5.3l0.8-0.8l6.2,6.2L10.7,11.5z" data-icon-path="inner-path" opacity="0"></path>
+			</svg>
```
Donisius, line 57: We should probably use the directive version of the icon, I believe it's `ibmIconErrorFilled` similar to how it's used in https://github.com/IBM/carbon-components-angular/blob/master/src/notification/notification.component.ts#L27-L32

src/inline-loading/inline-loading.component.ts
```diff
@@ -14,60 +27,88 @@ import {
...
 				</svg>
 			</div>
 			<svg
-				*ngIf="success === true"
+				*ngIf="state === InlineLoadingState.Finished"
 				class="bx--inline-loading__checkmark-container bx--inline-loading__svg"
 				xmlns="http://www.w3.org/2000/svg"
 				viewBox="0 0 10 10">
 				<polyline class="bx--inline-loading__checkmark" points="0.74 3.4 3.67 6.34 9.24 0.74"></polyline>
 			</svg>
```
Donisius, line 49: while we're here we should probably change this to use the directive as well, like the one in https://github.com/IBM/carbon-components-angular/blob/master/src/notification/notification.component.ts#L39-L44

src/inline-loading/inline-loading.component.ts
stanislavgeorgiev: Good point!

src/inline-loading/inline-loading.component.ts
stanislavgeorgiev: Certainly :)
No code duplication is allowed! :)

stanislavgeorgiev:
Thanks for the notes @Donisius, those updates are certainly a must do in my opinion and not nitpicks :)

PR #1402, opened 2020-07-11
[Updates: non-breaking] Select component

sjbeatle:
Now fully aligned with Carbon's documentation: https://the-carbon-components.netlify.app/?nav=select

* Correct class name for `optgroup` directive
* All variants considered.
* Better skeleton handling (label skeleton wasn't previously appearing because it was "below" the select skeleton)
* Now projecting the `select` into an additional wrapper when display property is set to "inline" (as according to Carbon specs)
* Now placing help text in correct position when display property is set to "inline"
* Unneeded inline style (width: 100%) removed. (wrapping div is already 100% by default, element's display is flex, and `.bx--select` CSS is dictating width of the `select` element; so, I believe this inline style to be superfluous)
* `invalid` and `disabled` selectors added based on their respective prop values
* `aria-invalid` attribute added when `invalid` prop is true
* The invalid text has role="alert" and aria-live="polite" added for screenreader support

**NOTE**
The `[data-invalid] ~ .bx--select__arrow` style block was removed. It isn't necessary since Carbon's CSS handles everything properly, but ultimately it wasn't doing anything in the first place because `.bx--select__arrow` was never a subsequent sibling of `[data-invalid`, it was a child element. Therefore no elements on the page ever matched that selector anyway.

**Final Commentary**
None of these changes are breaking changes with the extremely rare exception of the removal of the inline style `width: 100%` that was being applied to the `.bx--select` element. From all I can tell, that style definition was superfluous... **BUT** if there is any concern or reason to keep it, I'll happily put it back. Otherwise, these fixes are non-breaking and able to be included in the latest version of your library.

Any questions, please don't hesitate to ask. Thanks!


cal-smith (dismissed):
This is excellent! Nice work 😁 

I believe the `width: 100%;` was working around a bug in a previous version of `carbon-components` I expect that we'll be fine to remove that at this point 👍 

Thanks for the super detailed PR, very much appreciated 🎉

cal-smith:
Looks like dependabot added a merge conflict in the `package-lock.json` though 🤦

theiliad (dismissed):

sjbeatle:
> Looks like dependabot added a merge conflict in the `package-lock.json` though 🤦

Is there something I need to do? (ie, revert the package-lock change)
Or, is this something on your end to deal w/dependabot? (I'm not familiar w/dependabot)

cal-smith:
> Is there something I need to do? (ie, revert the package-lock change)
> Or, is this something on your end to deal w/dependabot? (I'm not familiar w/dependabot)

Ah yeah, you'll just need to pull and resolve the conflict ... running `npm i` should auto-resolve the `package-lock.json` conflicts though :+1:

jdoyle65:
Waiting on this for the fix to `<ibm-select>` label skeleton positioning. Has this PR been abandoned?

sjbeatle:
> Waiting on this for the fix to `<ibm-select>` label skeleton positioning. Has this PR been abandoned?

I hope not, lol. I never did get to resolving any conflicts, but I can if these changes haven't already been implemented elsewhere.

sjbeatle:
Just resolved the conflicts. Please consider still merging this. I think these are good changes...

I made sure to retain the new `size` Input logic. That was the only change since my original PR, other than the package-lock file, which should now be fixed, as well.

cal-smith (approved):
Awsome!

cal-smith:
Will definitely be merging this! Thanks for the update @sjbeatle

sjbeatle:
> Will definitely be merging this! Thanks for the update @sjbeatle

Fantastic!! Very exciting :)

I haven't been working on a Carbon project as of late, otherwise I'd contribute more!

PR #1425, opened 2020-07-24
fix(Combobox): added missing onBlur calls in template

Licen-it:
Closes IBM/carbon-components-angular#1424

`onBlur` method is defined, but never used. Just registered it where needed so that the reactive from control can be marked as touched when necessary.

#### Changelog

**Changed**

* `onBlur` is now called from some template elements

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -75,6 +75,7 @@ import { DropdownService } from "../dropdown/index";
 				[id]="id">
 				<button
 					*ngIf="type === 'multi' && pills.length > 0"
+					type="button"
 					(click)="clearSelected()"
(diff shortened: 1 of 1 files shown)


src/combobox/combobox.component.ts
```diff
@@ -71,14 +71,14 @@ import { DropdownService } from "../dropdown/index";
 				[attr.aria-label]="label"
 				aria-haspopup="true"
 				(click)="toggleDropdown()"
+				(blur)="onBlur()"
 				[id]="id">
-				<div
+				<button
```
cal-smith, line 76: If we're swapping this to `button` we'll need to add `type="button"` as well 👍

PR #1450, opened 2020-08-08
Ng v10

arindamdawn:
## Updated to Angular v10

This PR updates Angular to its latest version (v10+) along with updates to some other dependencies.

#### Changelog

**New**

 None

**Changed**

* Angular version updated to 10.0.8 (latest)

**Removed**

None


arindamdawn:
> 
> 
> Hey 👋 Thanks for taking a look at this. We generally try to maintain backwards compatibility for a few Angular versions, which usually means staying on an older version of Angular and bumping our `peerDependencies` ... We also use `peerDependencies` to indicate which major versions of Angular we're compatible with and which packages we require.
> 
> The test and build is also failing.
> 
> A good first step would be testing an Angular 10 app with the latest version of v4 - if there's no issues with that configuration we should be good to just bump the `peerDependencies`.

Sure @cal-smith I am closing this now. Will check it with the latest version of Angular and then update the peer-dependencies if everything works well :)

PR #1457, opened 2020-08-12
Add inline variant for links

elycheea:
Adds the Carbon [inline link variant](https://carbondesignsystem.com/components/link/usage#inline-link) intended for links within sentences/text.

image

#### Changelog

**New**

* Adds `inline` input to link directive

### DIFF
--- a/src/link/link.directive.ts
+++ b/src/link/link.directive.ts
@@ -47,5 +47,19 @@ export class Link {
 		return this._disabled;
 	}
 
+	/**
+	 * Set to true to inline link.
+	 */
+	@Input()
+	@HostBinding("class.bx--link--inline")
+	set inline(inline: boolean) {
+		this._inline = inline;
+	}
+
+	get inline(): boolean {
+		return this._inline;
+	}
+
 	private _disabled;
+	private _inline;
 }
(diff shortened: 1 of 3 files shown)


src/link/link.directive.ts
```diff
@@ -47,5 +47,19 @@ export class Link {
 		return this._disabled;
 	}
 
+	/**
+	 * Set to true to inline link.
+	 */
+	@Input()
+	@HostBinding("class.bx--link--inline")
+	set inline(inline: boolean) {
+		this._inline = inline;
```
cal-smith, line 56: This can just be a property 👍 no need for it to have a getter/setter combo

src/link/link.spec.ts
```diff
@@ -16,6 +16,12 @@ class TestLinkComponent {
 class TestDisabledLinkComponent {
 }
 
+@Component({
+	template: `<a href="https://angular.carbondesignsystem.com/" [inline]="1+1===2" ibmLink>link</a>`
```
cal-smith, line 20: Not sure why we need the `1+1===2` there? I would recommend just passing `true`
```suggestion
	template: `<a href="https://angular.carbondesignsystem.com/" [inline]="true" ibmLink>link</a>`
```

elycheea: Was mostly just following the test structure already in place for disabled links! Would normally do this otherwise! Good to know for future reference though.

cal-smith: Ah, not sure how that got like that 😮

PR #1460, opened 2020-08-17
fix(dialog): prevent dialog from remaining open

### LINKED ISSUE #1459: Tooltip message not disappearing
## Detailed description

> In Firefox  & Chrome browser, after hovering over the tooltip a couple of times , the message freezes and when navigating , the message remains in the page. There is no definite pattern of this behaviour ( for e.g like hovering over 5 times/ 10 times ) . Its a random behaviour. 

Edit : Happening in Chrome too,  updated the description.

 

Bug

Tooltip

The tooltip message is supposed to disappear, but it did not.

Firefox & Chrome

     "@carbon/icons-angular": "^11.1.4",
    "carbon-components": "^10.13.0",
    "carbon-components-angular": "^4.10.2",

    Release is around September 2nd week.

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

* Screenshots or code
Screen Shot 2020-08-13 at 8 32 02 AM
Screen Shot 2020-08-13 at 8 32 27 AM
* Notes

Donisius:
Closes #1459

### DIFF
--- a/src/dialog/dialog.service.ts
+++ b/src/dialog/dialog.service.ts
@@ -78,11 +78,9 @@ export class DialogService {
 			dialogRef = viewContainer.createComponent(componentFactory, 0, this.injector);
 		} else if (!this.placeholderService.hasPlaceholderRef()) {
 			dialogRef = viewContainer.createComponent(componentFactory, 0, this.injector);
-			setTimeout(() => {
-				if (dialogRef) {
-					window.document.querySelector("body").appendChild(dialogRef.location.nativeElement);
-				}
-			});
+			if (dialogRef) {
+				window.document.querySelector("body").appendChild(dialogRef.location.nativeElement);
+			}
 		} else {
 			dialogRef = this.placeholderService.createComponent(componentFactory, this.injector);
 		}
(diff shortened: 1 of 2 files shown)


src/dialog/dialog.service.ts
```diff
@@ -78,11 +78,9 @@ export class DialogService {
 			dialogRef = viewContainer.createComponent(componentFactory, 0, this.injector);
 		} else if (!this.placeholderService.hasPlaceholderRef()) {
 			dialogRef = viewContainer.createComponent(componentFactory, 0, this.injector);
-			setTimeout(() => {
-				if (dialogRef) {
-					window.document.querySelector("body").appendChild(dialogRef.location.nativeElement);
-				}
-			});
+			if (dialogRef) {
```
cal-smith, line 81: Is there a reasonable explanation as to how this would stop the dialog from staying open?

Donisius: I'm still not 100% sure why, removing the set timeout prevents dialogs from staying open if I try and toggle them as fast as I can. Increasing the timeout makes it easier to reproduce the problem

cal-smith: Huh 🤔 It's possible this was a fix for something that isn't broken anymore

cal-smith:
@Donisius one of the tests are failing: https://travis-ci.com/github/IBM/carbon-components-angular/builds/181124961

src/dialog/dialog.service.ts
Donisius: Adding a `setTimeout` to the close also seems to prevent it from staying open... I think that's a safer approach to avoid any Expression changed errors which can potentially pop up

cal-smith: That sounds like a potentially better solution 👍

PR #1488, opened 2020-09-03
fix(tag-filter): prevent click propagation

### LINKED ISSUE #1485: Removing tag filter causes dialog to close unexpectedly
## Detailed description

I've extended the dialog component and dialog directive to create a custom dialog that includes tag filters and a number of checkboxes for toggling those tag filters. When I check/uncheck a checkbox, a tag filter is added to / removed from the array of tags in the dialog. I also have a function that responds to the `close` output of the tag filter by removing that tag from the array. Removing tags by clicking the clear/delete button on the tag filter causes the dialog to close, whereas removing tags via the checkboxes does not.

bug

tag filter and possibly dialog

I expected the dialog to remain open instead of closing. Removing tags from the array should not trigger dialog closing.

N/A

3.54.0

IBM Sterling UIHub. Need this resolved before end of September 2020.

[Codesandbox](https://codesandbox.io/s/carbon-components-angular-forked-sri29)

1. Add dialog directive to a button or other element.
2. Add tag filters to the dialog's template via ngFor and provide an array as input.
3. Define a function to listen to the `close` output of the tag filter and remove the tag in response.
4. Click on the button to open the dialog.
5. Click the x on one of the tag filters.

Donisius:
Closes #1485

### DIFF
--- a/src/dialog/tooltip/tooltip.stories.ts
+++ b/src/dialog/tooltip/tooltip.stories.ts
@@ -66,7 +66,6 @@ storiesOf("Components|Tooltip", module)
 							{{ tag.content }}
 						</ibm-tag-filter>
 					</ng-template>
-	
 					<div class="bx--tooltip__label">
 						{{triggerText}}
 						<span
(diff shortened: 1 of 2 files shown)


cal-smith (commented):
Can we still get clicks on the immediate element? If not we may need a `@Output` `click`/`onClick` to handle that usecase

Donisius:
> Can we still get clicks on the immediate element? If not we may need a `@Output` `click`/`onClick` to handle that usecase

It seems to 🤔  Unless I'm misunderstanding this seems to work
Screen Shot 2020-09-08 at 11 02 32 AM

cal-smith (approved):

PR #1490, opened 2020-09-03
fix(tile): fix accessibility issue with expandable tile

DusanMilko:
Aligned html and focus elements with carbon components js structure. 
Previously there were 2 focusable elements and we only need 1

Closes https://github.com/IBM/carbon-components-angular/issues/1491

#### Changelog

**Changed**

* Changed which elements are focusable


Donisius (approved):
Nice LGTM

cal-smith (approved):

PR #1495, opened 2020-09-04
Updated the toggle component to fix the accessibility issue for label…

maheshkanchetti:
… and checkbox

{{short description}}

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}

### DIFF
--- a/src/tabs/tab-header-group.component.ts
+++ b/src/tabs/tab-header-group.component.ts
@@ -35,7 +35,7 @@ import { Subscription } from "rxjs";
 			<a
 				href="javascript:void(0)"
 				class="bx--tabs-trigger-text"
-				tabindex="-1">
+				tabindex="-1" *ngIf="(!getSelectedTab().headingIsTemplate && getSelectedTab().heading != '') || getSelectedTab().headingIsTemplate">
 				<ng-container *ngIf="!getSelectedTab().headingIsTemplate">
 					{{ getSelectedTab().heading }}
 				</ng-container>
--- a/src/tabs/tab-header-group.component.ts
+++ b/src/tabs/tab-header-group.component.ts
@@ -35,7 +35,8 @@ import { Subscription } from "rxjs";
 			<a
 				href="javascript:void(0)"
 				class="bx--tabs-trigger-text"
-				tabindex="-1" *ngIf="(!getSelectedTab().headingIsTemplate && getSelectedTab().heading != '') || getSelectedTab().headingIsTemplate">
+				tabindex="-1"
+				*ngIf="(!getSelectedTab().headingIsTemplate && getSelectedTab().heading != '') || getSelectedTab().headingIsTemplate">
 				<ng-container *ngIf="!getSelectedTab().headingIsTemplate">
 					{{ getSelectedTab().heading }}
 				</ng-container>
--- a/src/toggle/toggle.component.ts
+++ b/src/toggle/toggle.component.ts
@@ -110,6 +110,16 @@ export class Toggle extends Checkbox {
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


src/toggle/toggle.component.ts
```diff
@@ -109,6 +109,16 @@ export class Toggle extends Checkbox {
 		return this._offValues.value;
 	}
 
+	/**
+	 * Sets the aria labelledby on the `toggle`s label element.
+	 */
+	@Input() ariaLabelledby: string;
```
Donisius, line 115: I believe the checkbox component has these inputs, does setting the inputs still work without having these?

src/ui-shell/header/header-action.component.ts
```diff
@@ -16,7 +16,6 @@ import {
 			[ngClass]="{
 				'bx--header__action--active': active
 			}"
-			[attr.aria-label]="title"
```
Donisius, line 19: Why does this need to be taken out?

src/ui-shell/sidenav/sidenav.component.ts
```diff
@@ -17,7 +17,7 @@ import { NavigationItem } from "../header/header-navigation-items.interface";
 @Component({
 	selector: "ibm-sidenav",
 	template: `
-		<nav class="bx--side-nav__navigation" role="navigation">
+		<nav class="bx--side-nav__navigation" role="navigation" [attr.aria-label]="ariaLabel">>
```
Donisius, line 20:
```suggestion
		<nav class="bx--side-nav__navigation" role="navigation" [attr.aria-label]="ariaLabel">
```

Donisius: Also should probably be on multiple lines

src/tabs/tab-header-group.component.ts
```diff
@@ -35,7 +35,7 @@ import { Subscription } from "rxjs";
 			<a
 				href="javascript:void(0)"
 				class="bx--tabs-trigger-text"
-				tabindex="-1">
+				tabindex="-1" *ngIf="(!getSelectedTab().headingIsTemplate && getSelectedTab().heading != '') || getSelectedTab().headingIsTemplate">
```
Donisius, line 38:
```suggestion
				tabindex="-1"
				*ngIf="(!getSelectedTab().headingIsTemplate && getSelectedTab().heading != '') || getSelectedTab().headingIsTemplate">
```

Donisius (changes requested):
Thanks for taking this on! Just a few things

src/ui-shell/header/header-action.component.ts
maheshkanchetti: Header element with label as banner was reported as violation

src/ui-shell/sidenav/sidenav.component.ts
maheshkanchetti: yes, thanks

src/ui-shell/header/header-action.component.ts
maheshkanchetti: Facing the accessibility issue for that label 
Accessible name does not match or contain the visible label text

src/toggle/toggle.component.ts
maheshkanchetti: Thanks, we have overlooked it.
We are removing them from toggle as we have them in checkbox component

cal-smith (approved):
Thanks for taking this on 👍 everything seems good to me

src/ui-shell/header/header-action.component.ts
Donisius: Okay, that makes sense to me

PR #1508, opened 2020-09-10
feat(table): Add size input for toolbar and add missing header classes

### LINKED ISSUE #1484: DataTable toolbar and action bar should have option for size="small" to match design toolkit
The Carbon Sketch design toolkit provides both "large" and "small" versions of action bar and tool bar. It would be nice if the DataTable tool bar and action bar components supported this.

image

In the datagrid component, we can control the table row height with the property size=small/short/normal/large but it seems we need to modify CSS to get the action bar and toolbar to match the smaller height that design toolkit provides.
I am wondering if we can have at least option between small and normal heights of tool bar and action bar?

**Summary**
Request to development: a "small" property for datagrid tool bar and action bar so we can match what is provided in design kit.

**Justification**
These are elements that designers are designing with. If each dev team doesn't have to create css to do this for themselves we save effort!! :)

Donisius:
Closes #1484

### DIFF
--- a/src/dialog/overflow-menu/overflow-menu.component.ts
+++ b/src/dialog/overflow-menu/overflow-menu.component.ts
@@ -32,6 +32,7 @@ import { OverflowMenuDirective } from "./overflow-menu.directive";
 		<div
 			[ibmOverflowMenu]="options"
 			[ngClass]="{'bx--overflow-menu--open': open}"
+			[className]="'bx--overflow-menu ' + triggerClass"
 			[attr.aria-label]="buttonLabel"
 			[flip]="flip"
 			[isOpen]="open"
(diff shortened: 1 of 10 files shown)


src/dialog/overflow-menu/overflow-menu.component.ts
```diff
@@ -32,6 +32,7 @@ import { OverflowMenuDirective } from "./overflow-menu.directive";
 		<div
 			[ibmOverflowMenu]="options"
 			[ngClass]="{'bx--overflow-menu--open': open}"
+			[className]="'bx--overflow-menu ' + triggerClass"
```
cal-smith, line 35: The `bx--overflow-menu` should go in the normal `class` attribute, or the `[ngClass]` ... Then we can just pass `triggerClass` to the `[className]`

cal-smith (changes requested):
Looks good overall, just one nitpick

src/dialog/overflow-menu/overflow-menu.component.ts
Donisius: for some reason that creates the property `triggerClass` on the component and sets it to the value of trigger input which causes the icon to be misplaced. An alternative that works and I think looks better is `class="bx--overflow-menu {{triggerClass}}"`

cal-smith: Odd 🤔 Yeah I think I prefer `class="bx--overflow-menu {{triggerClass}}"` overall

cal-smith (approved):

PR #1511, opened 2020-09-10
fix(accessibility): fix interactive search element not being focusabl…

DusanMilko:
Closes: https://github.com/IBM/carbon-components-angular/issues/1510

Fixing accessibility issue for interactive elements

1. The magnify glass in search is clickable so it should be focusable. This causes issue in table toolbar.

2. Some table rows have a host click listener, because rows are missing a tabindex they are not focusable by keyboard

### DIFF
--- a/src/table/body/table-row.component.ts
+++ b/src/table/body/table-row.component.ts
@@ -183,6 +183,10 @@ export class TableRowComponent {
 		return this.expandable ? true : null;
 	}
 
+	@HostBinding("attr.tabindex") get isAccessible() {
+		return this.enableSingleSelect && !this.showSelectionColumn ? 0 : null;
+	}
+
 	protected _checkboxLabel = this.i18n.getOverridable("TABLE.CHECKBOX_ROW");
 	protected _expandButtonAriaLabel = this.i18n.getOverridable("TABLE.EXPAND_BUTTON");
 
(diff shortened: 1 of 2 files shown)


src/table/body/table-row.component.ts
```diff
@@ -183,6 +183,10 @@ export class TableRowComponent {
 		return this.expandable ? true : null;
 	}
 
+	@HostBinding("attr.tabindex") get isAccessible() {
+		return this.enableSingleSelect && !this.showSelectionColumn ? 0 : null;
```
Donisius, line 187: I seem to still be able to focus on table rows when `enableSingleSelect` is set to `true` and `showSelectionColumn` is set to `false`. Is there any other condition I'm missing that makes a row not focusable?

DusanMilko: The situation you explained is the only situation that makes it true and focusable. Can you better explain what you mean? 

Possible senarios
enableSingleSelect = true + showSelectionColumn = false => 0 (focusable)
enableSingleSelect = false + showSelectionColumn = false => null (not focusable)
enableSingleSelect = true + showSelectionColumn = true => null (not focusable)
enableSingleSelect = false + showSelectionColumn = true => null (not focusable)

cal-smith: @Donisius I'm also a bit confused ... Do you mean you see that behavior in the current version?

Otherwise I think we're all getting at the same thing, so we should be good to go.

Donisius: @DusanMilko @cal-smith Apologies, I thought I was able to focus on the rows in the current version so I was wondering why this was necessary.

Donisius (approved):
Looks good to me!

src/table/body/table-row.component.ts
cal-smith: No worries!

cal-smith (approved):

PR #1514, opened 2020-09-11
fix(ui-shell): more flexible icon projection

cal-smith:
Closes IBM/carbon-components-angular#925

We can't project all the icon components (There's no way to select `ibm-icon-*`), but we can project SVGs in addition to components or SVGs with an `icon` attribute.


src/ui-shell/sidenav/sidenav-item.component.ts
```diff
@@ -27,7 +27,7 @@ import { Router } from "@angular/router";
 				[attr.aria-current]="(active ? 'page' : null)"
 				(click)="navigate($event)">
 				<div *ngIf="!isSubMenu" class="bx--side-nav__icon">
-					<ng-content select="[icon]"></ng-content>
+					<ng-content select="svg, [icon]"></ng-content>
```
Donisius, line 30: Thats pretty neat 😮

Donisius (approved):

src/ui-shell/sidenav/sidenav-item.component.ts
cal-smith: Yep ✨ just a standard css selector!

PR #1544, opened 2020-09-29
fix(error): fix error not visually appearing on dropdown and combobox

DusanMilko:
Fixing error text not appearing for dropdown and combobox.
This makes error text handling the same across all components

Should fix: https://github.com/IBM/carbon-components-angular/issues/1446

#### Changelog

**Changed**

* Class location for error text in dropdown and combobox


cal-smith (approved):
Excellent 👍 Thanks!

PR #1545, opened 2020-09-30
fix(accessibility): fix accessibility violation on disabled element

DusanMilko:
Disabled button in tab was failing accessibility for contrast color.
Recommendation is for disabled element to have aria-disabled

fixes: https://github.com/IBM/carbon-components-angular/issues/1546

Before:
Screen Shot 2020-09-30 at 9 03 42 AM

After:
Screen Shot 2020-09-30 at 9 03 26 AM


theiliad (approved):

cal-smith (approved):
Great catch! Thanks! 😁

PR #1550, opened 2020-10-01
feat(accordion): adding align input to accordion

DusanMilko:
Add align input so user can select end or start alignment. This is from the carbon docs:

Fixes: https://github.com/IBM/carbon-components-angular/issues/1549

Screen Shot 2020-10-01 at 12 00 21 PM


theiliad (approved):

cal-smith (approved):
Nice!

PR #1551, opened 2020-10-01
fix(helper): fix helper text location select, dropdown, combobox, input

DusanMilko:
Fix helper text location for select, input, dropdown and combobox

before:
Screen Shot 2020-10-01 at 12 18 17 PM

after:
Screen Shot 2020-10-01 at 12 18 30 PM


cal-smith (approved):
Nice work 👍 Thanks for bringing us back inline here.

PR #1558, opened 2020-10-05
fix(dialog): fix dialog position or overflow menu

DusanMilko:
Open to feedback and discussion.
I know you can pass an offset to the component but it seems wrong for the component to break by default because some parent has some position set with some modifying positions. It should work as expected by default.

Fixes: https://github.com/IBM/carbon-components-angular/issues/1557

I've even tested this with multiple nesting elements: https://codepen.io/dusanmilko/pen/pobzbqR

Before:
Screen Shot 2020-10-05 at 10 50 29 AM

After:
Screen Shot 2020-10-05 at 10 50 40 AM

Before:
Screen Shot 2020-10-05 at 10 51 36 AM

After:
Screen Shot 2020-10-05 at 10 51 22 AM

### DIFF
--- a/src/dialog/overflow-menu/overflow-menu-pane.component.ts
+++ b/src/dialog/overflow-menu/overflow-menu-pane.component.ts
@@ -54,7 +54,7 @@ export class OverflowMenuPane extends Dialog implements AfterViewInit {
 			* we also move the element by half of it's own width, since
 			* position service will try and center everything
 			*/
-			const closestRel = this.closestAttr('position', ['relative', 'fixed', 'absolute']);
+			const closestRel = this.closestAttr("position", ["relative", "fixed", "absolute"]);
 			const topFix = closestRel ? closestRel.getBoundingClientRect().top * -1 : 0;
 			const leftFix = closestRel ? closestRel.getBoundingClientRect().left * -1 : 0;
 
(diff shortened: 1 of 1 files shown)


src/dialog/overflow-menu/overflow-menu-pane.component.ts
```diff
@@ -70,6 +74,30 @@ export class OverflowMenuPane extends Dialog implements AfterViewInit {
...
+	}
+
+	matchesAttr(el, attr, val) {
+		const styles = window.getComputedStyle(el);
+
+		if (val.includes(styles[attr])) {
+			return true;
+		} else {
+			return false;
+		}
```
theiliad, line 98: how about `return val.includes(styles[attr])`?

theiliad (changes requested):

src/dialog/overflow-menu/overflow-menu-pane.component.ts
DusanMilko: Updated, great feedback!

cal-smith (approved):
Nice! We'll have to integrate this into the upstream position utility at some point, but for now this looks great! 🎉

DusanMilko:
> Nice! We'll have to integrate this into the upstream position utility at some point, but for now this looks great! 🎉

Yeah, agreed. I'll also apply this to other components that have a similar issue like dropdown and etc.

cal-smith:
Sounds good

PR #1559, opened 2020-10-05
fix(file-uploader): multiple file upload/drop only shows one name

### LINKED ISSUE #1541: File upload , drop or upload of multiple files shows only one name
When you upload file by file, the list of uploaded files is shown. When you upload (or drop) multiple, only one is shown

demo https://angular.carbondesignsystem.com/?path=/story/components-file-uploader--drag-and-drop

brianbrowndev:
Fixes #1541  

When many files are uploaded, multiple fileChange events are triggered for each file. Combining this with the additional filtration will cause the files set to be overwritten multiple times and trigger files to self-remove in the file onDestroy method. Fixed to dispatch a single event for file changes on upload/drop.


cal-smith (approved):
Cool 👍 This makes way more sense ... thanks for sorting this out @brianbrowndev!

PR #1561, opened 2020-10-06
feat(pagination): add pagination nav component

whoisbob:
Adds an angular equivalent to Carbon React's pagination nav component: https://react.carbondesignsystem.com/?path=/story/paginationnav--pagination-nav

#### Changelog

**New**

* Added a new pagination nav component along with supporting components

**Changed**

* Not at thing

**Removed**

* Not a thing


theiliad:
🎉

cal-smith (changes requested):
Thanks for taking this on! Overall it looks great, just a couple comments:

I'm not sure we need a whole new module for the nav variant ... `pagination-nav` and `pagination-nav-item` seem like the right selectors to me though 👍

Would it make sense to expose the underlying components as well? I would imagine that some teams would want finer control over the item behaviour ... perhaps a `ibm-pagination-nav-list` that takes `ibm-pagination-nav-item`/`ibm-pagination-overflow` as `ng-content`?

cal-smith (changes requested):
I'm still not totally sure this needs to be a separate module. It should be sufficient to add the nav components to the existing `pagination` module (they can still have separate storybook entries, but I don't think we need a new module for this). Very interested in your thoughts on this @whoisbob 

There's also a few functionality things that aren't quite right (referencing the react impl https://react.carbondesignsystem.com/?path=/story/paginationnav--pagination-nav):
- the next/previous arrows should be disabled when the selection is the first or last page
- I can't seem to select the first page from the middle page when there are two overflow menus
- The model seems to get confused sometimes and I end up selecting the wrong items (I've seen page 5 selected when I clicked on page 9) ... this one is hard to repro so it's not a blocker)

Otherwise this is very close to done! Thanks for putting in the work on this!

cal-smith:
> @cal-smith Heya, sorry for the late reply! I definitely agree with you on the separate module issue. I'll push a fix for that. 

Excellent! Thanks!

> As for the three issues with the PR, I'm having trouble reproducing the second one `I can't seem to select the first page from the middle page when there are two overflow menus` :/ 
> 
> Any ideas on how I could reproduce it?

I was using 10 items with 5 visible ... Once two overflow menus were visible I wasn't able to select the first page ... This is using Firefox, I didn't check Chrome.

If we can't repro that and everything else is fine we can defer it.

cal-smith (dismissed):
I'm not entirely sure why the build is failing again (it can be picky about how exports are declared) but everything else looks good!

Thanks again @whoisbob!

src/pagination/index.ts
```diff
@@ -1,3 +1,6 @@
 export * from "./pagination-model.class";
 export * from "./pagination.component";
 export * from "./pagination.module";
+export * from "./pagination-nav/pagination-overflow.component";
```
cal-smith, line 4: Try exporting everything by name here ... I'm guessing the compiler doesn't like the autogenerated names when it's trying to compile the root module.

whoisbob:
This build doesn't like me

cal-smith:
> This build doesn't like me

I'll pull your branch and take some time to dig into this ... Super sorry about all this 😞

cal-smith:
Sorted!

There were two issues at play:

1. Ambient exports - When there's a component/directive/etc that's used in a module but not exported in the entrypoint ngc will add an export and assign it a generic name (θa, θb, θc, etc). Because we compile each component as a separate entrypoint and then compile the whole thing into the main `carbon-components-angular` entrypoint, it's possible (likely even!) for two entrypoints to generate identical ambient exports - in this case modal and pagination generated a set of the same exports. Explicitly exporting the components solved that issue.
2. "Barrel" imports - ngc isn't smart enough to figure out that `import foo from "./my-folder"` means `... from "./my-folder/index"` and ends up blowing up. The fix for this is to either import explicitly from `/index` or explicitly from each file.

In summary - Thanks for all the work! You did all the right things, `ngc` is just a bit picky.

cal-smith (approved):
LGTM!

whoisbob:
THANK YOU! : D

zvonimirfras (changes requested):
Minor code style changes and there are some things that look suspicious so I want to take a longer look at them so might be more later :)

zvonimirfras (changes requested):

src/pagination/pagination-nav/pagination-nav-model.class.ts
```diff
@@ -0,0 +1,11 @@
+export class PaginationNavModel {
```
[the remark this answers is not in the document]
whoisbob: As long as we're okay with making `pageLength` optional

src/pagination/pagination-nav/pagination-nav.component.ts
```diff
@@ -0,0 +1,249 @@
...
+
+	public getPages() {
+		if (this.totalDataLength <= 1) {
+			return null;
+		}
+		const cuts = this.getCuts(this.currentPage - 1, this.totalDataLength, this.numOfItemsToShow);
+		return this.totalNumbersArray.slice(this.startOffset + cuts.front, (1 + cuts.back) * -1);
+	}
+
+	private getCuts(page, totalItems, itemsThatFit, splitPoint = null) {
```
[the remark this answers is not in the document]
whoisbob: touche

src/pagination/pagination-model.class.ts
```diff
@@ -7,7 +7,7 @@ export class PaginationModel {
 	/**
 	 * Length of page.
 	 */
-	pageLength = 10;
+	pageLength ? = 10;
```
cal-smith, line 10: This is marking the property as optional right? So I think we may want to format it like this:

```suggestion
	pageLength? = 10;
```

cal-smith (changes requested):
One style comment, and there's an outstanding question from @zvonimirfras around the range function ... Otherwise LGTM! Thanks for the updates!

src/pagination/pagination-model.class.ts
whoisbob: I found that the linter didn't like that

cal-smith: Huh, interesting ... We'll take a look at that after this is done 🤔

cal-smith (approved):
LGTM!

src/pagination/pagination-nav/pagination-overflow.component.ts
```diff
@@ -0,0 +1,64 @@
...
+ * * ```html
+ * <ibm-pagination-overflow [fromIndex]="5" [count]="30" (change)="handleChange(value)"></ibm-pagination-overflow>
+ * ```
+ */
+@Component({
+	selector: "ibm-pagination-overflow",
+	template: `
+		<li class="bx--pagination-nav__list-item" *ngIf="count > 1">
+			<div class="bx--pagination-nav__select">
+			  <select
```
zvonimirfras, line 20: There are mixed tabs and spaces in the template indentation which makes it look inconsistent across code editors.
All lines should be indented with tabs only.

src/pagination/pagination.nav.stories.ts
```diff
@@ -0,0 +1,91 @@
+import {
```
zvonimirfras, line 1: The filename should probably be `pagination-nav.stories.ts` (dash instead of dot in the base name)

zvonimirfras (commented):
Overall a very well written PR 👍

cal-smith (approved):

cal-smith:
Thanks for contributing @whoisbob!

src/pagination/pagination-nav/pagination-overflow.component.ts
whoisbob: Sorry my editor tends to misrepresent what I'm looking at vs. whats being saved, if that even sounds possible

PR #1566, opened 2020-10-09
fix(select): synchronize inner select value with ngmodel on change

### LINKED ISSUE #1536: ibm-select value doenst update when ngmodel value changes
## Detailed description

When using the ibm-select component to filter, trying to reset the filter with another component by changing the value of the bound variable. This does not update the selectbox. 

Bug 

ibm-select

Value should reflect the value in the ngModel 

Newest chrome and firefox. 

  "carbon-components": "10.20.0",
  "carbon-components-angular": "4.17.0"

template:
```
<ibm-select [(ngModel)]="filterVal" (ngModelChange)="filter()" >
 <option value="one" >one</option>
 <option value="two" >two</option>
</ibm-select>
<button type="button" (click)="filterVal = 'one'" >Reset</button>
```
component:
```
filterval = 'one'
```

Reproduce:
Clicking the reset button, should reset the visual of the component, but doesn't.

valburyakov:
Fixes #1536 

After select change inner value was not updated and that's why value and ngModel was not in sync and it looks like view is not updated

#### Changelog

**Changed**

*  synchronize  inner value on select changes

### DIFF
--- a/src/select/select.component.ts
+++ b/src/select/select.component.ts
@@ -193,6 +193,7 @@ export class Select implements ControlValueAccessor {
 	 * Sends events to the change handler and emits a `selected` event.
 	 */
 	onChange(event) {
+		this.value = event.target.value;
 		this.onChangeHandler(event.target.value);
 		this.valueChange.emit(event.target.value);
 	}


cal-smith (approved):
Thanks for catching this 👍

PR #1579, opened 2020-10-27
feat(ui-shell): Add icon input to header-menu

Aracturat:
Sometimes for custom menu it is necessary to change a header menu icon. 

For example I want to create language menu (without text, only icon). It can look like this:
image

To support icon changing this PR will add optional icon input for header-menu

#### Changelog

**New**
- feat(ui-shell): Add icon input to header-menu


Aracturat:
In addition, it can help with implementation of notification list https://github.com/IBM/carbon-components-angular/issues/998.

cal-smith (approved):

PR #1582, opened 2020-10-28
1581 - Render custom template content for an overflow menu

blaise2s:
Closes IBM/carbon-components-angular#1581

Adds a new component to allow rendering of custom content into an overflow menu inserted into a div rather than a ul element

#### Changelog

**New**

* `carbon-components-angular/src/dialog/overflow-menu/overflow-menu-custom-pane.component.ts`

**Changed**

* `carbon-components-angular/src/dialog/dialog.module.ts`
* `carbon-components-angular/src/dialog/overflow-menu/overflow-menu.diretive.ts`
* `carbon-components-angular/src/dialog/overflow-menu/overflow-menu.stories.ts`

### DIFF
--- a/src/dialog/dialog.module.ts
+++ b/src/dialog/dialog.module.ts
@@ -15,6 +15,7 @@ import { EllipsisTooltip } from "./tooltip/ellipsis-tooltip.directive";
 
 import { OverflowMenu } from "./overflow-menu/overflow-menu.component";
 import { OverflowMenuPane } from "./overflow-menu/overflow-menu-pane.component";
+import { OverflowMenuCustomPane } from "./overflow-menu/overflow-menu-custom-pane.component";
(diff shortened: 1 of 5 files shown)


blaise2s:
@cal-smith let me know when you have had a chance to review. Thanks!

cal-smith (commented):
Overall everything looks good! To fix the build issues I think you’ll need to export the new component from the modules `index.ts`

I also noticed the menu didn’t stay open when I was testing on mobile... it kinda behaved like a hover menu? Not sure if I missed something there

cal-smith:
Actually it seems like the mobile issue happens with the other existing menus, so that's a seperate issue we'll have to sort out later

blaise2s:
Sounds good! I exported it in the index, so after this comes back successfully 🤞 then I'll just need to get one more approval as it looks like two are required on your side.

PR #1583, opened 2020-10-29
fix(combobox): fix multi combobox closing on selection

DusanMilko:
Combobox multiselect currently closes on selection. It should stay open like how the dropdown multiselect does so user can keep making selections if needed.

This code is borrowed from dropdown logic.

Fixes: https://github.com/IBM/carbon-components-angular/issues/1584


cal-smith (approved):
Cool! Makes sense to me 👍

Issue #1533, opened 2020-09-20
combobox : check on value


huineng:
in formcontrol it can happen that a value has not yet been assigned and that the formcontrol value remains undefined

https://github.com/IBM/carbon-components-angular/blob/27e3216576c871a99e06fdbcc25bc6a93a94ca27/src/combobox/combobox.component.ts#L531

this can lead to a bug on the combox 

```
 "TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))",
```

possible solution 
```
if (this.itemValueKey) {
  // clone the specified item and update its state
  const newValue = Object.assign({}, this.view.getListItems().find(item => item[this.itemValueKey] === value ? value || ""));
  newValue.selected = true;
this.view.propagateSelected([newValue]);
} else {
  this.view.propagateSelected([value ? value || ""]);
}
```

JoelLau:
hi @cal-smith, can I take this up as part of hacktoberfest

cal-smith:
I can't seem to assign you at the moment @JoelLau, but thanks for taking this on!

PR #1585, opened 2020-10-30
 fix(#1533): combobox undefined typeerror

JoelLau:
Closes IBM/carbon-components-angular#1533

Combobox (type: single) throws TypeError because of bad 

` "TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))"`

#### Changelog

**Changed**

* Added fallback value for combobox, specifically, when:
    * type: single
    * `itemValueKey` is falsey

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -545,6 +545,7 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
 				newValue.selected = true;
 				this.view.propagateSelected([newValue]);
 			} else {
+				// all items in propagateSelected must be iterable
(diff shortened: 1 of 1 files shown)


cal-smith (approved):
Thanks for catching this!

I'm going to approve this, but I wonder if this is a sign we need to double check the `propagateSelected` logic around falsey values ... There's several areas where we call `propagateSelected` and this is an easy bug to introduce.

If you could add comment to the line explaining _why_ we're doing this that would also help @JoelLau 

Thanks again!

JoelLau:
@cal-smith i missed the notification - i'm on it now

cal-smith (approved):
Perfect, thanks!

PR #1597, opened 2020-11-06
fix(uishell): add missing aria label on header action

DusanMilko:
Missing aria label makes IBM Access Accessibility fail test


cal-smith (approved):
Thanks for adding this! UI Shell definitely needs a pretty thorough pass over for small things like this

Issue #1607, opened 2020-11-10
Skeleton Placeholder restricted to a fixed size


jdoyle65:
## Detailed description

The `ibm-skeleton-placeholder` component has a fixed height/width of `6.5rem`, making it inflexible when trying to apply to many situations.

It would be better if this component could be styled to fit any sizing need.

PR #1608, opened 2020-11-10
fix: adds size control for skeleton placeholder

jdoyle65:
Addresses #1607 

The current `bx--skeleton__placeholder` class has a fixed `6.5rem` width and height. This PR changes `ibm-skeleton-placholder` to render as a block element, and sets the `bx--skeleton__placeholder` element inside of it to fill the width/height of its container.

Using these changes, we can freely manipulate the size of the skeleton placeholder to fit any need.

#### Changelog

**Changed**

* `skeleton-placeholder.component.ts`
* `skeleton.stories.ts`

#### Screenshot

image

### DIFF
--- a/src/skeleton/skeleton-placeholder.component.ts
+++ b/src/skeleton/skeleton-placeholder.component.ts
@@ -3,6 +3,18 @@ import { Component } from "@angular/core";
 @Component({
 	selector: "ibm-skeleton-placeholder",
 	template: `
-		<div class="bx--skeleton__placeholder"></div>`
+		<div class="bx--skeleton__placeholder"></div>`,
+	styles: [
+		`
+		:host {
+			display: block;
+		}
+
+		.bx--skeleton__placeholder {
+			width: 100%;
+			height: 100%;
+		}
+		`
+	]
 })
 export class SkeletonPlaceholder { }
--- a/src/skeleton/skeleton.stories.ts
+++ b/src/skeleton/skeleton.stories.ts
@@ -10,7 +10,11 @@ storiesOf("Components|Skeleton", module).addDecorator(
 )
 	.addDecorator(withKnobs)
 	.add("Skeleton Placeholder", () => ({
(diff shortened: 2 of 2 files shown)


cal-smith (changes requested):
This is by design: https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/skeleton/_skeleton-placeholder.scss

If we can get some consensus on the `carbon-components` side on how the skeleton should behave and be sized, we can make the updates here as needed. As is this would be a breaking change, since users are expecting a fixed width/height skeleton.

If there's a way to introduce this functionality in a non-breaking way, that preserves Carbon compatibility by default, we can go with that!

Overall I think skeleton _should_ work like this, I just want to make sure we're approaching it the right way.

jdoyle65:
@cal-smith it appears that the React library approaches this by allowing the developer to apply their own classes against the placeholder: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/SkeletonPlaceholder/SkeletonPlaceholder.js

I'm trying to identify the use-case of the fixed-size placeholder in the link provided. Any insight?

cal-smith:
> @cal-smith it appears that the React library approaches this by allowing the developer to apply their own classes against the placeholder: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/SkeletonPlaceholder/SkeletonPlaceholder.js

Ah interesting! It may be best if we take that approach ... That keeps our default behaviour for now, and allows users to override as needed

> I'm trying to identify the use-case of the fixed-size placeholder in the link provided. Any insight?

I'm not sure 😀 It's definitely a bit of a odd design decision.

theiliad (approved):

cal-smith:
Hey @jdoyle65 👋 this has been stagnant for a while, still totally down to take a generic placeholder component if you have time to work on this some more ... Otherwise we can close this out for now and pick it back up later 👍

cal-smith:
It's been a few days, so I'm going to close this one out. More than happy to take this contribution in the future!

PR #1610, opened 2020-11-13
feat(tabletoolbar): add output for custom cancel method

Pablodotnet:
Closes IBM/carbon-components-angular#1611

This feature just adds an output to the `ibm-table-toolbar` so developers can execute something else at the moment they click on cancel on the toolbar when items are selected.

#### Changelog

**New**

* `cancel` Output for `ibm-table-toolbar`

### DIFF
--- a/src/table/toolbar/table-toolbar.component.ts
+++ b/src/table/toolbar/table-toolbar.component.ts
@@ -106,6 +106,8 @@ export class TableToolbar {
 		return { CANCEL: this._cancelText.value as string };
 	}
 
+	@Output() cancel = new EventEmitter;
+
 	actionBarLabel: Overridable = this.i18n.getOverridable("TABLE_TOOLBAR.ACTION_BAR");
 	_cancelText: Overridable = this.i18n.getOverridable("TABLE_TOOLBAR.CANCEL");
 	_batchTextLegacy: Overridable = this.i18n.getOverridable("TABLE_TOOLBAR.BATCH_TEXT");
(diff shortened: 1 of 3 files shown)


zvonimirfras (changes requested):
Looks like spec file shouldn't have changed at all and it looks unrelated to the PR.
Looks good otherwise.

Pablodotnet:
I will add test for this, the other spec file modified here was to fix the lint, you can check the commits 😄

src/table/toolbar/table-toolbar.component.ts
```diff
@@ -106,6 +106,8 @@ export class TableToolbar {
 		return { CANCEL: this._cancelText.value as string };
 	}
 
+	@Output() cancel = new EventEmitter;
```
cal-smith, line 109: Huh I guess the original works, but this is a bit more consistent: 

```suggestion
	@Output() cancel = new EventEmitter();
```

cal-smith (changes requested):
There's also a bunch of random extra spaces added before some lines in the spec ... Overall LGTM!

src/table/toolbar/table-toolbar.component.ts
Pablodotnet: Done! Sorry I missed them haha and also deleted the double spaces, I don't know what happened with lint there :P

zvonimirfras (approved):
It's weird to see that there's still diff in the spec but ah well

cal-smith (approved):
LGTM :+1: Thanks @Pablodotnet!

cal-smith:
> It's weird to see that there's still diff in the spec but ah well

New, higher quality, spaces 😂

PR #1625, opened 2020-11-23
fix(accessibility): swap unnecessary html to button to match core carbon

DusanMilko:
Swap div role="button" to organic button


zvonimirfras (approved):

cal-smith (approved):
Thanks for this! I feel like Carbon has flipped between a div and a button a few times now 😆

DusanMilko:
> Thanks for this! I feel like Carbon has flipped between a div and a button a few times now 😆

Not sure why they are flipping but yeah we should stick with as much inherit utility from browsers as we can imo

Issue #916, opened 2019-11-15
Slider: Create a range slider


stanislavgeorgiev:
## Detailed description

> Describe in detail the issue you're having.

We need a range slider to specify ranges for filters and/or thresholds. We've already implemented it and we're asking if we could contribute it back to the community:

![image](https://user-images.githubusercontent.com/1253469/68965182-2a5fa480-07a9-11ea-8db9-90460d91ef7e.png)

> Is this a feature request (new component, new icon), a bug, or a general issue?

Feature request

> Is this issue related to a specific component?

Slider

> What version of the Carbon Design System are you using?

3.26.1

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

IBM ADDI

cal-smith:
Does this require any style changes @stanislavgeorgiev? If not I don't see a reason why it couldn't be contributed back

stanislavgeorgiev:
No style changes @cal-smith. It's reusing the thumb style and then it just modifies how long the track is to make it look like a range.
If you're ok with it then I will prepare a PR for review next week. In the meantime you could assign this one to me.

cal-smith:
Done! Sounds good!

stanislavgeorgiev:
Great!

stanislavgeorgiev:
@cal-smith What do you think about calling the second value for the range `value2`? In our implementation we have it `lowValue` and `highValue` but obviously I don't want to break anybody else's code and rename `value` to `lowValue` :)

I was initially thinking about creating a `range-slider.component` which will extend the `slider.component`. That didn't go well because I didn't want to clone the template code so I settled to do append some code to the `slider` itself.

zvonimirfras:
Looks like this might have been resolved in #1626. Feel free to reopen if that's not the case :)

PR #1626, opened 2020-11-23
feat(slider): Add 2nd slider handle for range

zvonimirfras:
closes #935 #916


src/slider/slider.component.ts
```diff
@@ -179,6 +191,44 @@ export class Slider implements AfterViewInit, ControlValueAccessor {
 	get value() {
 		return this._value;
 	}
+
+	/** Set the initial value. Available for two way binding */
+	@Input() set value2(v) {
```
cal-smith, line 196: We should be able to use an array to represent the high and low/left and right values ... That way we aren't specifically coding in `value1` and `value2` ... In theory we should be able to support `N` sliders, but we can start with just a tuple.

So `value` should have the type `number | number[]` ... As long as we eimt a new array each time `ngModel` should work fine with this setup. I would expect that we swap everything internally to use arrays, and just dump the contents of `this.value[0]`. `isRange` can probably just be a check for `this.value.length > 1` 🤔

cal-smith (changes requested):
Similar comments to before ... I'd rather not get stuck with `value` and `value2` if possible 🤔

src/slider/slider.component.ts
zvonimirfras: I had similar thoughts.

src/slider/slider.component.ts
```diff
@@ -72,6 +72,14 @@ import { EventService } from "carbon-components-angular/utils";
 						(mousedown)="onMouseDown($event)"
 						(keydown)="onKeyDown($event)">
 					</div>
+					<div
+						#thumb2
```
theiliad, line 76: Just a heads up, the slider in the future should add more aria-* attributes to be properly accessible.

theiliad: https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-1.html

theiliad: https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-2.html

theiliad (changes requested):
agreed wrt all the comments around value 1 & 2. The code can be a lot more flexible than that

src/slider/slider.component.ts
```diff
@@ -147,38 +167,71 @@ export class Slider implements AfterViewInit, ControlValueAccessor {
...
-		if (this.input) {
-			this.input.value = v.toString();
+		if (this.inputs && this.inputs.length) {
+			this.inputs.forEach((input, index) => {
+				input.value = this._value[index].toString();
+			});
 		}
 
 		this.propagateChange(v);
 		this.valueChange.emit(v);
```
cal-smith, line 225: Right now these will always emit an array

```suggestion
		const valueToEmit = this.isRange() ? v : v[0];
		this.propagateChange(valueToEmit);
		this.valueChange.emit(valueToEmit);
```

cal-smith: Thankfully this is the only place we call those!

cal-smith (changes requested):
This is actually kind of amazing ... It just works!

One small thing leftover though

src/slider/slider.component.ts
zvonimirfras: Ah yes, didn't come up in testing, good catch!

cal-smith (approved):

stanislavgeorgiev:
Thanks for putting up the work on this item guys!

cal-smith:
And thank you @stanislavgeorgiev for getting this conversation started!

PR #1631, opened 2020-11-25
chore(docs): add tracking to documentation site

cal-smith:
Should work 🤞


theiliad (approved):
I love that the theme supports hbs

cal-smith:
> I love that the theme supports hbs

The more I get to know typedoc the more neat little features I find ... And it all just works, with no messing around

Issue #1590, opened 2020-11-03
ibm-combo-box with (search) attribute has a violation for Accessibility


GSSCDylan:
## Detailed description

> Describe in detail the issue you're having.

- **ibm-combo-box with (search) attribute will generate an input as following:**
` <input type="text" role="searchbox"   autocomplete="list">`
-  **But the 'autocomplete' attribute's token "list" is not appropriate for the input form field.**

> Is this a feature request (new component, new icon), a bug, or a general issue?

- **bug**

> Is this issue related to a specific component?

- **ibm-combo-box**

> What version of the Carbon Design System are you using?
`"carbon-components": "10.20.0", "carbon-components-angular": "4.17.0"`

> What browser are you working in?
Firefox
## Steps to reproduce the issue

1. Use browser to open an webpage that contains ibm-combo-box with (search) attribute
2. Run IBM Equal Access Accessibility Checker

## Additional information

* Screenshots or code
<img width="1404" alt="image" src="https://user-images.githubusercontent.com/73825731/97976561-d14baf00-1e05-11eb-9b19-ec2e6da714fe.png">
<img width="976" alt="image" src="https://user-images.githubusercontent.com/73825731/97977433-2cca6c80-1e07-11eb-9fd2-4faabe25c668.png">

GSSCDylan:
Hi @cal-smith , please also help to fix this issue ASAP, it's also impact our AD AVT compliance. @stanislavgeorgiev  FYI

PR #1641, opened 2020-12-01
fix(combobox): improve accessibility for combobox

cal-smith:
Closes IBM/carbon-components-angular#1590

Moves around and cleans up all the roles and aria attrs in the component. I'm honestly not sure how we let it get this bad 😱


grenobnik:
> Closes #1590
> 
> Moves around and cleans up all the roles and aria attrs in the component. I'm honestly not sure how we let it get this bad

however, the autocomplete attribute itself is no longer available for combobox input after this 'fix'. What is the recommended way to disable it or provide a list of suggested values other than `getElementById....` in AfterViewInit, which is quite inconvenient for dynamic component.

PR #1643, opened 2020-12-03
fix(accessibility): fix accessibility error for data table with no data

DusanMilko:
Fixes https://github.com/IBM/carbon-components-angular/issues/1617

As per:
https://webaim.org/techniques/tables/data
https://webaim.org/techniques/tables/
https://www.w3.org/WAI/WCAG21/Techniques/html/H43


cal-smith (approved):
Oh wow, this is great 🥳 Thanks for sorting this one out!

PR #1644, opened 2020-12-03
Fix accessibility issues with expandable tile

bailinhuang:
Screen Shot 2020-12-03 at 10 53 26
Fixes accessibility issue shown by IBM Accessibility Checker

### DIFF
--- a/src/tiles/expandable-tile.component.ts
+++ b/src/tiles/expandable-tile.component.ts
@@ -21,7 +21,7 @@ export interface ExpandableTileTranslations {
 			[ngStyle]="{'max-height': expandedHeight + 'px'}"
 			type="button"
 			(click)="onClick()">
-			<div [attr.aria-label]="(expanded ? collapse : expand).subject | async" class="bx--tile__chevron">
+			<div class="bx--tile__chevron">
 				<svg *ngIf="!expanded" width="12" height="7" viewBox="0 0 12 7" title="{{expand.subject | async}}" role="img">
 					<title>{{expand.subject | async}}</title>
 					<path fill-rule="nonzero" d="M6.002 5.55L11.27 0l.726.685L6.003 7 0 .685.726 0z"/>
--- a/src/tiles/expandable-tile.component.ts
+++ b/src/tiles/expandable-tile.component.ts
@@ -22,11 +22,11 @@ export interface ExpandableTileTranslations {
 			type="button"
 			(click)="onClick()">
 			<div [attr.aria-label]="(expanded ? collapse : expand).subject | async" class="bx--tile__chevron">
-				<svg *ngIf="!expanded" width="12" height="7" viewBox="0 0 12 7" role="img">
+				<svg *ngIf="!expanded" width="12" height="7" viewBox="0 0 12 7" title="{{expand.subject | async}}" role="img">
 					<title>{{expand.subject | async}}</title>
 					<path fill-rule="nonzero" d="M6.002 5.55L11.27 0l.726.685L6.003 7 0 .685.726 0z"/>
 				</svg>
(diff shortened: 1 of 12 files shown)


src/tiles/expandable-tile.component.ts
```diff
@@ -22,11 +22,11 @@ export interface ExpandableTileTranslations {
 			type="button"
 			(click)="onClick()">
 			<div [attr.aria-label]="(expanded ? collapse : expand).subject | async" class="bx--tile__chevron">
-				<svg *ngIf="!expanded" width="12" height="7" viewBox="0 0 12 7" role="img">
+				<svg *ngIf="!expanded" width="12" height="7" viewBox="0 0 12 7" title="{{expand.subject | async}}" role="img">
 					<title>{{expand.subject | async}}</title>
 					<path fill-rule="nonzero" d="M6.002 5.55L11.27 0l.726.685L6.003 7 0 .685.726 0z"/>
 				</svg>
-				<svg *ngIf="expanded" width="12" height="7" viewBox="0 0 12 7" role="img">
+				<svg *ngIf="expanded" width="12" height="7" viewBox="0 0 12 7" title="{{expand.subject | async}}" role="img">
```
cal-smith, line 29: We'll want the "collapse" text for this

```suggestion
				<svg *ngIf="expanded" width="12" height="7" viewBox="0 0 12 7" title="{{collapse.subject | async}}" role="img">
```

src/tiles/expandable-tile.component.ts
```diff
@@ -22,11 +22,11 @@ export interface ExpandableTileTranslations {
 			type="button"
 			(click)="onClick()">
 			<div [attr.aria-label]="(expanded ? collapse : expand).subject | async" class="bx--tile__chevron">
```
cal-smith, line 24: Do we need the `aria-label` here if we have a `title` on the SVG? Seems like it may be redundant in this case

src/tiles/expandable-tile.component.ts
```diff
@@ -22,11 +22,11 @@ export interface ExpandableTileTranslations {
...
-				<svg *ngIf="!expanded" width="12" height="7" viewBox="0 0 12 7" role="img">
+				<svg *ngIf="!expanded" width="12" height="7" viewBox="0 0 12 7" title="{{expand.subject | async}}" role="img">
 					<title>{{expand.subject | async}}</title>
```
cal-smith, line 26: Do we need the `<title>` elements within the SVGs as well? Seems redundant...

cal-smith (changes requested):
Thanks for taking this on! Just a couple questions/suggestions before we merge this in...

src/tiles/expandable-tile.component.ts
bailinhuang: It was an issue picked by the IBM Accessibility Checker, it might be that some screen readers don't pick up the <title> element inside the svg. The solution given by the plugin for elements with a "img" role with non visible labels is to provide a title attribute within the them.

cal-smith: Gotcha!

src/tiles/expandable-tile.component.ts
bailinhuang: The attr.aria-label was not changed and was left the way it is

bailinhuang: However, divs should not have aria label attributes since they aren't picked by the screen readers anyway

cal-smith: Gotcha, we should drop it then 👍 No need to have excess attributes potentially messing things up

cal-smith (approved):
Looks good to me!

zvonimirfras (approved):

PR #1653, opened 2020-12-08
fix(table): Add title attr to table data cells

zvonimirfras:
Closes IBM/carbon-components-angular#1630

### DIFF
--- a/src/table/body/table-row.component.ts
+++ b/src/table/body/table-row.component.ts
@@ -56,7 +56,7 @@ import { TableRowSize } from "../table.types";
 					ibmTableData
 					[headers]="model.getHeaderId(j, item.colSpan)"
 					[item]="item"
-					[title]="getItemTitle(item)"
+					[title]="item.title"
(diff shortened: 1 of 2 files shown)


src/table/body/table-row.component.ts
```diff
@@ -218,4 +220,24 @@ export class TableRowComponent {
...
+		if (!item || !item.data) {
+			return "";
+		}
+
+		if (typeof item.data === "string") {
+			return item.data;
+		}
+
+		if (item.data.title) {
+			return item.data.title;
```
cal-smith, line 234: Is there somewhere we can document this? Also `item.data.title` vs. `item.title`? It may make more sense if we make it an optional property on the `TableItem` (for bonus points we could even have it proxy item.data when it's unset 🤔)

zvonimirfras: Good idea!

cal-smith (approved):
Nice!

PR #1656, opened 2020-12-09
fix(position): fix position calculation for dropdown and extract utils

DusanMilko:
Bringing this fix to dropdown https://github.com/IBM/carbon-components-angular/pull/1558/files

If a dropdown/combobox with placeholder had a parent with a position set with some modifying positions the position calculation broke. This fixes that scenario.

### Before:

3 different scenarios. 1st works, 2nd and 3rd break.

Screen Shot 2020-12-08 at 6 07 34 PM
Screen Shot 2020-12-08 at 6 07 44 PM
Screen Shot 2020-12-08 at 6 07 55 PM

### After:

as you can see the same previous situations above are replicated and dropdown positions correctly in all situations.

Screen Shot 2020-12-08 at 6 06 01 PM
Screen Shot 2020-12-08 at 6 06 16 PM
Screen Shot 2020-12-08 at 6 06 25 PM


zvonimirfras (approved):
Makes sense to me.

Donisius (dismissed):

src/utils/utils.ts
```diff
@@ -1,3 +1,22 @@
...
+	return val.includes(styles[attr]);
+}
+
+export function closestAttr(s, t, element) {
+	let el = element;
+
+	do {
+		if (
+			matchesAttr(el, s, t)
+		) {
```
cal-smith, line 16: We should be fine to have the condition on one line
```suggestion
		if (matchesAttr(el, s, t)) {
```

cal-smith (changes requested):
Looks good overall! One style suggestion, but otherwise makes sense 👍 Thanks!

src/utils/utils.ts
DusanMilko: Thanks, updated

cal-smith (approved):

zvonimirfras (approved):

PR #1660, opened 2020-12-10
fix(accessibility): fix title not appearing

DusanMilko:
Modify title attribute properly otherwise not title attribute gets applied


cal-smith (approved):
Excellent 👍 Thanks!

PR #1663, opened 2020-12-15
fix(position): fix no placeholder case with position, add animatio…

DusanMilko:
Fixing case where no placeholder exists and element has parents with position set and some spacing. Issue is fixed indirectly by adding animationService tick to mimic what dropdown is doing. Also fixing calculation for overflow menu as menu button is 40px now instead of 32 so half is 20px instead of the previous 16px.

Before:

Screen Shot 2020-12-14 at 6 09 31 PM

After:

Screen Shot 2020-12-14 at 6 07 08 PM

This should conclude all possible cases:

Case 1: no ibm placeholder and element has no parent with position set with some spacing
Case 2: ibm placeholder and element has no parent with position set with some spacing
Case 3: no ibm placeholder and element has parent with position set with some spacing
Case 4: ibm placeholder present next to element and element has parent with position set with some spacing
Case 5: ibm placeholder present next to some parent of element and element has parent with position set with some spacing

### DIFF
--- a/src/dialog/dialog.component.ts
+++ b/src/dialog/dialog.component.ts
@@ -18,7 +18,7 @@ import {
 import Position, { position, AbsolutePosition, Positions } from "@carbon/utils-position";
 import { cycleTabs, getFocusElementList } from "carbon-components-angular/common";
 import { CloseMeta, CloseReasons, DialogConfig } from "./dialog-config.interface";
-import { ElementService } from "carbon-components-angular/utils";
+import { AnimationFrameService, ElementService } from "carbon-components-angular/utils";
 
 /**
  * Implements a `Dialog` that can be positioned anywhere on the page.
@@ -55,6 +55,9 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
 	public placement: string;
 
 	protected visibilitySubscription = new Subscription();
+
+	protected animationFrameSubscription = new Subscription();
+
 	/**
 	 * Handles offsetting the `Dialog` item based on the defined position
 	 * to not obscure the content beneath.
@@ -80,7 +83,8 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
 	 */
 	constructor(
 		protected elementRef: ElementRef,
-		protected elementService: ElementService
+		protected elementService: ElementService,
+		protected animationFrameService: AnimationFrameService
 	) {}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/dialog/dialog.component.ts
```diff
@@ -80,7 +83,8 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
 	 */
 	constructor(
 		protected elementRef: ElementRef,
-		protected elementService: ElementService
+		protected elementService: ElementService,
+		protected animationFrameService: AnimationFrameService
```
cal-smith, line 87: Since this is contained to the dialog base component can we make this optional? Otherwise it ends up as a breaking change ...

```suggestion
		@Optional() protected animationFrameService: AnimationFrameService = null
```

src/dialog/dialog.component.ts
```diff
@@ -116,6 +120,10 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
 
 		const parentElement = this.dialogConfig.parentRef.nativeElement;
 
+		this.animationFrameSubscription = this.animationFrameService.tick.subscribe(() => {
+			this.placeDialog();
+		});
+
```
cal-smith, line 126: Also gotta wrap this in a conditional because the service is optional ...

```suggestion
		if (this.animationFrameService) {
			this.animationFrameSubscription = this.animationFrameService.tick.subscribe(() => {
				this.placeDialog();
			});
		}
```

src/dialog/dialog.component.ts
```diff
@@ -55,6 +55,9 @@ export class Dialog implements OnInit, AfterViewInit, OnDestroy {
 	public placement: string;
 
 	protected visibilitySubscription = new Subscription();
+
+	protected animationFrameSubscription = new Subscription();
```
cal-smith, line 59: Do we unsubscribe from this later?

src/dialog/overflow-menu/overflow-menu-custom-pane.component.ts
```diff
@@ -27,25 +27,26 @@ export class OverflowMenuCustomPane extends Dialog implements AfterViewInit {
 	constructor(
 		protected elementRef: ElementRef,
 		protected i18n: I18n,
+		protected animationFrameService: AnimationFrameService,
```
cal-smith, line 30: _probably_ want to make this optional as well? I'm not sure how many people are extending the pane, but we should at least be consistent with `elementService` ... Applies to all the dialog.component derivatives

cal-smith (changes requested):
Looks good overall! Just a couple things to make sure we're not breaking people.

cal-smith (approved):
LGTM! Thanks for the updates 🎉

zvonimirfras (approved):

PR #1666, opened 2020-12-15
fix(table-model): fixed issue when totalDataLength is 0

Licen-it:
Closes IBM/carbon-components-angular#1665

`0 ==  false`, this was causing an issue with pagination that was showing the entire array length instead of the manually set length when it was set to 0

#### Changelog

**Changed**

* `totalDataLength = 0` is no longer ignored


cal-smith (approved):
Makes sense to me :+1: Thanks!

Donisius (approved):

PR #1670, opened 2020-12-16
fix(icon): change import of icons from @carbon/icons

Aracturat:
Import from '@carbon/icons' leads to import of all icons from this package. 

When I changed import to direct imports of icons, the size of my application vendor bundle decreased:
Dev: **-9MB**
Prod: **-6KB**

#### Changelog

**Changed**

* Use direct import of icons in icon.module.ts

### DIFF
--- a/src/icon/icon.module.ts
+++ b/src/icon/icon.module.ts
@@ -7,35 +7,33 @@ import { IconDirective } from "./icon.directive";
 import { IconService } from "./icon.service";
 
 // icon imports
-import {
-	Add16,
-	Calendar16,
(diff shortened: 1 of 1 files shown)


cal-smith (approved):
I did some testing before and figured there wasn't a huge difference, but 9mb smaller for the dev bundle is really significant ... Thanks for doing the testing on this! 🎉

Aracturat:
I found these modules when I analyzed my app bundles using https://statoscope.tech/ , maybe this tool can help you to optimize library more.

PR #1675, opened 2020-12-21
Typo fix

Jaskon:
ibomIcon -> ibmIcon
Invalid icon displays incorrectly because of that


cal-smith (approved):
Thanks for catching this!

PR #1683, opened 2020-12-26
feat(tag-filter): emit original click event object on close

### LINKED ISSUE #1606: Need an event or an object to be passed from ibm-tag-filter component to determine the source of click event
Describe in detail the issue you're having.

- When clicked on close button of ibm-tag-filter, both click and close events are emitted (in this sequence : click and then close)
I need to differentiate whether the click target is close button or  .bx--tag__label /  .bx--tag--filter. 
If the click is on the filter I should call edit function.
If the click is on close button , I  do different operation of closing and clearing the filter.

Is this a feature request (new component, new icon), a bug, or a general issue?
general issue 

Is this issue related to a specific component?
ibm-tag-filter

What did you expect to happen? What happened instead? What would you like to see changed?
Expect to get an object or event that allows me to determine where the click came from  and emit close event before click event which allows me to set a flag.
But now , no event/object  is passed on click or close. And click is getting emitted first before close .

What browser are you working in?
Chrome

What version of the Carbon Design System are you using?
"carbon-components": "~10.17.0",
"carbon-components-angular": "^4.23.1",

Donisius:
Closes #1606


zvonimirfras (approved):

cal-smith (changes requested):
Emitting the original event puts us into a weird spot where people will use implementation details of the component to do things. This makes any change to the template a breaking change, which is not ideal.

Instead we should emit an object that provides information as needed. In this case, it should be enough to emit something to distinguish clicks on the tag vs clicks on the close button. `{ close: true }` or `{ action: "close" | "click" }` perhaps?

zvonimirfras:
> Emitting the original event puts us into a weird spot where people will use implementation details of the component to do things. This makes any change to the template a breaking change, which is not ideal.
> 
> Instead we should emit an object that provides information as needed. In this case, it should be enough to emit something to distinguish clicks on the tag vs clicks on the close button. `{ close: true }` or `{ action: "close" | "click" }` perhaps?

Let's go with `action`.

Donisius:
> Emitting the original event puts us into a weird spot where people will use implementation details of the component to do things. This makes any change to the template a breaking change, which is not ideal.
> 
> Instead we should emit an object that provides information as needed. In this case, it should be enough to emit something to distinguish clicks on the tag vs clicks on the close button. `{ close: true }` or `{ action: "close" | "click" }` perhaps?

Would it be a breaking change if the emitted value for clicks on the tag was changed to { action: "click" }, since for that case people could already be using the template to do things.

zvonimirfras:
> > Emitting the original event puts us into a weird spot where people will use implementation details of the component to do things. This makes any change to the template a breaking change, which is not ideal.
> > Instead we should emit an object that provides information as needed. In this case, it should be enough to emit something to distinguish clicks on the tag vs clicks on the close button. `{ close: true }` or `{ action: "close" | "click" }` perhaps?
> 
> Would it be a breaking change if the emitted value for clicks on the tag was changed to { action: "click" }, since for that case people could already be using the template to do things.

If it was emitting something and then it started emitting something else that doesn't include the information it previously included (and in the same form), that would be a breaking change.
You can add new things, though.

zvonimirfras (approved):

cal-smith (approved):
Makes sense to me 👍 Thanks!

PR #1693, opened 2021-01-07
fix(option): fix dropdown/combobox option emitting change on default …

DusanMilko:
Should fix https://github.com/IBM/carbon-components-angular/issues/1667

If you set the value in setTimeout to selected: true, the dropdown was getting focus. With initial values being set we don't want to focus on item. I believe this will fix the issue. I encourage feedback or other solutions if you see better ones.

I believe this line was added in Nov and started causing the issue in dropdown-list


theiliad (dismissed):

cal-smith:
This is tricky, since this is actually desired behaviour in a number of cases (perhaps most cases?) see here for the change: https://github.com/IBM/carbon-components-angular/pull/1634

If this is causing the combo box to get focused on init we should try to address that directly, since that's definitely not desirable behaviour, but I don't see a benefit to changing when the dropdown list emits items.

cal-smith (changes requested):
See above comment

DusanMilko:
> See above comment

By emitting select event it'll trigger the selected event on parent which could have unintended triggers, like triggering validation or etc.

cal-smith (approved):
Perfect! Thanks for sorting this out 🎉

zvonimirfras (approved):

PR #1695, opened 2021-01-11
fix(dropdown): correct isUpdate behaviour

cal-smith:
Follows from https://github.com/IBM/carbon-components-angular/pull/1693/files `isUpdate = true` is for updates originating from the component, `isUpdate = false` is for organic selections. This standardizes that usage, and expands the usage to `doEmitSelect` (since it wasn't handling everything like it should)


zvonimirfras (approved):

DusanMilko:
@cal-smith good catch

PR #1698, opened 2021-01-13
feat(button): add iconOnly option

### LINKED ISSUE #1593: Add copy button component
Screen Shot 2020-07-13 at 12 08 33 PM

Donisius:
Closes #1593

I included the tooltip classes within the button directive as hostbindings for the assistive text because using the tooltip directive would place the tooltip into a placeholder component or to the body of the document. This conflicts with the html structure in https://react.carbondesignsystem.com/?path=/story/button--icon-button, unless you put the placeholder component inside the button itself which obviously isn't an option.

### DIFF
--- a/src/button/button.directive.ts
+++ b/src/button/button.directive.ts
@@ -39,13 +39,13 @@ export class Button implements OnInit {
 	 * Possible placements are `top`, `bottom`, `left`, `right`.
 	 * If assistive text is not used, this can be left undefined.
 	 */
-	@Input() assistiveTextPlacement: string;
+	@Input() assistiveTextPlacement: "top" | "bottom" | "left" | "right";
 	/**
 	 * If assistive text is used, this specifies the alignment.
 	 * Possible alignments are `center`, `start`, `end`.
 	 * If assistive text is not used, this can be left undefined.
 	 */
-	@Input() assistiveTextAlignment: string;
+	@Input() assistiveTextAlignment: "center" | "start" | "end";
 
 	// a whole lot of HostBindings ... this way we don't have to touch the elementRef directly
 	@HostBinding("class.bx--btn") get baseClass() {
--- a/src/button/button.directive.ts
+++ b/src/button/button.directive.ts
@@ -34,6 +34,19 @@ export class Button implements OnInit {
 	 * Specify the size of the button
 	 */
 	@Input() size: ButtonSize;
+	/**
+	 * If assistive text is used, this specifies the placement.
+	 * Possible placements are `top`, `bottom`, `left`, `right`.
+	 * If assistive text is not used, this can be left undefined.
+	 */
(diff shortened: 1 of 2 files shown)


src/button/button.directive.ts
```diff
@@ -34,6 +34,19 @@ export class Button implements OnInit {
 	 * Specify the size of the button
 	 */
 	@Input() size: ButtonSize;
+	/**
+	 * If assistive text is used, this specifies the placement.
+	 * Possible placements are `top`, `bottom`, `left`, `right`.
+	 * If assistive text is not used, this can be left undefined.
+	 */
+	@Input() assistiveTextPlacement: string;
```
zvonimirfras, line 42:
```suggestion
	@Input() assistiveTextPlacement: "top" | "bottom" | "left" | "right";
```

src/button/button.directive.ts
```diff
@@ -34,6 +34,19 @@ export class Button implements OnInit {
...
+	/**
+	 * If assistive text is used, this specifies the alignment.
+	 * Possible alignments are `center`, `start`, `end`.
+	 * If assistive text is not used, this can be left undefined.
+	 */
+	@Input() assistiveTextAlignment: string;
```
zvonimirfras, line 48:
```suggestion
	@Input() assistiveTextAlignment: "center" | "start" | "end";
```

zvonimirfras (changes requested):

cal-smith (changes requested):

src/button/button.directive.ts
```diff
@@ -34,6 +34,19 @@ export class Button implements OnInit {
 	 * Specify the size of the button
 	 */
 	@Input() size: ButtonSize;
+	/**
+	 * If assistive text is used, this specifies the placement.
+	 * Possible placements are `top`, `bottom`, `left`, `right`.
+	 * If assistive text is not used, this can be left undefined.
+	 */
+	@Input() assistiveTextPlacement: "top" | "bottom" | "left" | "right";
```
cal-smith, line 42: Is there a sane default we can use for this and the alignment input? Perhaps `top` `center`?

cal-smith (changes requested):
Otherwise looks good

src/button/button.directive.ts
Donisius: If we have an initial value then the tooltip classes would be applied to buttons that don't have the assistive text. It wouldn't really do anything besides just being there though.

cal-smith: Could we make it conditional on `this.iconOnly === true`? That way users don't have to set the placement and alignment for simple cases

Donisius: That's a good idea!

zvonimirfras (approved):

cal-smith (approved):

PR #1708, opened 2021-01-19
fix(title): fix titles in overflow menu and table items

DusanMilko:
Fixes: https://github.com/IBM/carbon-components-angular/issues/1707

1. Moved offset/scrollwidth calculation to correct element for overflow menu

2. Pushed json object to JSON.stringify as better result than .toString.
- propose maybe doing this.data[Object.keys(this.data)[0]] and return the "first" value


zvonimirfras (approved):

cal-smith (approved):
This works :+1: ideally everyone should set a `title` for their table items when it contains complex data structures

PR #1721, opened 2021-02-01
feat(table): Enable expanded data to be table rows

### LINKED ISSUE #1441: Table expanded data with columns
## Detailed description

This is a feature request because I have a design that needs to have the expanded data separated with columns to be setted respecting the table headers, with the next screenshot it will be more understandable.
Currently if you set the expanded data with the template way, it sets the template to a td with colspan 2 I think, so it is limited to that td.

Screen Shot 2020-08-06 at 10 13 38 AM

zvonimirfras:
closes #1441

In order to make the expanded data table rows of the same table, one would pass in the expanded data in the same format as table model data, and set `expandAsTable` to `true` on the table item containing the expanded data. (See story, it's really simple to use)

That tells the table to use that data to render table rows instead of one row with the column that fills the majority of the table containing the expanded data (the default).

`showSelectionColumnCheckbox` works with `showSelectionColumn` to hide the checkbox in this case, but keep the column because it's needed for alignment with the header. This doesn't require any action from the dev.


src/table/body/table-body.component.ts
```diff
@@ -41,14 +41,26 @@ import { TableRowSize } from "../table.types";
...
 				</tr>
-				<tr
-					*ngIf="model.isRowExpanded(i) && !model.isRowFiltered(i)"
-					ibmTableExpandedRow
-					ibmExpandedRowHover
-					[row]="row"
-					[expanded]="model.isRowExpanded(i)"
-					[skeleton]="skeleton">
-				</tr>
+				<ng-container *ngIf="model.isRowExpanded(i) && !model.isRowFiltered(i)">
```
cal-smith, line 44: Kinda unrelated to this PR, but since you're digging around in here anyway - If you check out the react storybook you'll see that the child expandable rows are always present in the DOM. They use CSS to toggle the visibility. The big advantage there is that they already have code to handle striping the rows correctly - right now our rows end up with some weird styles when striped.

It would also be one less DOM change, which is always good from a performance perspective.

cal-smith (changes requested):
Overall looks good, just one suggestion while you're here, but really we could do that as it's own issue.

We'll also need to upgrade the datagrid interaction stuff to handle all of this (it kinda falls over at the moment, which isn't great)

zvonimirfras:
> Overall looks good, just one suggestion while you're here, but really we could do that as it's own issue.
> 
> We'll also need to upgrade the datagrid interaction stuff to handle all of this (it kinda falls over at the moment, which isn't great) but that can **definitely** be done as part of another issue.

I looked at that when I was working on this and datagrid interaction seemed to be falling over when expanded before this, and worked fine when not expanded. This doesn't change that, so it felt like a separate issue.

src/table/body/table-row.component.ts
```diff
@@ -28,18 +28,24 @@ import { TableRowSize } from "../table.types";
...
-				[selected]="selected"
-				[label]="getCheckboxLabel()"
-				[row]="row"
-				[skeleton]="skeleton"
-				[headers]="model.getHeaderId('select')"
-				(change)="onSelectionChange()">
-			</td>
+			<ng-container *ngIf="!skeleton && showSelectionColumn && !enableSingleSelect">
+				<td
+					*ngIf="!showSelectionColumnCheckbox; else tableCheckboxTemplate">
```
Donisius, line 33: Not sure if it's related to this, but if you turn off `showSelectionColumn`, there is still an extra `td` that shows up in expanded rows with more than one column.

image

Donisius (changes requested):

src/table/body/table-body.component.ts
zvonimirfras: I looked into it briefly and it's not clear how css way will work for the rows. We already have an issue open for the striping, so let's do this as part of that issue.

src/table/body/table-body.component.ts
```diff
@@ -152,4 +164,16 @@ export class TableBody {
 	getExpandButtonAriaLabel(): Observable<string> {
 		return this._expandButtonAriaLabel.subject;
 	}
+
+	firstExpandedDataInRow(row) {
+		const found = row.find(d => d.expandedData);
+		if (found) {
+			return found.expandedData;
+		}
+		return found;
```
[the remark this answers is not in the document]
zvonimirfras: How?

src/table/body/table-row.component.ts
zvonimirfras: It _is_ related. Good catch!

src/table/body/table-body.component.ts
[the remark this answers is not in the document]
Donisius: I thought it would return the found object if expandedData didn't exist. I guess it won't work

src/table/body/table-body.component.ts
cal-smith: Sounds like a plan 👍

src/table/body/table-body.component.ts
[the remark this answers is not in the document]
cal-smith: That should work actually 🤔 I'm just not sure if our version of typescript supports it (`return found.?expandedData`)

cal-smith (dismissed):
LGTM 👍 Thanks!

src/table/body/table-body.component.ts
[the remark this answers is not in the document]
zvonimirfras: Yeah, no. 😔

Donisius (approved):

cal-smith (approved):

Pablodotnet:
Thanks for this!! I'll check this definitely ASAP!!! 😄

PR #1736, opened 2021-02-22
feat(tabs): update to scrollable tabs

### LINKED ISSUE #1724: Support for scrolling tabs with overflow nav buttons
Currently when tabs overflow, it changes to a drop down, something that can be replicated by shrinking the window size here: https://angular.carbondesignsystem.com/?path=/story/components-tabs--basic

We have a requirement for our tabs to support the carbon components style scrolling with overflow nav buttons.
Something that has been working for the react framework, discussed here https://github.com/carbon-design-system/carbon/pull/6410

It's a feature request to replace the drop down or add support for scrollable.

Tabs

Chrome

carbon-components: v10.27.0 
carbon-components-angular: v4.39

We are working on a MVP deadline and want to know if you have plans to remedy this in the new feature and could provide ETA?

zvonimirfras:
Closes #1724 

Brings the tabs up to date with react implementation. It borrows some approaches and interfaces from react implementation to match the behavior as closely as possible.

Tabs could use a bit of refactoring. Because of differences between tab headers and tab header group, it might not be reasonable to have one extend the other. It might make sense to have a tab-headers-base that contains common functionality between the two, that both extend. Not in this PR though. If we agree on this approach, I'll open an issue with this info.

### DIFF
--- a/src/tabs/tab-header-group.component.ts
+++ b/src/tabs/tab-header-group.component.ts
@@ -323,13 +323,12 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy, OnChanges, O
 		const rightEdgeReached =
 			direction === 1 &&
 			scrollLeft + clientWidth >= scrollWidth - this.OVERFLOW_BUTTON_OFFSET;
-		if (leftEdgeReached || rightEdgeReached) {
-			if (leftEdgeReached) {
-				this.rightOverflowNavButton.nativeElement.focus();
-			}
-			if (rightEdgeReached) {
-				this.leftOverflowNavButton.nativeElement.focus();
-			}
+
+		if (leftEdgeReached) {
+			this.rightOverflowNavButton.nativeElement.focus();
+		}
+		if (rightEdgeReached) {
+			this.leftOverflowNavButton.nativeElement.focus();
 		}
 	}
 
--- a/src/tabs/tab-header-group.component.ts
+++ b/src/tabs/tab-header-group.component.ts
@@ -15,9 +15,10 @@ import {
 	OnInit
 } from "@angular/core";
 
+import { Subscription } from "rxjs";
+import { EventService } from "carbon-components-angular/utils";
+
 import { TabHeader } from "./tab-header.component";
-import { fromEvent, Subscription } from "rxjs";
-import { debounceTime } from "rxjs/operators";
 
 @Component({
 	selector: "ibm-tab-header-group",
@@ -93,7 +94,7 @@ import { debounceTime } from "rxjs/operators";
 	</nav>
 	`
 })
-export class TabHeaderGroup implements AfterContentInit, OnDestroy, OnChanges, OnInit {
+export class TabHeaderGroup implements AfterContentInit, OnChanges, OnInit {
 	/**
 	 * Set to 'true' to have tabs automatically activated and have their content displayed when they receive focus.
 	 */
@@ -166,9 +167,12 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy, OnChanges, O
 	private _cacheActive = false;
 
 	private overflowNavInterval;
-	private windowResizeSubscription;
 
-	constructor(protected elementRef: ElementRef, protected changeDetectorRef: ChangeDetectorRef) { }
+	constructor(
+		protected elementRef: ElementRef,
+		protected changeDetectorRef: ChangeDetectorRef,
+		protected eventService: EventService
+	) { }
 
 	// keyboard accessibility
 	/**
@@ -245,9 +249,7 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy, OnChanges, O
 	}
 
 	ngOnInit() {
-		this.windowResizeSubscription = fromEvent(window, "resize", { passive: true })
-			.pipe(debounceTime(200))
-			.subscribe(() => this.handleScroll());
+		this.eventService.on(window as any, "resize", () => this.handleScroll());
 	}
 
 	ngAfterContentInit() {
@@ -355,10 +357,5 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy, OnChanges, O
 
 	public handleOverflowNavMouseUp() {
 		clearInterval(this.overflowNavInterval);
-	}
-
-	ngOnDestroy() {
-		this.selectedSubscriptionTracker.unsubscribe();
-		this.windowResizeSubscription.unsubscribe();
 	}
 }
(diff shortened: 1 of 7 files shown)


cal-smith (approved):
Looks good to me!

> Tabs could use a bit of refactoring. Because of differences between tab headers and tab header group, it might not be reasonable to have one extend the other. It might make sense to have a tab-headers-base that contains common functionality between the two, that both extend. Not in this PR though. If we agree on this approach, I'll open an issue with this info.

I would be inclined to take a bit of a different approach - if we're going to do a major refactor, we should decompose the tabs into a bunch of components and directives and use that to build up whatever variants we need. The scrolling behaviour could be a directive, the `nav` container a component, the header items a component, etc.

src/tabs/tab-header-group.component.ts
```diff
@@ -104,20 +128,47 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy, OnChanges {
 	 * ContentChildren of all the tabHeaders.
 	 */
 	@ContentChildren(TabHeader) tabHeaderQuery: QueryList<TabHeader>;
+	// @ts-ignore
```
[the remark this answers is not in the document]
zvonimirfras:
```
Argument of type '{ static: boolean; }' is not assignable to parameter of type '{ read?: any; }'.
Object literal may only specify known properties, and 'static' does not exist in type '{ read?: any; }'
```

zvonimirfras:
> Looks good to me!
> 
> > Tabs could use a bit of refactoring. Because of differences between tab headers and tab header group, it might not be reasonable to have one extend the other. It might make sense to have a tab-headers-base that contains common functionality between the two, that both extend. Not in this PR though. If we agree on this approach, I'll open an issue with this info.
> 
> I would be inclined to take a bit of a different approach - if we're going to do a major refactor, we should decompose the tabs into a bunch of components and directives and use that to build up whatever variants we need. The scrolling behaviour could be a directive, the `nav` container a component, the header items a component, etc.

I did say "a bit" 😄 
If we're doing a major refactor, it might as well be all of it.

src/tabs/tab-headers.component.ts
```diff
@@ -281,86 +329,75 @@ export class TabHeaders implements AfterContentInit, OnChanges {
...
+			direction === 1 &&
+			scrollLeft + clientWidth >= scrollWidth - this.OVERFLOW_BUTTON_OFFSET;
+		if (leftEdgeReached || rightEdgeReached) {
+			if (leftEdgeReached) {
+				this.rightOverflowNavButton.nativeElement.focus();
+			}
+			if (rightEdgeReached) {
+				this.leftOverflowNavButton.nativeElement.focus();
+			}
 		}
```
Donisius, line 375: Is the parent `leftEdgeReached || rightEdgeReached`  necessary here?

src/tabs/tab-header-group.component.ts
```diff
@@ -259,67 +304,62 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy, OnChanges {
...
-		if (!this.tabListVisible) { return; }
+		tabList.scrollLeft += direction * multiplier;
 
-		const target = (event.target as HTMLElement).closest("a");
+		const leftEdgeReached =
+			direction === -1 && scrollLeft < this.OVERFLOW_BUTTON_OFFSET;
+		const rightEdgeReached =
+			direction === 1 &&
+			scrollLeft + clientWidth >= scrollWidth - this.OVERFLOW_BUTTON_OFFSET;
+		if (leftEdgeReached || rightEdgeReached) {
```
Donisius, line 326: Is the parent `leftEdgeReached || rightEdgeReached` necessary here?

Donisius (changes requested):

src/tabs/tab-headers.component.ts
zvonimirfras: You know what? Probably not. This logic comes from the reference implementation.

Donisius (approved):

src/tabs/tab-header-group.component.ts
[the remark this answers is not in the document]
cal-smith: Huh ... odd

src/tabs/tab-header-group.component.ts
```diff
@@ -191,18 +242,12 @@ export class TabHeaderGroup implements AfterContentInit, OnDestroy, OnChanges {
...
 
-	@HostListener("focusout", ["$event"])
-	focusOut(event) {
-		if (this.tabListVisible && !this.elementRef.nativeElement.contains(event.relatedTarget)) {
-			this.tabListVisible = false;
-		}
+	ngOnInit() {
+		this.windowResizeSubscription = fromEvent(window, "resize", { passive: true })
+			.pipe(debounceTime(200))
+			.subscribe(() => this.handleScroll());
```
cal-smith, line 250: Would the `DocumentService` or `EventService` in `utils/UtilsModule` help here? Both of them automatically handle and clean up subscriptions, which would make this a bit cleaner

cal-smith: No debounce, but honestly all that usually does is slow down the UI - most changes are fast enough without a debounce that it feels way better.

src/tabs/tab-headers.component.ts
```diff
@@ -232,18 +281,12 @@ export class TabHeaders implements AfterContentInit, OnChanges {
...
 
-	@HostListener("focusout", ["$event"])
-	focusOut(event) {
-		if (this.tabListVisible && !this.elementRef.nativeElement.contains(event.relatedTarget)) {
-			this.tabListVisible = false;
-		}
+	ngOnInit() {
+		this.windowResizeSubscription = fromEvent(window, "resize", { passive: true })
+			.pipe(debounceTime(200))
+			.subscribe(() => this.handleScroll());
```
cal-smith, line 289: Same comment here with regards to the event handling

cal-smith (changes requested):
Overall looks good, just one small thing I missed the first time around

src/tabs/tab-header-group.component.ts
zvonimirfras: The debounce makes it function equivalently to reference implementation.

cal-smith: Do we need it though? Playing with the react implementation, it definitely makes it a bit weird (the overflow button hangs around for a while after it should disappear), and 200ms is pretty short - how much does that really improve performance?

Either way, if we do need it we can adjust our services as needed, and my original question is still relevant: can we use `DocumentService` or `EventService` for the resize events? Auto-cleanup is a pretty compelling bonus I think.

zvonimirfras: From a usability perspective, `EventService` takes gold compared to `fromEvent`

Donisius (approved):

cal-smith (approved):
Nice! Everything seems good to me 🎉

PR #1742, opened 2021-02-25
fix(search): added aria-label

### LINKED ISSUE #1739: Accessibility Issue: <ibm-search>
## Detailed description

> 
If you use two ibm search components on the same page, you get an accessibility violation from IBM's accessibility plugin about "Multiple elements with "search" role do not have unique labels" and the only fixes are to add the aria-label or aria-labeledby attributes to the elements with the "search" role.
Edit: issue also happens if you have one `<ibm-search>` and one `<ibm-table-toolbar-search>`
>
General Issue
>
Search component, `<ibm-search>`
>
Arialabel input needs to be added and binded to the [div with the 'search' role](https://github.com/IBM/carbon-components-angular/blob/a4953f2de3b2b3b37541978ea65cb589313005cd/src/search/search.component.html#L1-L14)
>
Chrome
>
v4
>  What offering/product do you work on? Any pressing ship or release dates we should be aware of?
>
MCMP
1. Install the [IBM accessibility checker](https://www.ibm.com/able/toolkit/tools/)
2. Run it on a webpage with two <ibm-search> components or one `<ibm-search>` and one `<ibm-table-toolbar-search>
`

amchris98:
Fixes #1739

### DIFF
--- a/src/search/search.component.html
+++ b/src/search/search.component.html
@@ -10,6 +10,7 @@
 		'bx--toolbar-search--active': toolbar && active
 	}"
 	role="search"
+	aria-label="ariaLabel"
 	tabindex="0"
 	(click)="openSearch()">
 	<label class="bx--label" [for]="id">{{label}}</label>
--- a/src/search/search.component.html
+++ b/src/search/search.component.html
@@ -10,7 +10,7 @@
 		'bx--toolbar-search--active': toolbar && active
 	}"
 	role="search"
-	aria-label="ariaLabel"
+	[attr.aria-label]="ariaLabel"
 	tabindex="0"
 	(click)="openSearch()">
 	<label class="bx--label" [for]="id">{{label}}</label>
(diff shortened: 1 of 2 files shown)


src/search/search.component.html
```diff
@@ -10,6 +10,7 @@
 		'bx--toolbar-search--active': toolbar && active
 	}"
 	role="search"
+	aria-label="ariaLabel"
```
zvonimirfras, line 13: This puts literal string `"ariaLabel"` for aria-label. It should probably be something like 
```suggestion
	[attr.aria-label]="ariaLabel"
```
to make it take the input value.

zvonimirfras (changes requested):
Thanks for the quick turnaround. Some changes requested :)

src/search/search.component.html
amchris98: oops my bad, muscle memory has me used to feeding an aria-label as an input to other components.

cal-smith (approved):
LGTM! Thanks for this! 🚀

zvonimirfras (approved):

PR #1753, opened 2021-03-14
feat(Toggle): add ControlValueAccessor method for disable

### LINKED ISSUE #1555: disabled status in toggle component give angular warning
There's a warning when using the disabled attribute when using formcontrols
Now when not adding this directive but doing it via the control as recommended. It doesn't work , even if the formcontrol is DISABLED, the input element is not. So i have to use this [disabled]

```
It looks like you're using the disabled attribute with a reactive form directive. If you set disabled to true
      when you set up this control in your component class, the disabled attribute will actually be set in the DOM for
      you. We recommend using this approach to avoid 'changed after checked' errors.

      Example:
      form = new FormGroup({
        first: new FormControl({value: 'Nancy', disabled: true}, Validators.required),
        last: new FormControl('Drew', Validators.required)
      });
```

Donisius:
Closes #1555

### DIFF
--- a/src/toggle/toggle.component.ts
+++ b/src/toggle/toggle.component.ts
@@ -156,6 +156,17 @@ export class Toggle extends Checkbox {
 		Toggle.toggleCount++;
 	}
 
+	/**
+	 * `ControlValueAccessor` method to programmatically disable the toggle input.
+	 *
+	 * ex: `this.formGroup.get("myToggle").disable();`
+	 *
+	 * @param isDisabled `true` to disable the input
+	 */
+	setDisabledState(isDisabled: boolean) {
+		this.disabled = isDisabled;
+	}
+
 	getOffText(): Observable<string> {
 		return this._offValues.subject;
 	}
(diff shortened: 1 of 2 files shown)


zvonimirfras (approved):
Cool. Are there any other components we can do this for?

Donisius:
> Cool. Are there any other components we can do this for?

It looks like it, do you want me to include them all in this PR?

zvonimirfras:
> > Cool. Are there any other components we can do this for?
> 
> It looks like it, do you want me to include them all in this PR?

Not my initial idea but if it wouldn't make this PR too big, sure. Then rename the PR.
If it would, then separate PRs

Donisius:
> > > Cool. Are there any other components we can do this for?
> > 
> > 
> > It looks like it, do you want me to include them all in this PR?
> 
> Not my initial idea but if it wouldn't make this PR too big, sure. Then rename the PR.
> If it would, then separate PRs

It might make it too long, looks like there are more than one other input components missing it.

zvonimirfras (approved):

cal-smith (approved):

PR #1756, opened 2021-03-16
fix(datepicker): range selection

### LINKED ISSUE #1738: Carbon date picker problem setting date range
## Detailed description

Using the calendar widget to set start and end dates in a range is buggy (see steps below)

no

yes -- date picker for range, Angular

First setting start date, then trying to set the end date fails.  I'd like it to succeed 😄 

Both Chrome and FF

```
    "carbon-components": "10.14.0",
    "carbon-components-angular": "^4.40.1",
```
Also, whatever is on https://angular.carbondesignsystem.com/?path=/story/components-date-picker--range

We are using this in an IBM product.

Can be run on https://angular.carbondesignsystem.com/?path=/story/components-date-picker--range

1. click on left calendar (start), select 1/1/2020
2. click outside the widget — now the start date is set, end date is empty
3. click on the right calendar (end), select 2/2/2021

-> the START date gets changed!  (rather than the end date)

## Additional information
- It appears that when the end date is not set, trying to set it via the calendar widget does the wrong thing
- Entering the dates by typing via the keyboard seems to work fine.

* Screenshots or code
At step 3:
image

* Notes

zvonimirfras:
closes #1738

This empties the fields if both sides of the range aren't selected, which mimics the behavior in reference implementation.

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -224,6 +224,13 @@ export class DatePicker implements
 			// This makes sure that the `flatpickrInstance selectedDates` are in sync with the values of
 			// the inputs when the calendar closes.
 			if (this.range && this.flatpickrInstance) {
+				if (this.flatpickrInstance.selectedDates.length !== 2) {
+					// we could `this.flatpickrInstance.clear()` but it insists on opening the second picker
+					// in some cases, so instead we do this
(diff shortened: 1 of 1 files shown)


cal-smith (approved):
Praise flatpickr magic! ✨

Donisius (approved):

PR #1757, opened 2021-03-16
fix(modal): scroll auto detection

KTruong88:

there is a bug in modals where the .bx--modal-content--overflow-indicator appears when there is no scroll on the element due to a float returned by getBoundingClientRect().height that gets rounded when clicking on the text.

Screen Shot 2021-03-16 at 1 53 05 PM
Screen Shot 2021-03-16 at 1 53 14 PM

#### Changelog

**New**

since getBoundingClientRect().height returns a float, this can cause a bug when it is technically < modalContent.scrollHeight but the difference does not cause a scroll bar to appear. 

added Math.round() to the value returned by the modalContent.getBoundingClientRect().height to prevent the error when comparing against scrollHeight which returns an integer.

### DIFF
--- a/src/modal/modal.component.ts
+++ b/src/modal/modal.component.ts
@@ -217,7 +217,8 @@ export class Modal implements AfterViewInit, OnChanges {
 	get shouldShowScrollbar() {
 		const modalContent = this.modal ? this.modal.nativeElement.querySelector(".bx--modal-content") : null;
 		if (modalContent) {
-			const modalContentHeight = Math.round(modalContent.getBoundingClientRect().height); // get rounded value from height to match integer returned from scrollHeight
+			// get rounded value from height to match integer returned from scrollHeight
+			const modalContentHeight = Math.round(modalContent.getBoundingClientRect().height);
 			const modalContentScrollHeight = modalContent.scrollHeight;
 			return modalContentScrollHeight > modalContentHeight;
 		} else {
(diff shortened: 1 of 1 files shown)


src/modal/modal.component.ts
```diff
@@ -217,7 +217,7 @@ export class Modal implements AfterViewInit, OnChanges {
 	get shouldShowScrollbar() {
 		const modalContent = this.modal ? this.modal.nativeElement.querySelector(".bx--modal-content") : null;
 		if (modalContent) {
-			const modalContentHeight = modalContent.getBoundingClientRect().height;
+			const modalContentHeight = Math.round(modalContent.getBoundingClientRect().height); // get rounded value from height to match integer returned from scrollHeight
```
zvonimirfras, line 220:
```suggestion
			// get rounded value from height to match integer returned from scrollHeight
			const modalContentHeight = Math.round(modalContent.getBoundingClientRect().height);
```

gotta do it in separate lines cause the line gets too long otherwise and linter complains.

zvonimirfras (changes requested):

src/modal/modal.component.ts
zvonimirfras: Kudos for figuring it out, tho, this one appears super randomly and I was never able to replicate it.

zvonimirfras (approved):
Whoever merges this, please create a `fix(modal): scroll auto detection` titled commit to trigger a new version.

cal-smith (approved):
This is such a wild bug 😱 Awesome work debugging this and sorting it out!

KTruong88:
@zvonimirfras @cal-smith thanks guys! glad to contribute finally :)

PR #1762, opened 2021-03-18
fix(checkbox): Avoid double checkbox classes

yinglejia:

PR https://github.com/IBM/carbon-components-angular/pull/1337 added `<div class="bx--form-item bx--checkbox-wrapper">` container. However, it didn't remove the checkbox classes added on the host element. Therefore, we ended up with double checkbox classes:

image

This PR removes the checkbox classes on the host element:

image

Note with this fix, the `inline` input property is no longer being used anymore. Maybe we should mark that as obsoleted?

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}

### DIFF
--- a/src/checkbox/checkbox.component.ts
+++ b/src/checkbox/checkbox.component.ts
@@ -217,13 +217,6 @@ export class Checkbox implements ControlValueAccessor, AfterViewInit {
 		return this._checked;
 	}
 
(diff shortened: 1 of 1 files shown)


zvonimirfras (approved):
Makes sense to me, thanks!

cal-smith (approved):
Cool :+1: Thanks for getting this done!

PR #1774, opened 2021-03-31
feat(datepicker): add ControlValueAccessor method for disable

### LINKED ISSUE #1771: Can not disable the carbon date picker component when using reactive forms.
Can not disable the carbon date picker component when using reactive forms. 

> I have a date picker component and before I implemented reactive forms I was able to add the disable attribute directly in the html and it worked like that, but I implemented reactive forms and I was getting and error saying I have to disable it different since I'm using reactive forms, I found the way to do this and I can make a text input component to disable but it does not works for the date picker. 

> I think this is a bug?

> Yes this issue is related specifically to the date picker component. 

> I expect that the user can't change the value from the date picker, but user can still click see the calendar and change the date. 

> I'm using Google Chrome. 

> I'm using v4 of carbon. 

Screen Shot 2021-03-30 at 8 55 56 AM
Screen Shot 2021-03-30 at 8 55 26 AM
Screen Shot 2021-03-30 at 8 55 14 AM

Donisius:
Closes #1771

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -376,6 +376,17 @@ export class DatePicker implements
 		});
 	}
 
+	/**
+	 * `ControlValueAccessor` method to programmatically disable the DatePicker.
+	 *
+	 * ex: `this.formGroup.get("myDatePicker").disable();`
+	 *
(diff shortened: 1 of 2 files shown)


cal-smith (approved):
Awesome 🎉

PR #1777, opened 2021-04-02
Add support for IME input in Safari

elycheea:
Observed an issue in Safari when using IME composition. First reported by Japanese input users and confirmed with Chinese input. Screenshots show IME input with traditional Chinese (zhuyin).

**Original on Storybook**
2021-04-02 at 09 50 10 IME safari

**Correct input**
2021-04-02 at 09 52 02 IME safari fix

Added a check for `isComposing` to prevent Safari from overwriting the search value with IME text, and verified with traditional Chinese in latest versions of Safari, Chrome, Edge, and Firefox.

#### Changelog

**New**

* Adds support for IME input in Safari.

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}

### DIFF
--- a/src/search/search.component.ts
+++ b/src/search/search.component.ts
@@ -181,9 +181,10 @@ export class Search implements ControlValueAccessor {
 	 * @param search The input text.
 	 */
 	onSearch(search: string) {
-		if (!this.isComposing) { // check for IME use
-			this.value = search;
+		if (this.isComposing) { // check for IME use
+			return;
 		}
+		this.value = search;
 		this.doValueChange();
 	}
 
--- a/src/search/search.component.ts
+++ b/src/search/search.component.ts
@@ -252,6 +252,6 @@ export class Search implements ControlValueAccessor {
 	@HostListener("compositionend", ["$event"])
 	compositionEnd(event) {
 		this.isComposing = false;
-		this.onSearch(event.data);
+		this.onSearch(this.value + event.data);
 	}
 }
--- a/src/search/search.component.ts
+++ b/src/search/search.component.ts
@@ -128,6 +128,12 @@ export class Search implements ControlValueAccessor {
 	// @ts-ignore
 	@ViewChild("input", { static: false }) inputRef: ElementRef;
 
+	/**
+	 * Sets `true` when composing text via IME.
+	 */
+	isComposing = false;
+
+
 	protected _size: "sm" | "md" | "xl" = "md";
 
 	/**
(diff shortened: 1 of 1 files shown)


src/search/search.component.ts
```diff
@@ -128,6 +128,12 @@ export class Search implements ControlValueAccessor {
 	// @ts-ignore
 	@ViewChild("input", { static: false }) inputRef: ElementRef;
 
+	/**
+	 * Sets `true` when composing text via IME.
+	 */
+	isComposing = false;
+
```
zvonimirfras, line 135: Should be protected unless there's a specific reason for it being public.

```suggestion
	protected isComposing = false;
```

src/search/search.component.ts
```diff
@@ -176,7 +182,9 @@ export class Search implements ControlValueAccessor {
 	 * @param search The input text.
 	 */
 	onSearch(search: string) {
-		this.value = search;
+		if (!this.isComposing) { // check for IME use
+			this.value = search;
+		}
 		this.doValueChange();
```
zvonimirfras, line 188: Do we need to do `this.doValueChange()` if it's composing? If not, let's guard it:
```suggestion
		if (this.isComposing) { // check for IME use
			return;
		}
		this.value = search;
		this.doValueChange();
```

If we do, it's fine.

zvonimirfras (changes requested):
Thanks for doing this, Elysia!
Just a few potential changes :)

src/search/search.component.ts
elycheea: Double checked and seems like guarding it still works! 👍

elycheea:
Just changed this back to draft after realizing the current approach to `compositionEnd` conveniently breaks for Korean which ends composition after each character ... I’m technically on vacation so might be a few days before I get a chance to debug it fully!

elycheea:
Thankfully a minor change addressed it for Korean — just appended the current data/value from `compositionEnd` to any existing search terms.

@kwak123 also noticed that the same issue could Chinese (and presumably Japanese) compositionEnd data to overwrite existing search values if someone ends composition and then restarts in the same field. Appending composition to existing search should now work for multilingual searches too and not replace the whole search term with only the composition characters.

Retested composition for both Korean and Chinese (also mixed with English) in Safari, Firefox, and Chrome.

cal-smith (approved):
This is awesome! Thanks for taking the time to implement this 😁

PR #1778, opened 2021-04-02
fix(select): fix reactive form functionality and allow selected to be initialized through value and options tags

### LINKED ISSUE #1776: ibm-select doesn't follow ngModel
## Detailed description

Using ngModel to initialise selected option for ibm-select doesn't update the component.
I see the issue has been reported previously and marked as resolved 
for example, https://github.com/IBM/carbon-components-angular/issues/1536

Bug

ibm-select

change selected option programatically

latest Chrome/Safari

 "carbon-components": "^10.31.0",
    "carbon-components-angular": "^4.46.0",

```
<div style="width: 165px;">
  <ibm-select [(ngModel)]="selected">
    <option value="option1">Option 1</option>
    <option value="option2">Option 2</option>
    <option value="option3">Option 3</option>
  </ibm-select>
  <br />
  <span>Selected: {{ selected }}</span>
</div>
```

Sandbox https://codesandbox.io/s/carbon-components-angular-forked-0y7q2?file=/src/app/app.component.html:0-275
image

Donisius:
Closes #1776

We cannot use the `value` property on select since default selected values through options tags would be ignored if a `value` is not set by the caller. #1754 addresses this but using an `attribute` means that the `attribute value` cannot change after initialization, which makes setting values through reactive forms not possible.

### DIFF
--- a/src/select/select.component.ts
+++ b/src/select/select.component.ts
@@ -207,12 +207,12 @@ export class Select implements ControlValueAccessor, AfterViewInit {
 
 	ngAfterViewInit() {
 		if (
-			this._value !== undefined &&
+			this.value !== undefined &&
 			this.value !== null &&
 			this.select &&
-			this.select.nativeElement.value !== this._value
+			this.select.nativeElement.value !== this.value
 		) {
-			this.select.nativeElement.value = this._value;
+			this.select.nativeElement.value = this.value;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


zvonimirfras (changes requested):

cal-smith (approved):
LGTM!

maksym-plotnikov:
Can You check CI errors?

cal-smith:
> Can You check CI errors?

Looks like the GitHub backend borked. Just resent the CodeQL jobs.

PR #1781, opened 2021-04-04
fix cannot read 'instance' of null when closing tooltip on weak devices

Santalov:
No related issue

Fix error "TypeError: Cannot read property 'instance' of null", that appears when two or more close events occur on tooltip simultaniously and device handles them with significant delay.

#### Changelog

**Changed**

* fix error with tooltip close events


zvonimirfras (changes requested):
Interesting. It looks like the whole point of the existing `if` is to fix this problem. Could we just ditch the outer if then and only check it in setTimeout?

Santalov:
> Interesting. It looks like the whole point of the existing `if` is to fix this problem. Could we just ditch the outer if then and only check it in setTimeout?

Done

zvonimirfras (approved):
👍

cal-smith (approved):
Interesting 🤔

PR #1785, opened 2021-04-06
feat(label): Make labelInputId an @Input

ozatici:
ibm-label doesn't support the use of "for" as an input, which is an accessibility violation - add it as an input rather than auto-generating it.

### DIFF
--- a/src/input/label.component.ts
+++ b/src/input/label.component.ts
@@ -38,7 +38,7 @@ import { TextArea } from "./text-area.directive";
 	selector: "ibm-label",
 	template: `
 		<label
-			[for]="labelInputID"
+			[for]="for"
 			[attr.aria-label]="ariaLabel"
 			class="bx--label"
 			[ngClass]="{
@@ -87,15 +87,14 @@ export class Label implements AfterContentInit, AfterViewInit {
 	 */
 	static labelCounter = 0;
 	/**
-	 * The id of the input item associated with the `Label`. This value is also used to associate the `Label` with
-	 * its input counterpart through the 'for' attribute.
-	 */
-	labelInputID = "ibm-label-" + Label.labelCounter;
-	/**
 	 * The class of the wrapper
 	 */
 	wrapperClass = "bx--text-input__field-wrapper";
-
+ 	/**
+        * The id of the input item associated with the `Label`. This value is also used to associate the `Label` with
+        * its input counterpart through the 'for' attribute.
+     */
+  	@Input() for: string;
 	/**
 	 * State of the `Label` will determine the styles applied.
 	 */
@@ -142,6 +141,9 @@ export class Label implements AfterContentInit, AfterViewInit {
 	 */
 	constructor() {
 		Label.labelCounter++;
+		if (this.for === undefined) {
+            this.for = "ibm-label-" + Label.labelCounter;
+        }
 	}
 
 	/**
--- a/src/input/label.component.ts
+++ b/src/input/label.component.ts
@@ -38,7 +38,7 @@ import { TextArea } from "./text-area.directive";
 	selector: "ibm-label",
 	template: `
 		<label
-			[for]="for"
+			[for]="labelInputID"
 			[attr.aria-label]="ariaLabel"
 			class="bx--label"
 			[ngClass]="{
--- a/src/input/label.component.ts
+++ b/src/input/label.component.ts
@@ -90,11 +90,11 @@ export class Label implements AfterContentInit, AfterViewInit {
 	 * The class of the wrapper
 	 */
 	wrapperClass = "bx--text-input__field-wrapper";
- 	/**
-        * The id of the input item associated with the `Label`. This value is also used to associate the `Label` with
-        * its input counterpart through the 'for' attribute.
-     */
-  	@Input() for: string;
+	/**
+	 * The id of the input item associated with the `Label`. This value is also used to associate the `Label` with
+	 * its input counterpart through the 'for' attribute.
+	*/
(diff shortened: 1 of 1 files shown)


zvonimirfras (changes requested):
Thanks for the effort, Sarah!
`ibm-label` doesn't need to support `for` attribute, `label` needs to have one, and it does have an auto-generated one. Yes, it would be better if it was customizable with a sensible default.
Instead of adding a `for` input to `ibm-label`, which would make DOM have more than one element with the same `for`, how about we make `labelInputId` an input?

```
	@Input labelInputID = "ibm-label-" + Label.labelCounter;
```

Is there a scenario which can't be solved with that?

ozatici:
How is it possible to make the label match the id of the input it corresponds to if it is auto generated? I need to be able to modify the for of label so that it matches the id of the input box it describes, to match accessibility standards. 

The point of PR wasn't to give a labelInputId to ibm-label, it was to be able to control what the 'for' of label is so that i can modify it to match the input it describes. 


This is what the accessibility flag is identifying as a violation - we have a lot of these coming up. If there is a workaround to fix this violation, I'm open to that as well

Screen Shot 2021-04-07 at 14 49 47

zvonimirfras:
> How is it possible to make the label's for match the id of the input it corresponds to if it is auto-generated? I need to be able to modify the for of the label so that it matches the id of the input box it describes, to meet accessibility standards.
> 
> The point of PR wasn't to give a labelInputId to ibm-label that is modifiable, it was to be able to control what the 'for' of label is so that I can modify it to match the id of the input it describes.
> 
> This is what the accessibility flag is identifying as a violation - we have a lot of these coming up. If there is a workaround to fix this violation, I'm open to that as well.
> 

Well, if `labelInputId` is an `@Input`, developer could provide the value for it, which is then passed down to label's `for` attribute.

By assigning a default value, as in my previous comment, if developer doesn't provide the value, the component would pass the generated id to the label's `for` attribute.

Note that if any of the supported input elements (`input,textarea,div`) are children of `ibm-label`, then developers don't even need to assign anything manually (since the generated id will be used) and won't get an accessibility violation.

ozatici:
> Note that if any of the supported input elements (`input,textarea,div`) are children of `ibm-label`, then developers don't even need to assign anything manually (since the generated id will be used) and won't get an accessibility violation.


You mean as it is now? Or if I add the labelInputId input? Right now i have input elements that are children of ibm-label and am getting accessibility violations..

zvonimirfras:
> > Note that if any of the supported input elements (`input,textarea,div`) are children of `ibm-label`, then developers don't even need to assign anything manually (since the generated id will be used) and won't get an accessibility violation.
> 
> You mean as it is now? Or if I add the labelInputId input? Right now i have input elements that are children of ibm-label and am getting accessibility violations..

From what I can see [in the demo](https://angular.carbondesignsystem.com/?path=/story/components-input--label), generated `for` attribute on the `label` element and `id` on the `input` element are correctly set, and there shouldn't be any violations around that.
If `labelInputId` was an input, it would allow developers to customize it and even use it with `input` elements placed elsewhere.

ozatici:
when i assign an id to the input, it doesn't match the randomly generated for on the label. this is my code:

```
<ibm-label [invalidText]="'Country required'"  id="overview_genInfo_countryLabel"> Country
            <input ibmText [value]="countryDesc" [disabled] = "true" [theme] = "'light'" id="overview_genInfo_countryInput" aria-labelledby="overview_genInfo_countryLabel" [epDisabler]="isQuoteUpdatable">
          </ibm-label>
```

I am getting this:

The value "ibm-label-70" of the 'for' attribute is not the 'id' of a valid <input> element

zvonimirfras:
> when i assign an id to the input, it doesn't match the randomly generated for on the label. this is my code:
> 
> ```
> <ibm-label [invalidText]="'Country required'"  id="overview_genInfo_countryLabel"> Country
>             <input ibmText [value]="countryDesc" [disabled] = "true" [theme] = "'light'" id="overview_genInfo_countryInput" aria-labelledby="overview_genInfo_countryLabel" [epDisabler]="isQuoteUpdatable">
>           </ibm-label>
> ```
> 
> I am getting this:
> 
> The value "ibm-label-70" of the 'for' attribute is not the 'id' of a valid element

Which is a good reason for giving developers control over `labelInputId`. Which this PR does now.
When this PR is merged, you won't have to explicitly assign that id to `input`, it should be enough to just give `labelInputId` to `ibm-label` and it should do the right thing.

ozatici:
> > when i assign an id to the input, it doesn't match the randomly generated for on the label. this is my code:
> > ```
> > <ibm-label [invalidText]="'Country required'"  id="overview_genInfo_countryLabel"> Country
> >             <input ibmText [value]="countryDesc" [disabled] = "true" [theme] = "'light'" id="overview_genInfo_countryInput" aria-labelledby="overview_genInfo_countryLabel" [epDisabler]="isQuoteUpdatable">
> >           </ibm-label>
> > ```
> > 
> > 
> > I am getting this:
> > The value "ibm-label-70" of the 'for' attribute is not the 'id' of a valid element
> 
> Which is a good reason for giving developers control over `labelInputId`. Which this PR does now.
> When this PR is merged, you won't have to explicitly assign that id to `input`, it should be enough to just give `labelInputId` to `ibm-label` and it should do the right thing.

Okay great, makes sense, thanks for being willing to comment back and forth, appreciate the feedback!!!

zvonimirfras (approved):

cal-smith (approved):
LGTM! Thanks for contributing this 🚀

PR #1819, opened 2021-05-13
feat(combobox): update html structure and classes

### LINKED ISSUE #1783: Dropdown and Combobox Multi-select tags don't match
The tag in the Combobox multi-select doesn't match the tag in the Dropdown multi-select.
If I compare angular components to [carbon dropdown style page](https://www.carbondesignsystem.com/components/dropdown/style/), it seems like the combobox should **not** have the bevel.
image

Donisius:
Closes #1783

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10171,9 +10171,9 @@
       "dev": true
     },
     "carbon-components": {
-      "version": "10.31.0",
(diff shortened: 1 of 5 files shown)


cal-smith (changes requested):
@Donisius looks like tests are failing ... seems to be expecting a different value for the label?

cal-smith (approved):

PR #1820, opened 2021-05-14
Update of bx--search-magnifier-icon in search component

PPPiQ:

Update of class for search component from 
bx--search-magnifier
to 
bx--search-magnifier-icon

acording to recomendation from carbon-designe under this merge: [https://github.com/carbon-design-system/carbon/pull/8101](https://github.com/carbon-design-system/carbon/pull/8101)

#### Changelog

**Changed**

search.component.html 
class: 
bx--search-magnifier
changed to: 
bx--search-magnifier-icon


src/search/search.component.html
```diff
@@ -34,12 +34,12 @@
...
 			(click)="openSearch()"
 			aria-label="Open search">
-			<svg ibmIcon="search" size="16" class="bx--search-magnifier"></svg>
+			<svg ibmIcon="search" size="16" class="bx--search-magnifier-icon"></svg>
 		</button>
 		<svg
 			ibmIcon="search"
 			*ngIf="tableSearch || !toolbar"
-			class="bx--search-magnifier"
+			class="bx--search-magnifier-icon"
```
cal-smith, line 42: So for me I get this: 
image

If I change it to this:
```suggestion
			class="bx--search-magnifier"
```
It fixes the position of the icon:
image

The toolbar search looks correct though :+1:

cal-smith (changes requested):
Thanks for taking the time to fix this! Just one suggestion, but otherwise looks good to me

dbrandl28:
Hi, just checking on the status of this issue - it is blocking me from upgrading.

src/search/search.component.html
subson: hey @cal-smith - So i tested it with 
```suggestion
			class="bx--search-magnifier"
```
but it keeps the magnifier outside of search box, but with 
```suggestion
			class="bx--search-magnifier-icon"
``` 
it works for me. 

search-bar

cal-smith: Interesting ... I think the `carbon-components` version needs to be updated in this PR then (I believe @PPPiQ is working on that)

cal-smith (approved):
Updating `carbon-components` did the trick, thanks for your patience @PPPiQ!

dbrandl28:
Thank you for fixing this!

PR #1837, opened 2021-05-28
chore: update README with support notices, add demo component notices to all app-* components

cal-smith:

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,38 @@ That's it! Now we can run `npm start` and start building out our application!
 
 [![Edit Carbon Components Angular](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/0129r494ql)
 
+### Support
+
+- For general usage questions, or other discussions visit our [Github Discussions](https://github.com/IBM/carbon-components-angular/discussions)
+- For things that are broken or need enhancement, [open an issue](https://github.com/IBM/carbon-components-angular/issues/new)
+- For quick usage questions:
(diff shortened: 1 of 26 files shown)


README.md
```diff
@@ -40,6 +40,38 @@ That's it! Now we can run `npm start` and start building out our application!
...
+| ------- | -- | -- | ------------------- |
+| 6       | ✅ | ❎ | ❎                   |
+| 7       | ✅ | ✅ | ❎                   |
+| 8       | ✅ | ✅ | ❎                   |
+| 9       | ❎ | ✅ | ❎                   |
+| 10      | ❎ | ✅ | ✅                   |
+| 11      | ❎ | ✅ | ✅                   |
+
+Carbon Components Angular version support matrix:
+| Carbon Components Angular version | Not supported | Community support | Active support |
```
zvonimirfras, line 62: "Not supported" column has a weird vibe to it because of double negation 🤔 
It makes it look like "no support" is a feature that v3 and v4 don't have 😄 

It might be cleaner to remove it completely (it should be enough that they don't have community or active support)
If necessary for clarity list it separately as
"Not supported: v1, v2"

zvonimirfras (approved):

README.md
cal-smith: I don't think there's a way to win with this 😆 If we don't have a statement of non-support people will ask is if they're supported, if we have this layout people may be confused ... I pulled the bullet up to directly below the table and bolded it, hopefully that's clear enough.

PR #1855, opened 2021-06-21
fix(slider): Fix accessibility check "1.1.1 Non-text Content: ViolationForm control with "slider" role has no associated label

stanislavgeorgiev:
Closes IBM/carbon-components-angular#1846

I think the `IBM Equal Access Accessibility Checker` might have some issues on its own and not detecting the already existing  `for` attribute on the main `<label>` element. This PR adds a safe-guard, which feels very much a redundant labeling, by adding an `aria-labeledby` attribute on the slider handle(s)

Slider:
image

Range:
image

### DIFF
--- a/src/slider/slider.component.ts
+++ b/src/slider/slider.component.ts
@@ -54,7 +54,7 @@ import { EventService } from "carbon-components-angular/utils";
 	selector: "ibm-slider",
 	template: `
 		<ng-container *ngIf="!skeleton; else skeletonTemplate">
-			<label *ngIf="label" [for]="id" class="bx--label">
+			<label *ngIf="label" [for]="id" [id]="labelId" class="bx--label">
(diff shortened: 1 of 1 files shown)


cal-smith (approved):
Thanks for figuring this out! Looks good to me, I'd be curious to know if this is intentional by the a11y team or if it's actually a bug

stanislavgeorgiev:
Thanks for merging this so quickly @cal-smith 
I plan on working on the other 2 issues coming from ADDI tomorrow 🤞 
https://github.com/IBM/carbon-components-angular/labels/ADDI

As for whether it's a bug or not, I can't tell for certain. `IBM Equal Access Accessibility Checker` is an IBM product and I think the AVT team just runs it and reports any issues reported by the tool. If the tool is faulty then I guess they would perform some manual checks - in the issue, they even suggested how to fix the issue (which didn't work for me but still)

PR #1858, opened 2021-06-23
fix(overflow-menu-pane): include floating menu direction data attribute

petersandor:
Adds missing data attribute so that Carbon styles apply on the menu pane and cover the gap between the trigger and the pane itself.

Jun-23-2021 13-05-55

### DIFF
--- a/src/dialog/overflow-menu/overflow-menu-custom-pane.component.ts
+++ b/src/dialog/overflow-menu/overflow-menu-custom-pane.component.ts
@@ -11,6 +11,7 @@ import { Dialog } from "../dialog.component";
 	template: `
 		<div
 			[attr.aria-label]="dialogConfig.menuLabel"
+			[attr.data-floating-menu-direction]="placement ? placement : null"
 			[ngClass]="{'bx--overflow-menu--flip': dialogConfig.flip}"
 			class="bx--overflow-menu-options bx--overflow-menu-options--open"
 			role="menu"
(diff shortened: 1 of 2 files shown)


cal-smith (approved):
Cool! Thanks for catching this 👍

petersandor:
@cal-smith Thanks for review. I have some concerns that I'd like to share before we merge this:

1. How is this change going to play with `gap` input which is inherited by `OverflowMenuDirective` from `DialogDirective`?
  https://github.com/IBM/carbon-components-angular/blob/ac404f5f34049379270b38d6be29aad51c9c87ab/src/dialog/dialog.directive.ts#L73-L76

2. When overflow menu is used in table toolbar, the default width of the trigger button is bigger than the gap covering pseudo-element (inherited fixed width from Carbon styles). So this has to be addressed individually with custom styles. 
  
    Screenshot 2021-06-25 at 13 53 22

    (red used to highlight the default width of the pseudo element)

cal-smith:
> 
> 
> @cal-smith Thanks for review. I have some concerns that I'd like to share before we merge this:
> 
>     1. How is this change going to play with `gap` input which is inherited by `OverflowMenuDirective` from `DialogDirective`?
>        https://github.com/IBM/carbon-components-angular/blob/ac404f5f34049379270b38d6be29aad51c9c87ab/src/dialog/dialog.directive.ts#L73-L76

I'm guessing this will end up looking a little odd ... But I don't expect many consumers are going to want to add a gap to the overflow menu ... and if they do they're probably already dealing with some pretty custom stuff, so the extra styling overhead shouldn't be a big deal. 

>     2. When overflow menu is used in table toolbar, the default width of the trigger button is bigger than the gap covering pseudo-element (inherited fixed width from Carbon styles). So this has to be addressed individually with custom styles.
>         Screenshot 2021-06-25 at 13 53 22
>        (red used to highlight the default width of the pseudo element)

This appears to be an issue in the React implementation as well ... I would think this would be something we'd want to fix upstream at the styling level. Might be worth opening an issue over in https://github.com/carbon-design-system/carbon to discuss

PR #1859, opened 2021-06-24
Remove @carbon/icons-angular completely from this repo

cduran:
This uses latest v11 version of @angular/cli
This is because the latest @angular/cli version is 12+ and @carbon/icons-angular complains about dependencies. It needs @angular/cli from v7 to v11.

And this fixes following the readme Getting started. Project now successfully builds.

#### Changelog

**New**

* README - Getting started now uses angular latest v11

**Changed**

* Changed the README.md

### DIFF
--- a/src/accordion/accordion-item.component.ts
+++ b/src/accordion/accordion-item.component.ts
@@ -7,8 +7,6 @@ import {
 	EventEmitter
 } from "@angular/core";
 
-import { IconModule } from "../icon/icon.module";
-
 @Component({
 	selector: "ibm-accordion-item",
 	template: `
--- a/src/accordion/accordion-item.component.ts
+++ b/src/accordion/accordion-item.component.ts
@@ -7,6 +7,8 @@ import {
 	EventEmitter
 } from "@angular/core";
 
+import { IconModule } from "../icon/icon.module";
+
 @Component({
 	selector: "ibm-accordion-item",
 	template: `
--- a/src/code-snippet/code-snippet.component.ts
+++ b/src/code-snippet/code-snippet.component.ts
@@ -8,6 +8,7 @@ import {
 } from "@angular/core";
 
 import { I18n } from "carbon-components-angular/i18n";
+import { IconModule } from "../icon/icon.module";
 
 export enum SnippetType {
 	single = "single",
--- a/src/code-snippet/code-snippet.component.ts
+++ b/src/code-snippet/code-snippet.component.ts
@@ -8,7 +8,6 @@ import {
 } from "@angular/core";
 
 import { I18n } from "carbon-components-angular/i18n";
-import { IconModule } from "../icon/icon.module";
 
 export enum SnippetType {
 	single = "single",
--- a/README.md
+++ b/README.md
@@ -21,9 +21,9 @@
 Assuming we're starting with a new @angular/cli project:
 
 ```shell
-$ npx @angular/cli@v11-lts new my-project --style=scss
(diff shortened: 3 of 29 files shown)


cal-smith (changes requested):
Thanks for catching this issue! Our approach for new versions of Angular is to add support for that version, rather than pinning our dependencies on an outdated version.

We should, however, update our getting started to remove `@carbon/icons-angular` - it's no longer the recommended solution for icons, and I wouldn't expect it to be updated to Angular 12. (Use the `IconService` and `IconDirective` with `@carbon/icons` instead :+1:)

cduran:
> Thanks for catching this issue! Our approach for new versions of Angular is to add support for that version, rather than pinning our dependencies on an outdated version.
> 
> We should, however, update our getting started to remove `@carbon/icons-angular` - it's no longer the recommended solution for icons, and I wouldn't expect it to be updated to Angular 12. (Use the `IconService` and `IconDirective` with `@carbon/icons` instead 👍)

Awesome understood. But now after doing some digging.

When you say to use `@carbon/icons` would be this one [https://github.com/carbon-design-system/carbon/tree/main/packages/icons](https://github.com/carbon-design-system/carbon/tree/main/packages/icons).
And in that repo, the [Usage](https://github.com/carbon-design-system/carbon/tree/main/packages/icons#usage) for Angular, it says to use `@carbon/icons-angular` [https://github.com/carbon-design-system/carbon-icons-angular](https://github.com/carbon-design-system/carbon-icons-angular).
So to clarify a bit we should stop using `@carbon/icons-angular` and just use the vanilla approach in our Angular app instead. Is that correct?

cduran:
Hello @cal-smith Can I remove the @carbon/icons-angular from every part of this repo? And use the correct form with @carbon/icons? It would be a big change thought...

cal-smith:
> Hello @cal-smith Can I remove the @carbon/icons-angular from every part of this repo? And use the correct form with @carbon/icons? It would be a big change thought...

That would be awesome!

I see that you've made a ton of changes already, I'll take some time to go through it 👍

src/accordion/accordion-item.component.ts
```diff
@@ -7,6 +7,8 @@ import {
 	EventEmitter
 } from "@angular/core";
 
+import { IconModule } from "../icon/icon.module";
+
```
cal-smith, line 11: I don't think we need this import here
```suggestion
```

src/code-snippet/code-snippet.component.ts
```diff
@@ -8,6 +8,7 @@ import {
 } from "@angular/core";
 
 import { I18n } from "carbon-components-angular/i18n";
+import { IconModule } from "../icon/icon.module";
```
cal-smith, line 11: Same comment here, I don't think we need to import the module in the component
```suggestion
import { I18n } from "carbon-components-angular/i18n";
```

cal-smith (changes requested):
Just a couple questions, but otherwise this looks great! Thanks for taking this on!

src/code-snippet/code-snippet.component.ts
cduran: Thanks for reviewing this. I accepted all of your suggestions.

cal-smith (approved):
LGTM! Sorry for the delay on my end, thanks for getting this updated 👍

cduran:
> LGTM! Sorry for the delay on my end, thanks for getting this updated 👍

No problem, thanks for taking time reviewing it.

PR #1862, opened 2021-06-24
fix(sidenav): Fixed an `undefined` tooltip always showing

stanislavgeorgiev:
Closes IBM/carbon-components-angular#1861

#### Changelog

* Allow `sidenav` to process the `title` property

**Changed**

* Remove `title` attribute if the `title` property is not set


cal-smith (approved):
LGTM! Thanks @stanislavgeorgiev 🎉

PR #1864, opened 2021-06-29
fix(combobox): don't trigger search for type 'single' on items change…

fionamurphy:
…s if there is a selected value

Closes IBM/carbon-components-angular#1733

When items are changed it is triggering a search and filtering items based on  the native input element value. For type "single" comboboxes this means the selected value is used to filter the list. Added a condition to prevent this from happening for comboboxes with type 'single' if there is a selected value. 

See scenario to reproduce in linked issue

#### Changelog

**Changed**

* ngOnChanges

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -454,10 +454,12 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
 	ngOnChanges(changes) {
 		if (changes.items) {
 			this.view.items = changes.items.currentValue;
-			// If new items are added into the combobox while there is search input,
-			// repeat the search.
-			this.onSearch(this.input.nativeElement.value, false);
 			this.updateSelected();
+			// If new items are added into the combobox while there is search input,
+			// repeat the search. Search should only trigger for type 'single' when there is no value selected.
+			if ( this.type === "multi" || (this.type === "single" && !this.selectedValue)) {
+				this.onSearch(this.input.nativeElement.value, false);
+			}
 		}
 	}
 


fionamurphy:
fyi @zvonimirfras @cal-smith This issue is a blocker for me at the moment for some dependant comboboxes ( the list available in  the next is dependant on the value of the previous).  Would either of you have time to review?
 
 Thanks
 Fiona

fionamurphy:
Bump ... @Donisius Should this have been fixed in https://github.com/IBM/carbon-components-angular/pull/1339

src/combobox/combobox.component.ts
```diff
@@ -454,10 +454,12 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
 	ngOnChanges(changes) {
 		if (changes.items) {
 			this.view.items = changes.items.currentValue;
-			// If new items are added into the combobox while there is search input,
-			// repeat the search.
-			this.onSearch(this.input.nativeElement.value, false);
 			this.updateSelected();
+			// If new items are added into the combobox while there is search input,
+			// repeat the search. Search should only trigger for type 'single' when there is no value selected.
+			if ( this.type === "multi" || (this.type === "single" && !this.selectedValue)) {
```
Donisius, line 460:
```suggestion
			if (this.type === "multi" || (this.type === "single" && !this.selectedValue)) {
```

Donisius (approved):
Sorry for the delay and thanks for taking this on! #1339 just prevents the search event from emitting when onSearch is triggered when new items are added to the list. It makes sense to me that the selected value shouldn't be used to filter the list for single select when new items are added 👍

src/combobox/combobox.component.ts
```diff
@@ -454,10 +454,12 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
...
+				this.onSearch(this.input.nativeElement.value, false);
+			}
```
Donisius, line 462: Should `updateSelected()` be called before `onSearch` in this case? Would updating the selected value after new items are added would make `this.selectedValue` always false, and make `onSearch` always trigger?

Donisius (changes requested):
Pulled the trigger a little too quick! Just one have question

src/combobox/combobox.component.ts
fionamurphy: Thanks for looking at this @Donisius 

The proposed change has the conditional `onSearch` moved after `updateSelected()` so that `this.selectedValue` is updated before the search is called - is that what you mean?

Donisius: @fionamurphy Yes, testing it locally with single select still triggered search when new items were added and a selected value was present. This is the example I used to test it:

```
@Component({
	selector: "app-dynamic-list-combobox",
	template: `
		<ibm-combo-box
			[(items)]="items"
			type="single">
			<ibm-dropdown-list></ibm-dropdown-list>
		</ibm-combo-box>
	`
})
class DynamicListComboBox implements AfterViewInit {
	items = [
		{
			content: "one"
		},
		{
			content: "two"
		},
		{
			content: "three"
		},
		{
			content: "four"
		}
	];

	ngAfterViewInit() {
		setInterval(() => {
			const newItems = JSON.parse(JSON.stringify(this.items));
			newItems.push({ content: `New ${newItems.length}` });
			this.items = newItems;
		}, 4000);
	}
}
```

fionamurphy: @Donisius `this.updateSelected()` is getting the selected value from `dropdown-list.component.ts`  `getSelected()`

```
getSelected(): ListItem[] {
		let selected = this.getListItems().filter(item => item.selected);
		if (selected.length === 0) {
			return [];
		}
		return selected;
	}
```

If no item has a truthy selected value this.selectedValue won't be set and the search is triggered.
I guess your example would work if the selected property was set on one of the items.

Donisius: @fionamurphy Okay, that makes sense to me!

Donisius (approved):

cal-smith (approved):
Makes sense to me ... Sorry for the delay!

PR #1868, opened 2021-07-16
fix(table-toolbar-search): fixing styling issue with latest carbon-components

Licen-it:
Closes IBM/carbon-components-angular#1856

Fixing styling issues with table-toolbar-search when using latest styling from carbon-components

#### Changelog

**Changed**

* moved relevant classes into search component
* show toolbar classes only when search is only in toolbar mode
* moved `expandable` property into search component and applied new classes from carbon styles
* fixed issue with close on blur when focusing a focusable element (eg a button) outside the search component which wasn't closing the expandable search

### DIFF
--- a/src/search/search.component.html
+++ b/src/search/search.component.html
@@ -6,11 +6,13 @@
 		'bx--search--xl': size === 'xl',
 		'bx--search--light': theme === 'light',
 		'bx--skeleton': skeleton,
-		'bx--toolbar-search': toolbar && !tableSearch,
-		'bx--toolbar-search--active': toolbar && !tableSearch && active,
-		'bx--toolbar-search-container-persistent': tableSearch && !toolbar,
-		'bx--toolbar-search-container-expandable': tableSearch && toolbar,
-		'bx--toolbar-search-container-active': tableSearch && toolbar && active
+		'bx--search--expandable': expandable && !tableSearch,
+		'bx--search--expanded': expandable && !tableSearch && active,
+		'bx--toolbar-search': toolbar && !expandable,
+		'bx--toolbar-search--active': toolbar && !expandable && active,
+		'bx--toolbar-search-container-persistent': tableSearch && !expandable,
+		'bx--toolbar-search-container-expandable': tableSearch && expandable,
+		'bx--toolbar-search-container-active': tableSearch && expandable && active
 	}"
 	role="search"
 	[attr.aria-label]="ariaLabel"
--- a/src/search/search.component.html
+++ b/src/search/search.component.html
@@ -6,8 +6,11 @@
 		'bx--search--xl': size === 'xl',
 		'bx--search--light': theme === 'light',
 		'bx--skeleton': skeleton,
-		'bx--toolbar-search': toolbar,
-		'bx--toolbar-search--active': toolbar && active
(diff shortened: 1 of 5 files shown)


Gabuardi (dismissed):

src/search/search.component.html
```diff
@@ -6,8 +6,11 @@
 		'bx--search--xl': size === 'xl',
 		'bx--search--light': theme === 'light',
 		'bx--skeleton': skeleton,
-		'bx--toolbar-search': toolbar,
-		'bx--toolbar-search--active': toolbar && active
+		'bx--toolbar-search': toolbar && !tableSearch,
+		'bx--toolbar-search--active': toolbar && !tableSearch && active,
+		'bx--toolbar-search-container-persistent': tableSearch && !toolbar,
+		'bx--toolbar-search-container-expandable': tableSearch && toolbar,
```
zvonimirfras, line 12: Should this be
```suggestion
		'bx--toolbar-search-container-expandable': tableSearch && toolbar && expandable,
```
otherwise expandable input has no effect

src/search/search.component.html
```diff
@@ -6,8 +6,11 @@
 		'bx--search--xl': size === 'xl',
 		'bx--search--light': theme === 'light',
 		'bx--skeleton': skeleton,
-		'bx--toolbar-search': toolbar,
-		'bx--toolbar-search--active': toolbar && active
+		'bx--toolbar-search': toolbar && !tableSearch,
+		'bx--toolbar-search--active': toolbar && !tableSearch && active,
+		'bx--toolbar-search-container-persistent': tableSearch && !toolbar,
```
zvonimirfras, line 11: Should this be
```suggestion
		'bx--toolbar-search-container-persistent': tableSearch && !toolbar && !expandable,
```

src/table/toolbar/table-toolbar-search.component.ts
```diff
@@ -22,16 +22,12 @@ import { NG_VALUE_ACCESSOR } from "@angular/forms";
...
-	@HostBinding("class.bx--toolbar-search-container-persistent") get persistentClass() { return !this.expandable; }
-
-	@HostBinding("class.bx--toolbar-search-container-active") get activeClass() {
-		return this.active && (this.value !== null || this.value !== "");
-	}
+	@HostBinding("class.bx--toolbar-content") hostClass = true;
 
 	ngOnInit() {
-		this.size = "sm";
+		this.size = "xl";
```
zvonimirfras, line 30: We need to change default size?

zvonimirfras (changes requested):

src/search/search.component.html
Licen-it: so expandable is not defined in the Search component, and the previous implementation was setting toolbar property to trigger the expandable in the table. (same answer for the comment below)

src/table/toolbar/table-toolbar-search.component.ts
Licen-it: i changed this as i noticed the react version is setting to `xl` while in the table, but i don't think it's making any difference there taking in consideration the css applied while in the table

cal-smith (approved):
Looks good to me, thanks! 🎉

zvonimirfras (approved):

PR #1875, opened 2021-07-27
fix(number-input): Setting ariaLabel as a directive for number input

luizamendes:
Adds ariaLabel as a directive for the number input component

#### Changelog

**New**

* Number input component now can receive ariaLabel directive, which value will be passed as aria-label to the input tag.


cal-smith (approved):
Looks good to me! Thanks for adding this 🥳

Issue #1865, opened 2021-07-05
Feature Request: Add an option to enable "title" attribute for Combobox dropdown items


namnguyen191:
## Detailed description

> Combobox dropdown items don't have the "title" attribute any more and there's no option to enable it

> This is a feature request

> This issue relate to the Combobox component

> I expect to have an option to enable to "title" attribute for the dropdown items

>  "carbon-components": "10.24.0" and "carbon-components-angular": "^4.26.1",

> UIHUB-CORE-UI

## Additional information

![title-attribute](https://user-images.githubusercontent.com/38139823/124510131-34c91a80-dda1-11eb-9b5a-1dca1ec4958f.png)

PR #1908, opened 2021-09-17
#1865 Added item title to dropdown-list.component

michalkmet:
#1865 
Added item title to dropdown list. Title were in the code before, but they were removed by this change: https://github.com/IBM/carbon-components-angular/commit/4e7a5b95bd6b4385327b1254dacf59b4027c720e#diff-706d42b7488bc5798a1d6c651e13f184b3e173958dd7f2ff8198bab0063ad818

I consider missing title attribute in <li> elements  as bug and our customers want it back.
I find issue related to this (#1865 ).


grenobnik:
If the title is hardwired to the item.content, I do not see how I would be able to disable it, if I do not want to have another overlay which repeats the content?

michalkmet:
Added option to disable title attribute.

PR #1920, opened 2021-10-15
Update button.types.ts: support button type 'normal'

MEKNI-D:

Button type 'normal' is not supported in the last version but present in the published storybook version of the library

#### Changelog

**Changed**

* include button type 'normal'


Donisius:
@MEKNI-D the button size is "normal" by default, no classes need to be applied. If `size` is set to anything other than `sm` | `field` | `lg` | `xl` then it will be regular size. You can set it to nothing if it is causing type issues in your build. Please feel free to reopen if I'm missing anything!

PR #1944, opened 2021-12-18
fix(Combobox): fix reactive form being dirty after setting value

Licen-it:
Closes IBM/carbon-components-angular#1943

Fixed combobox setting reactive form dirty after value being set by the form

#### Changelog

**Changed**

* Combobox no longer set reactive form control dirty after form's `setValue()`


Donisius (approved):
LGTM, thanks for taking this on!

PR #1948, opened 2021-12-24
feat: add integration for Angular 13

Donisius:


zvonimirfras (commented):
Is it ok to also update Angular version support matrix in the project readme?

PR #1954, opened 2022-01-13
fix(datepicker): reduce flatpickr reset on changes

yinglejia:

Currently, the date picker component resets the flatpickr instance on all input changes. This is unnecessary for input changes irrelevant to flatpickr, and can cause problems in some cases esp. for range date picker. For example, if the caller updates the validation status when the user selects a start date, then the flatpickr instance will be reset (destroyed) before the user can select an end date.

This PR limits the reset to relevant input changes only.

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}

### DIFF
--- a/src/datepicker/datepicker.component.ts
+++ b/src/datepicker/datepicker.component.ts
@@ -305,18 +305,17 @@ export class DatePicker implements
 
 	ngOnChanges(changes: SimpleChanges) {
 		// Reset the flatpickr instance on input changes that affect flatpickr.
-		const resetOnInputs = [
+		const flatpickrChangeKeys = [
 			"range",
 			"dateFormat",
 			"language",
 			"id",
 			"value",
-			"theme",
 			"plugins",
 			"flatpickrOptions"
 		];
 		const changeKeys = Object.keys(changes);
-		if (changeKeys.some(key => resetOnInputs.includes(key))) {
+		if (changeKeys.some(key => flatpickrChangeKeys.includes(key))) {
 			this.resetFlatpickrInstance(changes.value);
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/datepicker/datepicker.component.ts
```diff
@@ -298,13 +298,27 @@ export class DatePicker implements
...
-				this.resetFlackpickrInstance();
+				this.resetFlatpickrInstance();
 			});
 		}
 	}
 
 	ngOnChanges(changes: SimpleChanges) {
-		this.resetFlackpickrInstance(changes.value);
+		// Reset the flatpickr instance on input changes that affect flatpickr.
+		const resetOnInputs = [
```
Donisius, line 308:
```suggestion
		const flatpickerChangeKeys = [
```

src/datepicker/datepicker.component.ts
```diff
@@ -298,13 +298,27 @@ export class DatePicker implements
...
+			"dateFormat",
+			"language",
+			"id",
+			"value",
+			"theme",
+			"plugins",
+			"flatpickrOptions"
+		];
+		const changeKeys = Object.keys(changes);
+		if (changeKeys.some(key => resetOnInputs.includes(key))) {
```
Donisius, line 319:
```suggestion
		if (changeKeys.some(key => flatpickerChangeKeys.includes(key))) {
```

Donisius (changes requested):
Thanks for doing this @yinglejia! Looks good overall just a few nitpicks

src/datepicker/datepicker.component.ts
```diff
@@ -298,13 +298,27 @@ export class DatePicker implements
...
 	ngOnChanges(changes: SimpleChanges) {
-		this.resetFlackpickrInstance(changes.value);
+		// Reset the flatpickr instance on input changes that affect flatpickr.
+		const resetOnInputs = [
+			"range",
+			"dateFormat",
+			"language",
+			"id",
+			"value",
+			"theme",
```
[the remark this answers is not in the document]
yinglejia: Seems not needed, so removed.

Donisius (approved):

zvonimirfras (approved):

PR #1956, opened 2022-01-13
Fixing disabled not doing anything in RadioGroup

Licen-it:
Closes IBM/carbon-components-angular#1955

#### Changelog

**Changed**

* disable property is now working correctly
* disable through reactive form is now working as well


src/radio/radio-group.component.ts
```diff
@@ -152,7 +152,17 @@ export class RadioGroup implements AfterContentInit, AfterViewInit, ControlValue
...
+	@Input()
+	set disabled(disabled: boolean) {
+		this._disabled = disabled;
+		this.updateRadios();
+	}
+	/**
+	 * Returns the disabled value for the `RadioGroup`.
+	 */
+	get disabled(): boolean {
+		return this._disabled;
```
zvonimirfras, line 164: Where is `_disabled` defined and initialized?

zvonimirfras (changes requested):

src/radio/radio-group.component.ts
Licen-it: line 198...it was already there but never used

Donisius (approved):

zvonimirfras (approved):

PR #1969, opened 2022-01-14
fix(radio-group): fix reactive form set value not always working

Licen-it:
Closes IBM/carbon-components-angular#1968

RadioGroup wasn't clearing previous radio `checked` status which was resulting in updating the wrong selection at line 237 as multiple items in the array were having `checked` set to `true`

#### Changelog

**Changed**

* RadioGroup now clears previous `checked` status when a new radio in the group is selected


Donisius (approved):
LGTM, nice catch!

zvonimirfras (approved):

PR #2072, opened 2022-04-01
fix(TableModel): setting totalDataLength to 0 is no longer ignoring it

Licen-it:
Closes IBM/carbon-components-angular#2071

`0 || null` will always evaluate as `null` making the getter pickup the data length instead. Checking if the provided length is not a number would fix the issue.

#### Changelog

**Changed**

* setting TableModel.totalDataLength to 0 is no longer ignoring it


zvonimirfras (dismissed):

Licen-it:
@Donisius could you review as well please?

Donisius (approved):

PR #2128, opened 2022-06-02
fix(table): enable horizontal scroll in table

yinglejia:

Enable horizontal scroll in table to match Carbon React implementation. Refer Carbon React implementation: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/DataTable/Table.js#L34

image

image

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}


Akshat55:
Holding on merging - as we discuss with @carbon/styles team to add display block to `.bx--data-table-content` class. We can then use @hostbind the class - This would prevent unnecessary elements in template & make it easier to manage.

For the time being, you can add `.bx--data-table-content` class to ibm-table tag with display set to block as follows:

```html
<ibm-table class="bx--data-table-content" style="display: block">
     ....
</ibm-table>
```

yinglejia:
> Holding on merging - as we discuss with @carbon/styles team to add display block to `.bx--data-table-content` class. We can then use @hostbind the class - This would prevent unnecessary elements in template & make it easier to manage.
> 
> For the time being, you can add `.bx--data-table-content` class to ibm-table tag with display set to block as follows:
> 
> ```
> <ibm-table class="bx--data-table-content" style="display: block">
>      ....
> </ibm-table>
> ```

Thanks for the update, @Akshat55. Actually my initial attempt was to use @hostbind the class. I did not end up using that approach since the table component also @hostbind `bx--data-table_inner-container` for sticky header. So we would end up having both `.bx--data-table-content` and `bx--data-table_inner-container` on the host. I don't know Carbon team intends to support this or not. Adding display: block is the easy part.

src/table/table.component.ts
```diff
@@ -173,74 +173,76 @@ import { TableRowSize } from "./table.types";
...
-		ibmTable
-		[sortable]="sortable"
-		[noBorder]="noBorder"
-		[ngClass]="{'bx--data-table--sticky-header': stickyHeader}"
-		[size]="size"
-		[striped]="striped"
-		[skeleton]="skeleton">
-		<thead
-			ibmTableHead
+	<div class="bx--data-table-content">
```
Akshat55, line 176:
```suggestion
```

Let's apply the style directly to host & add styles.

src/table/table.component.ts
```diff
@@ -173,74 +173,76 @@ import { TableRowSize } from "./table.types";
...
+					<td class="table_end-indicator">
+						<h5>{{getEndOfDataText() | async}}</h5>
+						<button (click)="scrollToTop($event)" class="btn--secondary-sm">
+							{{getScrollTopText() | async}}
+						</button>
+					</td>
+				</tr>
+			</tfoot>
+		</table>
+	</div>
```
Akshat55, line 245:
```suggestion
```

src/table/table.component.ts
```diff
@@ -173,74 +173,76 @@ import { TableRowSize } from "./table.types";
...
 	`
```
Akshat55, line 246:
```suggestion
	`,
	styles: [`
		:host {
			display: block;
		}
	`]
```

src/table/table.component.ts
```diff
@@ -173,74 +173,76 @@ import { TableRowSize } from "./table.types";
...
 })
 export class Table implements AfterViewInit, OnDestroy {
```
Akshat55, line 248: Where best fits in class, add @HostBinding as seen below. Run npm run lint:fix to ensure the order is corect.

```suggestion
export class Table implements AfterViewInit, OnDestroy {
	@HostBinding("class.bx--data-table-content") tableContent = true;
```

Akshat55 (changes requested):
Hi @yinglejia, Thanks for your contribution & sorry for the long delay!

We have decided to go with leaving the styles in components file in carbon v10 (CCA v4), but for v11 they will be added in @carbon/styles package.

That being said, I have also merged master into the branch so just a friendly reminder,  please git pull.

yinglejia:
New PR created: https://github.com/IBM/carbon-components-angular/pull/2235

PR #2214, opened 2022-08-02
fix(modal): fix rounding issue in shouldShowScrollbar

yinglejia:
`shouldShowScrollbar` uses `scrollHeight` to determine whether the modal content is scrollable or not. The problem is that `scrollHeight` returns a rounded value, and its implementation (round up or down) is browser-specific. For text elements, with the default Carbon styles, the modal content height is `28.02 px`. Different browsers may return different values. e.g.:

- Chrome seems to always return `28`.
- FireFox returns `28` for some languages (e.g. English), but returns `29` for some other languages (e.g. Chinese).

The previous implementation uses `Math.round` to round `getBoundingClientRect().height` before comparison. This results in false positive for some languages like Chinese. It seems we should use `Math.ceil` instead.

Before:

image

After:

image


Akshat55 (approved):

zvonimirfras (approved):
Good job catching this!

PR #2242, opened 2022-09-09
fix: Remove tabindex

makandre:
Closes IBM/carbon-components-angular#2184

~Change role from `search` to `searchbox`~

~I believe `searchbox` is the correct widget role based on https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#2._widget_roles~

Remove the unnecessary `tabindex`

#### Changelog

**New**

* {{new thing}}

**Changed**

* [src/search/search.component.html](https://github.com/IBM/carbon-components-angular/compare/master...makandre:carbon-components-angular:patch-2#diff-d18ddec221752adaa51aabfcf9d75769a9608dc390c37b3722eb87cb9433cbb6)

**Removed**

* {{removed thing}}

### DIFF
--- a/src/search/search.component.html
+++ b/src/search/search.component.html
@@ -14,7 +14,7 @@
 		'bx--toolbar-search-container-expandable': tableSearch && expandable,
 		'bx--toolbar-search-container-active': tableSearch && expandable && active
 	}"
-	role="search"
+	role="searchbox"
 	[attr.aria-label]="ariaLabel"
 	tabindex="0"
 	(click)="openSearch()">
--- a/src/search/search.component.html
+++ b/src/search/search.component.html
@@ -14,9 +14,8 @@
 		'bx--toolbar-search-container-expandable': tableSearch && expandable,
 		'bx--toolbar-search-container-active': tableSearch && expandable && active
 	}"
-	role="searchbox"
+	role="search"
 	[attr.aria-label]="ariaLabel"
-	tabindex="0"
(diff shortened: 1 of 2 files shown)


Dagdelo (dismissed):

src/search/search.component.html
```diff
@@ -14,7 +14,7 @@
 		'bx--toolbar-search-container-expandable': tableSearch && expandable,
 		'bx--toolbar-search-container-active': tableSearch && expandable && active
 	}"
-	role="search"
+	role="searchbox"
```
Akshat55, line 17: Seems like the accessibility error is caused by `tabindex`, not the role.

```suggestion
	role="search"
```

image

src/search/search.component.html
```diff
@@ -14,7 +14,7 @@
...
 	[attr.aria-label]="ariaLabel"
 	tabindex="0"
```
Akshat55, line 19: `div` should not have tabindex. 

```suggestion
```

Akshat55: We don't need tabindex here since `input` element will have it.

Akshat55 (changes requested):
Thanks for taking on this issue, @makandre! 

I've requested a few changes.

Also, for the new commit, please include `fix:` at the beginning!

Akshat55 (approved):

zvonimirfras (approved):

PR #2248, opened 2022-09-14
fix: wrap close svg icon in a button to fix accessibility

Akshat55:
Closes IBM/carbon-components-angular#2234

#### Changelog
* Removed click/keyup events from `span` element events bubbled from `warning--filled` icon as well. This is unexpected behavior.
* Wrapped `close` icon in a `button` element.
* Made  `checkmark-filled` icon tab-able instead of it's parent (`span`).


Akshat55:
Some additional information that I found interesting... when passing `aria-label` in getAttributes (icon.directive.ts line 70). It returns a role attribute, otherwise it does not.

zvonimirfras (approved):

PR #2254, opened 2022-09-15
fix: allow layer level to be overriden

Akshat55:
#### Changelog
* Allows users to override layer level at ANY level.


src/layer/layer.directive.ts
```diff
@@ -32,39 +47,42 @@ export class LayerDirective implements AfterContentInit {
 					if (layer === this) {
 						return;
 					}
-					layer.ibmLayer = this.ibmLayer + 1;
+					layer.layer = typeof layer._passedLevel === "number" ? layer.layer = layer._passedLevel : this.layer + 1;
```
zvonimirfras, line 50: Why do we need to check if `_passedLevel` is a number all the time?

Akshat55: `_passedLevel` is undefined (Line 78). It is only assigned a value if user passes a input property. 

So I can either check if it is undefined or it has a number. I opted for number since we expect a number.

zvonimirfras: and why does it need to be passed in setter?

PR #2263, opened 2022-09-21
Merge branch 'master' of https://github.com/IBM/carbon-components-angular into next

Akshat55:


Akshat55:
Merging since we are merging from master into next. Style classes have been updated to match v11 prefix along with minor typescript changes to simplify code (Like using optional chaining).

PR #2269, opened 2022-09-30
fix: Remove inline js from anchors

Akshat55:
Closes IBM/carbon-components-angular#2223

#### Changelog
* Remove inline JS from anchors to enable secure CSP headers
* Set default `_href` value to `#` and return false if `_href` value equals `#` on click.

### DIFF
--- a/src/ui-shell/header/header-menu.component.ts
+++ b/src/ui-shell/header/header-menu.component.ts
@@ -24,7 +24,7 @@ import { HeaderItemInterface } from "./header-navigation-items.interface";
 				tabindex="0"
 				aria-haspopup="true"
 				[attr.aria-expanded]="expanded"
-				(click)="navigate()">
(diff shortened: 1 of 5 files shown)


src/ui-shell/header/header-menu.component.ts
```diff
@@ -23,7 +23,8 @@ import { HeaderItemInterface } from "./header-navigation-items.interface";
 				[href]="href"
 				tabindex="0"
 				aria-haspopup="true"
-				[attr.aria-expanded]="expanded">
+				[attr.aria-expanded]="expanded"
+				(click)="navigate()">
```
Akshat55, line 27: Why are we using an anchor in header menu? I noticed that all of the other carbon libs are using `<a>` as well 🤔

zvonimirfras (approved):

PR #2272, opened 2022-10-03
fix: prioritize input/textarea over div when setting id

Akshat55:
Closes IBM/carbon-components-angular#2136

#### Changelog
* Prioritize input & textarea over div - this will ensure that the id is not assigned to the div if the textarea & input are wrapped within a div.

### DIFF
--- a/src/input/label.component.ts
+++ b/src/input/label.component.ts
@@ -169,9 +169,16 @@ export class Label implements AfterContentInit, AfterViewInit {
 	 */
 	ngAfterViewInit() {
 		if (this.wrapper) {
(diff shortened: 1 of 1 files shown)


src/input/label.component.ts
```diff
@@ -169,9 +169,16 @@ export class Label implements AfterContentInit, AfterViewInit {
 	 */
 	ngAfterViewInit() {
 		if (this.wrapper) {
-			const inputElement = this.wrapper.nativeElement.querySelector("input,textarea,div");
```
Akshat55, line 172: Why are we querying for a div anyways? I'm keeping it for the time being since it's considered a breaking change if I remove it.

zvonimirfras (approved):

PR #2274, opened 2022-10-04
fix(file-uploader): Ensure change event is fired when uploading the same file

### LINKED ISSUE #2268: change function in file upload component is not getting executed if same file is uploaded again
## Detailed description
I am using the **ibm-file-uploader** component in my app and I have implemented some validation for the uploaded file. There is scenario where I need the change function to get executed irrespective of the file name (same file or a different file) on upload.

I uploaded file and it was uploaded successfully. Than i went inside the file in my system make some syntax error issues and tried to upload the same file again. It should have thrown error but instead the upload function was not even called. 

General issue

Yes, https://github.com/IBM/carbon-components-angular/tree/88e4998fefb66714bfdbef3e45842510433ca1bc/src/file-uploader

I wanted to confirm if there is any way to call the change function on every upload even if the file is same as previous uploaded file.

Chrome

**"carbon-components": "^10.56.0"**

exequiel09:
Closes #2268 

Signed-off-by: Exequiel Ceasar Navarrete <dev@exequielnavarrete.me>

Make sure change event is fired when uploading the same file. Also make sure that the set data structure is properly updated.

#### Changelog

**New**

N/A

**Changed**

1. Make sure change event is fired when uploading the same file. Also make sure that the set data structure is properly updated.
2. Added tests for it

**Removed**

N/A

### DIFF
--- a/src/file-uploader/file-uploader.component.spec.ts
+++ b/src/file-uploader/file-uploader.component.spec.ts
@@ -1,6 +1,6 @@
 import { FormsModule } from "@angular/forms";
 import { TestBed } from "@angular/core/testing";
-import { Component } from "@angular/core";
+import { Component, ViewChild } from "@angular/core";
 import { ButtonModule } from "carbon-components-angular/button";
 import { LoadingModule } from "carbon-components-angular/loading";
 import { FileUploader } from "./file-uploader.component";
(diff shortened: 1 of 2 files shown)


exequiel09:
Closing for now. Will re-open once the real issue is confirmed.

PR #2275, opened 2022-10-05
fix: bind attributes to host instead of adding unnecessary elements

Akshat55:
Closes IBM/carbon-components-angular#2103

According to many accessibility checkers, `ul/ol` elements should only contain `li` items.

#### Changelog
* Updates templates to use correct list structure
  * Instead of having custom elements as children of `ul` or `ol`, they are now a child of a `div` with the role set to `list`.
  * `ul` and `ol` have the same display property of div & children display property has been updated to `list-item`.


src/ui-shell/header/header-navigation.component.ts
```diff
@@ -11,7 +11,7 @@ import { NavigationItem } from "./header-navigation-items.interface";
 	selector: "ibm-header-navigation",
 	template: `
 		<nav class="cds--header__nav" [attr.aria-label]="ariaLabel">
-			<ul class="cds--header__menu-bar" role="menubar">
+			<div class="cds--header__menu-bar" role="list">
```
Akshat55, line 14: Switch to menubar

src/ui-shell/sidenav/sidenav-item.component.ts
```diff
@@ -88,9 +95,11 @@ export class SideNavItem implements OnChanges {
 	 */
 	@Output() selected = new EventEmitter<boolean>();
 
+	@HostBinding("attr.role") role = "listitem";
+
 	protected _href = "javascript:void(0)";
 
-	constructor(protected domSanitizer: DomSanitizer, @Optional() protected router: Router) {}
+	constructor(protected domSanitizer: DomSanitizer, @Optional() protected router: Router) { }
```
Akshat55, line 102:
```suggestion
	constructor(protected domSanitizer: DomSanitizer, @Optional() protected router: Router) {}
```

src/ui-shell/header/header-navigation.component.ts
Akshat55: Not switching to menubar as the pattern is not [necessary for most sites that are styled to look like a menubar](https://www.w3.org/WAI/ARIA/apg/example-index/menubar/menubar-navigation). We have expandable flyouts or sections but these [submenus are indicated visually by an icon & have the appropriate `aria-expanded`](https://www.w3.org/WAI/tutorials/menus/flyout/).

zvonimirfras (approved):

PR #2278, opened 2022-10-06
fix: add aria-labelledby attr to buttons in tooltip

Akshat55:
#### Changelog
* If a button element exists in tooltip add the aria-labelledby attribute to it with ID used by tooltip.


src/tooltip/tooltip.component.ts
```diff
@@ -99,4 +102,14 @@ export class Tooltip extends PopoverContainer {
...
+
+	ngAfterViewInit() {
+		if (this.wrapper) {
+			// Buttons
+			const buttonElement = this.wrapper.nativeElement.querySelector("button");
+			if (buttonElement && !buttonElement.getAttribute("aria-labelledby")) {
+				buttonElement.setAttribute("aria-labelledby", this.id);
+			}
+		}
+	}
```
Akshat55, line 114: I'm doing some magic here, there is an easier way but it would require users to do the following:

```html
<!-- Setting `#tooltip` as a template reference variable to the component -->
<ibm-tooltip
			#tooltip
			[isOpen]="isOpen"
			[caret]="caret"
			[align]="align"
			(onOpen)="onOpen($event)"
			(onClose)="onClose($event)"
			(isOpenChange)="isOpenChange($event)"
			[description]="description">
			<!-- Retrieving the id & assigning it to the button -->
			<button type="button" class="tooltip-trigger" [attr.aria-labelledby]="tooltip.id">
				<svg preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 32 32">
					<path d="M26,4H6A2,2,0,0,0,4,6V26a2,2,0,0,0,2,2H26a2,2,0,0,0,2-2V6A2,2,0,0,0,26,4ZM6,26V6H26V26Z"></path>
				</svg>
			</button>
</ibm-tooltip>
```

src/tooltip/tooltip.component.ts
```diff
@@ -13,16 +16,14 @@ import { PopoverContainer } from "carbon-components-angular/popover";
 @Component({
 	selector: "ibm-tooltip",
 	template: `
-		<span
-			[attr.aria-labelledby]="this.isTemplate(this.description) ? id : null"
-			[attr.aria-describedby]="!this.isTemplate(this.description) ? id : null">
+		<span #contentWrapper>
 			<ng-content></ng-content>
 		</span>
```
Akshat55, line 21: Why I removed the attributes? If I have a button as a child, it should be assigned the `aria` attribute not the span (wrapper). So if it is just an SVG for example, it does NOT need to have an aria attribute 🤔 atleast according to Accessibility checker

src/tooltip/tooltip.component.ts
```diff
@@ -99,4 +102,13 @@ export class Tooltip extends PopoverContainer {
 	public isTemplate(value) {
 		return value instanceof TemplateRef;
 	}
+
+	ngAfterViewInit() {
```
Akshat55, line 106: Switch to ngOnChanges after testing with conditions

zvonimirfras (approved):

PR #2302, opened 2022-10-29
chore: Run build check in next

Akshat55:
* Runs build checks in next branch


zvonimirfras (commented):
Also disables  <12 from running on main?

zvonimirfras (approved):

PR #2305, opened 2022-11-04
chore: Add carbon DCO workflow

Akshat55:
# Part of migration from `IBM` org to `carbon-design-system`
* Adds Carbon DCO workflow
* Adds legal notice in `contribution.md`

### Todo
- [x] Add secrets to run workflow

`GITHUB_TOKEN` is generated every new JOB.


Akshat55 (commented):
Merging to test DCO with users that haven't already signed off

PR #2323, opened 2022-11-11
fix: Avoid null pointers by not using nativeElement

anemonetea:
Closes IBM/carbon-components-angular#2228

 - The error message was about the checkbox's `nativeElement` being `undefined` when accessed before view initialization, as seen in the screenshot and in the CodeSandBox when opened in Chrome. Although I wasn't able to reproduce this in a local Storybook, the error makes sense because `enable()`/`disable()` gets called in `ngOnInit` lifecycle hook, at which point there is no (fully-formed) DOM yet.
 - Accessing the nativeElement shouldn't be needed in this case -- it was patching a change detection bug.
 - The change detection issue became apparent when using the `disabled` field directly in the template, like `<p>{{disabled}</p>`, which wouldn't change in the template even though the method ran.

#### Changelog

**New**

Storybook example of the reactive checkbox being disabled by default.

**Changed**

How `ControlValueAccessor.setDisabledState()` is implemented in the checkbox component.

**Removed**

N/A


anemonetea:

Akshat55 (approved):

PR #2324, opened 2022-11-11
feat: adjusted table sorting behavior

llccing:
hi, 

thanks for your wonderful work, our team use carbon, and i found that table sort behavior is different from react or svelte carbon. so i think this should update to consistent.

thanks

#### Changelog

**Changed**

* table sorting behavior, add default sort


llccing:

llccing:

llccing:
ok，i see the test failed. because when i add originIdx, this property will always keep. i need a time to fixed it. may be when not sorted, the originIdx can be remove, use like `delete item.originIdx`. but when sorted, this should keep.

Akshat55:
Thanks for contributing, @llccing! 

> I found that table sort behavior is different from react or svelte carbon. so i think this should update to consistent.
How exactly is the sort behavior different? Would make it easier to review & suggest changes if we know the differences.

Akshat55:
My apologies, the formatting on my last message was incorrect and you may have missed my question.

> I found that table sort behavior is different from react or svelte carbon. so i think this should update to consistent.

How exactly is the sort behavior different? Would make it easier to review & suggest changes if we know the differences.

llccing:
when click head in react and svelet table, sorting is `ascending/descending/default`, `ascending/descending/default` ... 
but in angular, sorting is ascending/descending, `asceding/descending` ... , can not return to the default sorting.

and in this pr, i just add the default sorting. carbon angular will have the same behavior like ascending/descending/`default` ...

Akshat55:
@llccing Thanks for pointing that out & making this contribution, I totally missed that! 😃

src/table/table-item.class.ts
```diff
@@ -106,6 +106,11 @@ export class TableItem {
 	 */
 	colSpan = 1;
 
+	/**
+	 * The original index of data
+	 */
+	originIdx?: number;
```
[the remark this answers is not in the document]
Akshat55: We spell the entire word to make it easier to read.

src/table/table-model.class.ts
```diff
@@ -43,7 +43,7 @@ export class TableModel implements PaginationModel {
 			newData = [[]];
 		}
 
-		this._data = newData;
+				this._data = newData;
```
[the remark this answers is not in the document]
Akshat55: Should run `npm run lint:fix` before committing & pushing.

src/table/table-model.class.ts
```diff
@@ -546,6 +546,22 @@ export class TableModel implements PaginationModel {
...
+		this.pushRowStateToModelData();
+		this.data.sort((a, b) => a[0].originIdx - b[0].originIdx);
+		this.popRowStateFromModelData();
+		this.header.forEach(column => (column.sorted = false));
+
+		this.data.forEach(item => {
+			item.forEach(subItem => {
+				delete subItem.originIdx;
+			});
+		});
```
Akshat55, line 562: Curious, do we _NEED_ to delete the index attribute from the cell object?

src/table/table.component.ts
```diff
@@ -837,13 +837,25 @@ export class Table implements AfterViewInit, OnDestroy {
...
+				});
+			});
+			currentHeader.ascending = true;
+				this.model.sort(index);
+			} else if (currentHeader.sorted && currentHeader.ascending) {
+				currentHeader.ascending = false;
+				currentHeader.descending = true;
+				this.model.sort(index);
+			} else {
+				this.model.resetSort();
```
Akshat55, line 857:
```suggestion
			if (!currentHeader.sorted) {
				// add origin sort for newData
				this.model.data.forEach((rowItem, index) => {
					rowItem.forEach(colItem => {
						colItem.index = index;
					});
				});
				currentHeader.ascending = true;
				this.model.sort(index);
			} else if(currentHeader.sorted && currentHeader.ascending) {
				// if already sorted flip sorting direction
				currentHeader.descending = true;
				this.model.sort(index);
			} else {
				this.model.resetSort();
```

Akshat55: No need to set ascending to false since headerItem class uses `ascending` for both ascending (true) & descending (false). Instead of using `idx` just use index or `i`.

Akshat55 (changes requested):
Hi @llccing, I'm sorry for not reviewing this PR sooner. Thank you so much for bringing this too our attention & making this PR!

Once these issues are addressed I can take another look and merge the PR!

src/table/table-model.class.ts
```diff
@@ -546,6 +546,22 @@ export class TableModel implements PaginationModel {
 		this.header[index].sorted = true;
 	}
 
+	/**
+   * reset the original sort
+   */
+	resetSort() {
+		this.pushRowStateToModelData();
+		this.data.sort((a, b) => a[0].originIdx - b[0].originIdx);
```
Akshat55, line 554:
```suggestion
		this.data.sort((a, b) => a[0].index - b[0].index);
```

src/table/table-model.class.ts
```diff
@@ -546,6 +546,22 @@ export class TableModel implements PaginationModel {
...
+   */
+	resetSort() {
+		this.pushRowStateToModelData();
+		this.data.sort((a, b) => a[0].index - b[0].index);
+		this.popRowStateFromModelData();
+		this.header.forEach(column => (column.sorted = false));
+
+		this.data.forEach(item => {
+			item.forEach(subItem => {
+				delete subItem.originIdx;
```
Akshat55, line 560:
```suggestion
				delete subItem.index;
```

src/table/table-model.class.ts
llccing: yes, we need. otherwise, the test will failed.

llccing:
hi, please let me know, will this pr be merged?

llccing:
I am disappointed by the slow pace of community action.

Akshat55:
@llccing So sorry, this is definitely a mistake on our end. I have updated the repository watch status to get notified on ALL messages to ensure this doesn't happen again. The team has been going through some restructuring & has caused a delay on this Angular issue. 

The Data Table is a complex implementation that we need to tread carefully since there are other packages depending on this component, hence the slow response as we figure things out. We should've been keeping you posted on this.

Now I understand that your experience with contributing to this repo has been bad, but I strongly encourage YOU to reopen this PR if you would still like to contribute.

llccing:
Of course, our team will use carbon-component-angular for a long time, and I want this issue to be fixed. Also, we've spent some time dealing with this issue, and we both don't want this time to be wasted. Thank you for your feedback.

Akshat55:
Thank you for reopening, @llccing!

src/table/table-item.class.ts
```diff
@@ -106,6 +106,11 @@ export class TableItem {
 	 */
 	colSpan = 1;
 
+	/**
+	 * The original index of data
+	 */
+	index?: number;
+
```
Akshat55, line 113: We shouldn't apply an index to each table item as at that point, it becomes redundant. Ideally, we should store the index in `row`. 🤔

Akshat55:
Closing this PR.

I don't think this will be updated based on my recent comments requested changes - that's fine! We still wanted to release this feature, so we've cooked up PR #2622 which has took a more flexible approach and ensures there aren't any breaking changes.

Issue #2397, opened 2023-01-03
Time picker component missing properties


MoizMasud:
Time picker missing props:

- Placeholder
- Size
- readOnly
- hideLabel

Akshat55:
Time picker missing props:

- [x] Placeholder
- [x] Size
- [x] readOnly (This is v11 only)
- [x] hideLabel

PR #2409, opened 2023-01-10
feat: Add missing time picker attributes

Akshat55:
* Add ariaLabel Input property to time picker select
* Add size option to time picker
* Allow users to override max length
* Add option to hide label


MoizMasud:
https://github.com/carbon-design-system/carbon-components-angular/issues/2397

PR #2414, opened 2023-01-18
fix(ibm-dropdown-list): ensure ListItem.disabled is treated as a boolean by styling logic

anemonetea:
The disabled styling was treating ListItem.disabled input as a disabled attribute (i.e. present or not) even though the doc for ListItem says that it is a boolean. Now, ListItem.disabled input is treated as a boolean.

Add tests for disabling a list-item

Fix linting error in grid

Signed-off-by: anemonetea <laz1anastasiya@gmail.com>

#### Changelog

**New**

New tests for the presence of a disabled attribute on ibm-dropdown-list components when given an input of `items` with disabled attributes in 3 varieties -- `disabled: true`, `disabled: false`, `disabled: undefined`.

Added disabled examples to the dropdown storybook.

**Changed**

What value we expect the `item.disabled` input to have. It is now consistently treated as a boolean.
"With Observable items" storybook example now uses the same object editable as other examples.

### DIFF
--- a/src/dropdown/list/dropdown-list.component.ts
+++ b/src/dropdown/list/dropdown-list.component.ts
@@ -67,7 +67,7 @@ import { ScrollCustomEvent } from "./scroll-custom-event.interface";
 				[attr.aria-selected]="item.selected"
 				[id]="getItemId(i)"
 				[attr.title]=" showTitles ? item.content : null"
-				[attr.disabled]="item.disabled"
+				[attr.disabled]="item.disabled ? true : null"
 				[ngClass]="{
 					'bx--list-box__menu-item--active': item.selected,
 					'bx--list-box__menu-item--highlighted': highlightedItem === getItemId(i)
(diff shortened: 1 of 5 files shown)


src/dropdown/list/dropdown-list.component.ts
```diff
@@ -67,7 +67,7 @@ import { ScrollCustomEvent } from "./scroll-custom-event.interface";
 				[attr.aria-selected]="item.selected"
 				[id]="getItemId(i)"
 				[attr.title]=" showTitles ? item.content : null"
-				[attr.disabled]="item.disabled"
+				[attr.disabled]="item.disabled ? '' : undefined"
```
Akshat55, line 70:
```suggestion
				[attr.disabled]="item.disabled ? true : undefined"
```

This will make it easier to read. Other than that, looks good to me!

Akshat55 (changes requested):

src/dropdown/list/dropdown-list.component.ts
anemonetea: Ok. I see other examples in the code base using `[attr.disabled]="item.disabled ? true : null"`, so I changed it to that.

Akshat55: That's even better! 😄

Akshat55 (approved):

PR #2416, opened 2023-01-25
feat: Add Edit on Carbon UI builder buttons to storybook

MoizMasud:
Signed-off-by: Moiz Masud <moizkhan_007@hotmail.com>
Closes [carbon-design-system/carbon-components-angular#](https://github.com/carbon-design-system/carbon-components-angular/issues/2415)


Akshat55 (changes requested):
Should remove the link to the stories that aren't possible in UI builder. Examples of this would be toolbar search, expandable toolbar search, skeletons, etc.

For the time being it's best to add links to the `basic` stories. We can circle back in the future and the story links for the more complicated stories.

Akshat55 (changes requested):
Hey, can we also readd the links to the `Clickable`, `Selectable`, `Multi-select` & `Expandable` tiles? We have those as a component so we can add the link.

Akshat55 (approved):

zvonimirfras (approved):

Issue #2426, opened 2023-01-27
Combobox dropdown is closed when new items match current search string


SMassola:
## Detailed description

After a combobox receives a new list of items to display in its dropdown list, rather than the dropdown list being open, it is closed.

> Is this a feature request (new component, new icon), a bug, or a general issue?

bug

> Is this issue related to a specific component?

Combobox

> What did you expect to happen? What happened instead? What would you like to see changed?

I expect that the new items would show right away since there is a search string present in the combobox input (assuming the content of the items includes the current search string).

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

"carbon-components": "^10.24.0",
"carbon-components-angular": "^4.59.20",

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

UIHub / SCIS

## Steps to reproduce the issue

Behaviour can be observed in the [mock query search](https://angular.carbondesignsystem.com/?path=/story/components-combobox--mock-query-search) example for a combobox in the storybook

1. Since all the items passed to the combobox start with the word "Random", type in "R". When the combobox receives the incoming items which all include the letter "R", the dropdown list doesn't remain open.

## Additional information

* Screenshots or code
* Notes

Issue #802, opened 2019-10-03
Dropdown multi-select does not behave correctly when any option is disabled


calesce:
## Detailed description

> Describe in detail the issue you're having.

When using the multiselect dropdown and setting one of the options to `disabled`, selecting any of the non-disabled options does not toggle the item correctly (see detail in step 3 below)

> Is this a feature request (new component, new icon), a bug, or a general issue?

Bug

> Is this issue related to a specific component?

Dropdown

> What did you expect to happen? What happened instead? What would you like to see changed?

Each non-disabled option should be able to be toggled on and off without issue.

> What browser are you working in?

Chrome, Firefox

> What version of the Carbon Design System are you using?

"carbon-components": 10.6.0
"carbon-components-angular": 3.17.3

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Product within IBM Watson Health

## Steps to reproduce the issue

I can create a Codesandbox if necessary, but the issue can be demonstrated on the [storybook]((https://angular.carbondesignsystem.com/?path=/story/dropdown--multi-select))

1. Navigate to [Dropdown -> Multi-select](https://angular.carbondesignsystem.com/?path=/story/dropdown--multi-select) in the angular storybook
2. Go to "knobs" tab and update "items" field to be this (first option is disabled):
```json
[
  {
    "content": "one",
    "disabled": true
  },
  {
    "content": "two"
  },
  {
    "content": "three"
  },
  {
    "content": "four"
  }
]
```
3. Click around on options 2-4; you'll notice that the initial click doesn't select/deselect the item, and clicking on a _different_ item will toggle the previously-selected item.

PR #2427, opened 2023-01-27
fix(ibm-combo-box): show list of new items that match current search

SMassola:
- removed setTimeout from updateList method inside the dropdown list component which was delaying the execution of the closeDropdown method inside the combo box component until after the openDropdown method was called
- updated combo box component to filter dropdown list items by the current search string when new items are passed into the combo box component
- migrated tick call on the application reference from the dropdown component into the dropdown list component to trigger change detection when a dropdown list item is selected rather than waiting until the dropdown receives the new selection event

Closes carbon-design-system/carbon-components-angular#2426

### DIFF
--- a/src/dropdown/dropdown.component.ts
+++ b/src/dropdown/dropdown.component.ts
@@ -357,7 +357,6 @@ export class Dropdown implements OnInit, AfterContentInit, AfterViewInit, OnDest
 		protected elementRef: ElementRef,
 		protected i18n: I18n,
 		protected dropdownService: DropdownService,
-		protected ngZone: NgZone,
 		protected elementService: ElementService) {}
 
 	/**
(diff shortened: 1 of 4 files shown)


SMassola:

src/dropdown/dropdown.component.ts
```diff
@@ -387,36 +387,36 @@ export class Dropdown implements OnInit, AfterContentInit, AfterViewInit, OnDest
...
-				this.selected.emit(event);
-			}
-			// manually tick the app so the view picks up any changes
-			this.appRef.tick();
+				// only emit selected for "organic" selections
+				if (!isUpdate(event)) {
+					this.checkForReorder();
+					this.selected.emit(event);
+				}
+			});
```
Akshat55, line 419: Any specific reason why you switched from `ApplicationRef` to `NgZone`? Isn't `NgZone.Run` practically equivalent to ApplicationRef.tick() in this case?

View is a ContentChild & it's run _within_ Angular. We are subscribing to the select event & making any appropriate changes based on the stream of data.

Akshat55: To be honest, I don't think we even need `ApplicationRef.tick()` here.... I think it might be there to fix bug caused by one of the Angular versions since we support from 7 to 15...

Akshat55: >updated dropdown component to run code inside the Angular zone to force a change detection cycle to take place rather than calling tick method on the application reference as the tick call was throwing an error after the change to the dropdown list was introduced

That's interesting, what was the error thrown?

SMassola: Screenshot 2023-01-31 at 1 36 58 PM

This was the error.

I could have used `setTimeout` instead of `ngZone.run` but it felt counter-productive to use `setTimeout` when I'm removing `setTimeout` in another area of the code which is the cause of the bug.

SMassola: From the slack thread I opened in the carbon help channel:

> Basically, I had to find an alternate solution to the one introduced in [809](https://github.com/carbon-design-system/carbon-components-angular/pull/809) because the appRef.tick call was throwing an error after I introduced the change in the dropdown list.
>
> I will say that the solution added by 809 and even the solution I have replaced it with are unconventional and there might be a deeper issue with the implementation of the dropdown component that is being masked.
>
> I say that because the combobox component has almost identical code with the subscribe block of the dropdown list select event emitter subscription yet it doesn't need an appRef.tick call to update the view.
>
> For clarity, I've linked the select subscribe block of the [dropdown component](https://github.com/carbon-design-system/carbon-components-angular/blob/842778a25d2db17fc19655f949dde4e4c32f08c6/src/dropdown/dropdown.component.ts#L389) and the [combo box component](https://github.com/carbon-design-system/carbon-components-angular/blob/842778a25d2db17fc19655f949dde4e4c32f08c6/src/combobox/combobox.component.ts#L482).

Akshat55 (approved):

src/dropdown/dropdown.component.ts
```diff
@@ -12,8 +12,8 @@ import {
 	OnDestroy,
 	HostBinding,
 	TemplateRef,
-	ApplicationRef,
-	AfterViewInit
+	AfterViewInit,
+	NgZone
```
Akshat55, line 16:
```suggestion
	AfterViewInit
```

src/dropdown/dropdown.component.ts
```diff
@@ -357,7 +357,7 @@ export class Dropdown implements OnInit, AfterContentInit, AfterViewInit, OnDest
 		protected elementRef: ElementRef,
 		protected i18n: I18n,
 		protected dropdownService: DropdownService,
-		protected appRef: ApplicationRef,
+		protected ngZone: NgZone,
```
Akshat55, line 360:
```suggestion
```

src/dropdown/dropdown.component.ts
```diff
@@ -387,36 +387,36 @@ export class Dropdown implements OnInit, AfterContentInit, AfterViewInit, OnDest
 		const isUpdate = event => event && event.isUpdate;
 
 		this.view.select.subscribe(event => {
-			if (this.type === "single" && !isUpdate(event)) {
-				this.closeMenu();
-				if (event.item && event.item.selected) {
-					if (this.itemValueKey) {
-						this.propagateChange(event.item[this.itemValueKey]);
+			this.ngZone.run(() => {
```
Akshat55, line 390:
```suggestion
			setTimeout(() => {
```

Akshat55 (changes requested):

src/dropdown/dropdown.component.ts
SMassola: Unfortunately, using `setTimeout` does resolve the issue highlighted in https://github.com/carbon-design-system/carbon-components-angular/issues/802.

Akshat55 (approved):

zvonimirfras (approved):

Issue #2430, opened 2023-01-30
ColumnDirective not correctly applying bx--col class when no inputs are provided


Licen-it:
## Detailed description

> Describe in detail the issue you're having.

if the `ibmCol` directive is used without any input, the `ngOnChange` method is never called and the default class never applied resulting in wrong styling

> Is this a feature request (new component, new icon), a bug, or a general issue?

bug

> Is this issue related to a specific component?

ColumnDirective

> What did you expect to happen? What happened instead? What would you like to see changed?

the `bx--col` class should be applied in any case when no `columnNumbers` are passed

> What browser are you working in?

any

> What version of the Carbon Design System are you using?

latest

PR #2431, opened 2023-01-30
fix(ColumnDirective): fixed missing class if no inputs provided

Licen-it:
Closes carbon-design-system/carbon-components-angular#2430

`ngOnChanges` was never called if no inputs were passed

#### Changelog

**Changed**

* added `bx--col` by default
* added UTs to cover this scenario

@zvonimirfras


Licen-it:
recheck

zvonimirfras (approved):

Issue #1999, opened 2022-02-11
Theme input is missing from most of the tile components


Akshat55:
## Detailed description

> Describe in detail the issue you're having.

Option to toggle between the light and default theme is missing in some of the tiles. The following code is only available in default tile components but not clickable, expandable, etc.
```ts
	@HostBinding("class.bx--tile--light") get lightThemeEnabled() {
		return this.theme === "light";
	}

	@Input() theme: "light" | "dark" = "dark";
```

> Is this a feature request (new component, new icon), a bug, or a general issue?

General issue

> Is this issue related to a specific component?

Tiles

> What did you expect to happen? What happened instead? What would you like to see changed?

I expect the components to accept a `light` input attribute that would allow me to toggle between light & dark themes available for tile.

> What browser are you working in?

Firefox

> What version of the Carbon Design System are you using?

Latest - 4.57.6

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

NA

Akshat55:
Reference: https://react.carbondesignsystem.com/?path=/story/components-tile--clickable

All tile components in react have an option to switch to light theme.

PR #2451, opened 2023-02-09
feat: Add theme inputs

MoizMasud:
Closes https://github.com/carbon-design-system/carbon-components-angular/issues/1999

### DIFF
--- a/src/tiles/clickable-tile.component.ts
+++ b/src/tiles/clickable-tile.component.ts
@@ -35,6 +35,7 @@ import { Router } from "@angular/router";
 })
 export class ClickableTile {
 	@Input() theme: "light" | "dark" = "dark";
+
 	/**
 	 * Sets the `href` attribute on the `ibm-clickable-tile` element.
 	 */
--- a/src/tiles/clickable-tile.component.ts
+++ b/src/tiles/clickable-tile.component.ts
@@ -3,7 +3,8 @@ import {
 	Input,
 	Output,
 	EventEmitter,
-	Optional
+	Optional,
+	HostBinding
 } from "@angular/core";
 import { Router } from "@angular/router";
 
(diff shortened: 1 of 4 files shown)


src/tiles/clickable-tile.component.ts
```diff
@@ -33,6 +35,11 @@ import { Router } from "@angular/router";
 	</a>`
 })
 export class ClickableTile {
+	@HostBinding("class.bx--tile--light") get lightThemeEnabled() {
+		return this.theme === "light";
+	}
```
Akshat55, line 40: Do we need to assign the class to the host or the anchor link? Class is being assigned to the host & anchor link (see line 28).

Pretty sure you do not need to assign to the host.

Akshat55: Also if you don't need hostbinding after, please remove host import in line 7

src/tiles/expandable-tile.component.ts
```diff
@@ -39,6 +43,12 @@ export interface ExpandableTileTranslations {
 	`
 })
 export class ExpandableTile implements AfterContentInit {
+	@HostBinding("class.bx--tile--light") get lightThemeEnabled() {
+		return this.theme === "light";
+	}
```
Akshat55, line 48: Same thing as the clickable-tile here. Why do we need to assign the `bx--tile--light` class to the host & the button?

src/tiles/selection-tile.component.ts
```diff
@@ -45,6 +49,13 @@ export class SelectionTile implements AfterViewInit {
 	/**
 	 * The unique id for the input.
 	 */
+
+	@HostBinding("class.bx--tile--light") get lightThemeEnabled() {
+		return this.theme === "light";
+	}
```
Akshat55, line 55: Same as previous comments.

Akshat55 (changes requested):

Akshat55 (approved):

zvonimirfras (approved):

Issue #2453, opened 2023-02-11
Disable input is missing from some tile components 


MoizMasud:
Option to disable tiles is missing in clickable tile and selectable tile

PR #2454, opened 2023-02-11
feat: Add disable property to tiles

MoizMasud:
Signed-off-by: Moiz Masud <moizkhan_007@hotmail.com>

closes https://github.com/carbon-design-system/carbon-components-angular/issues/2453

### DIFF
--- a/src/tiles/clickable-tile.component.ts
+++ b/src/tiles/clickable-tile.component.ts
@@ -29,7 +29,7 @@ import { Router } from "@angular/router";
 			'bx--tile--disabled bx--link--disabled' : disabled
 		}"
 		tabindex="0"
-		(click)="disabled ? null : navigate($event)"
+		(click)="navigate($event)"
 		[attr.href]="disabled ? null : href"
 		[attr.target]="target"
 		[attr.aria-disabled]="disabled">
@@ -74,7 +74,7 @@ export class ClickableTile {
 	constructor(@Optional() protected router: Router) {}
 
 	navigate(event) {
-		if (this.router && this.route) {
+		if (this.router && this.route && !this.disabled) {
 			event.preventDefault();
 			const status = this.router.navigate(this.route, this.routeExtras);
 			this.navigation.emit(status);
--- a/src/tiles/clickable-tile.component.ts
+++ b/src/tiles/clickable-tile.component.ts
@@ -30,7 +30,7 @@ import { Router } from "@angular/router";
 		}"
 		tabindex="0"
 		(click)="!disabled ? navigate($event) : null"
-		[href]="href"
+		[attr.href]="!disabled ? href : null"
 		[attr.target]="target"
 		[attr.aria-disabled]="disabled">
 		<ng-content></ng-content>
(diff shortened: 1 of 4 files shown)


Akshat55:
Thanks @MoizMasud, could you please look into why the tests are failing?

MoizMasud:
> Thanks @MoizMasud, could you please look into why the tests are failing?

should be okay now!

Akshat55 (commented):
You'll need to add the disabled property to the anchor link & input as well. Try disabling the tiles & clicking on the `clickable-tile` or `selection-tile`.

src/tiles/clickable-tile.component.ts
```diff
@@ -24,10 +24,13 @@ import { Router } from "@angular/router";
...
-		[ngClass]="{'bx--tile--light': theme === 'light'}"
+		[ngClass]="{
+			'bx--tile--light': theme === 'light',
+			'bx--tile--disabled bx--link--disabled' : disabled
+		}"
 		tabindex="0"
-		(click)="navigate($event)"
-		[href]="href"
+		(click)="!disabled ? navigate($event) : null"
+		[attr.href]="!disabled ? href : null"
```
Akshat55, line 33:
```suggestion
		(click)="disabled ? null : navigate($event)"
		[attr.href]="disabled ? null : href"
```

src/tiles/selection-tile.component.ts
```diff
@@ -19,18 +19,20 @@ import { I18n } from "carbon-components-angular/i18n";
...
 				#input
 				tabindex="-1"
 				class="bx--tile-input"
 				[id]="id"
+				[disabled]="disabled"
 				[type]="(multiple ? 'checkbox': 'radio')"
 				[value]="value"
 				[name]="name"
-				(change)="onChange($event)"/>
+				(change)="!disabled ? onChange($event) : null"/>
```
Akshat55, line 35:
```suggestion
				(change)="disabled ? null : onChange($event)"/>
```

src/tiles/selection-tile.component.ts
```diff
@@ -19,18 +19,20 @@ import { I18n } from "carbon-components-angular/i18n";
 			[for]="id"
 			[ngClass]="{
 				'bx--tile--is-selected' : selected,
-				'bx--tile--light': theme === 'light'
+				'bx--tile--light': theme === 'light',
+				'bx--tile--disabled' : disabled
```
Akshat55, line 23: Please update **line 18** (tabindex), when disabled, it should be set to `null` removing the attribute to match React. Otherwise, users will still be able to `focus` the `disabled` tile.

image

Akshat55 (changes requested):

src/tiles/clickable-tile.component.ts
Akshat55: I don't see the point in doing an extra operation here.

src/tiles/selection-tile.component.ts
MoizMasud: Updated

Akshat55 (approved):

zvonimirfras (approved):

Issue #2459, opened 2023-02-16
Error NG0100 is triggered by Label component if id already exists on the child element


Licen-it:
## Detailed description

> Describe in detail the issue you're having.

Error `NG0100: Expression has changed after it was checked` is triggered by Label component if the input/textarea/div passed as child already defines an id

> Is this a feature request (new component, new icon), a bug, or a general issue?

bug

> Is this issue related to a specific component?

LabelComponent

> What did you expect to happen? What happened instead? What would you like to see changed?

No error should be triggered

> What browser are you working in?

any

> What version of the Carbon Design System are you using?

latest

![Screenshot](https://user-images.githubusercontent.com/35496522/219619874-0a11c00b-5dc0-4a51-af11-2c8b959969ec.png)

PR #2460, opened 2023-02-16
Fixed error NG0100 happening on LabelComponent

Licen-it:
Closes carbon-design-system/carbon-components-angular#2459

added `changeDetectorRef.detectChanges()` to trigger a re-evaluation of the changes. Suggested solution from angular is to use `ngAfterContentInit` instead of `ngAfterViewInit` (https://angular.io/errors/NG0100), but I believe afterViewInit was used to catch elements generated by child components that would only appear at that stage. @reviewers if you believe the suggested change from angular might not break current behaviour i'm happy to apply that change instead.

Screenshot of the error:
Screenshot

#### Changelog

**Changed**

* fixed NG0100 error happening when child nodes passed to LabelComponent already define an id

### DIFF
--- a/src/input/label.component.ts
+++ b/src/input/label.component.ts
@@ -7,7 +7,8 @@ import {
 	TemplateRef,
 	ViewChild,
 	ContentChild,
-	AfterContentInit
+	AfterContentInit,
+	ChangeDetectorRef
 } from "@angular/core";
 
 import { TextArea } from "./text-area.directive";
@@ -151,7 +152,7 @@ export class Label implements AfterContentInit, AfterViewInit {
 	/**
 	 * Creates an instance of Label.
 	 */
-	constructor() {
+	constructor(protected changeDetectorRef: ChangeDetectorRef) {
 		Label.labelCounter++;
 	}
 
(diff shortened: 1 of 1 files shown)


Akshat55:
Which version of Angular are you seeing this error on @Licen-it? I've tried to reproduce the error but haven't been able to. I've tried with Angular 13 & don't run into this error 🤔

Licen-it:
@Akshat55 (not sure why not seeing your comment here even if i got the email notification...anyway)...i'm using angular 14, but it should appear on 13 as well. Make sure that you are defining an id on the input/textarea/div you are inserting inside the label as it's only appearing when it's reusing the existing id for the `for=` property. Also make sure you are running in development mode as i believe it's suppressed when running in production mode

Akshat55:
I deleted the comment @Licen-it since I reproduced it - I was checking the wrong version of carbon angular. My fault! Sorry about that.

Akshat55 (approved):

Akshat55:
I've approved the fix!

>  but I believe afterViewInit was used to catch elements generated by child components that would only appear at that stage.

Yes, we are using AfterViewInit to catch `this.wrapper` (ViewChild)... We can alternatively use ElementRef? I'll experiment with that in V11.

zvonimirfras (approved):

Issue #2389, opened 2023-01-02
When you click on the Dropdown and keep going down the list with the arrow key in keyboard, the scrubber in scrollbar does not go down or moving to proper scroll position.


ravikumar-sundaram:
1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

* Screenshots or code
* Notes

PR #2471, opened 2023-03-02
fix: dropdown list keyboard scroll not working

rhythm55:
Closes: [#2389](https://github.com/carbon-design-system/carbon-components-angular/issues/2389)

Dropdown list was scrollable using mouse but with keyboard list was not scrolling. checked keyboard navigation of dropdown list `navigateList` and found out that upon navigation from arrow keys html elements were missing with scroll behavior. On adding `scrollIntoView` the list scroll is working as expected with keyboard.
Took dropdown list scroll behavior reference from - [carbon design react](https://react.carbondesignsystem.com/?path=/story/components-dropdown--default)

**Before**
issue

**After**
issue

### DIFF
--- a/src/dropdown/list/dropdown-list.component.ts
+++ b/src/dropdown/list/dropdown-list.component.ts
@@ -539,13 +539,13 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 			event.preventDefault();
 			if (event.key === "ArrowDown" || event.key === "Down") {
 				if (this.hasNextElement()) {
-					this.getNextElement();
+					this.getNextElement().scrollIntoView({behavior: 'smooth',block: 'end'});
 				} else {
 					this.blurIntent.emit("bottom");
 				}
 			} else if (event.key === "ArrowUp" || event.key === "Up") {
 				if (this.hasPrevElement()) {
-					this.getPrevElement();
+					this.getPrevElement().scrollIntoView({behavior: 'smooth'});
 				} else {
 					this.blurIntent.emit("top");
 				}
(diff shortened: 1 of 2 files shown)


rhythm55:

Akshat55:
Hey @rhythm55, could you please add additional list items to the story? This will make it easier to test out.

src/dropdown/list/dropdown-list.component.ts
```diff
@@ -539,13 +539,13 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 			event.preventDefault();
 			if (event.key === "ArrowDown" || event.key === "Down") {
 				if (this.hasNextElement()) {
-					this.getNextElement();
+					this.getNextElement().scrollIntoView({behavior: "smooth", block: "end"});
```
Akshat55, line 542:
```suggestion
					this.getNextElement().scrollIntoView({ block: "end" });
```

src/dropdown/list/dropdown-list.component.ts
```diff
@@ -539,13 +539,13 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
...
 				} else {
 					this.blurIntent.emit("bottom");
 				}
 			} else if (event.key === "ArrowUp" || event.key === "Up") {
 				if (this.hasPrevElement()) {
-					this.getPrevElement();
+					this.getPrevElement().scrollIntoView({behavior: "smooth"});
```
Akshat55, line 548:
```suggestion
					this.getPrevElement().scrollIntoView();
```

Akshat55 (changes requested):
Removing smooth scrolling behavior to match the React behavior. There _could_ also be accessibility concerns with smooth scrolling behavior, hence best to remove in this case.

zvonimirfras (approved):

Issue #2479, opened 2023-03-15
Progress Indicator: Design is different from the Carbon Components


stanislavgeorgiev:
## Detailed description

> Describe in detail the issue you're having.

This is how the [Carbon Progress Indicator](https://carbondesignsystem.com/components/progress-indicator/usage) is supposed to look like:

![image](https://user-images.githubusercontent.com/1253469/225424775-3ad7b986-7d88-4b79-9277-7362dd297c9d.png)

And this is how the [Angular version ](https://angular.carbondesignsystem.com/?path=/story/components-progress-indicator--basic)looks like:

![image](https://user-images.githubusercontent.com/1253469/225424901-0e05054f-f716-40ea-9876-d9d3bd7cc184.png)

Also note the additional defect of showing 2 icons in the error state:

![image](https://user-images.githubusercontent.com/1253469/225425210-63d4ec3b-5109-4992-baab-8c58c8d624ae.png)

> Is this a feature request (new component, new icon), a bug, or a general issue?

Bug

> Is this issue related to a specific component?

Progress Indicator

> What did you expect to happen? What happened instead? What would you like to see changed?

The Angular design should match the Carbon styling

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

4.63.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

ADDI

PR #2483, opened 2023-03-16
Issue 2479: Progress Indicator: Design is different from the Carbon Components

stanislavgeorgiev:
Closes carbon-design-system/carbon-components-angular#2479

#### Changelog

**Changed**
Before: 
image

After: Update icons to the [Carbon v10 Design](https://v10.carbondesignsystem.com/components/progress-indicator/usage/)
image

Before:
image

After: Fix doubled icons for the "error" state
image

### DIFF
--- a/src/progress-indicator/progress-indicator.component.ts
+++ b/src/progress-indicator/progress-indicator.component.ts
@@ -30,40 +30,11 @@ import { Step } from "./progress-indicator-step.interface";
 			[ngClass]="{'bx--progress-step--disabled' : step.disabled}">
 			<div class="bx--progress-step-button bx--progress-step-button--unclickable" role="button" tabindex="-1">
 				<ng-container *ngIf="!step.state.includes('error')">
-					<svg *ngIf="step.state.includes('complete')" focusable="false" preserveAspectRatio="xMidYMid meet"
-						xmlns="http://www.w3.org/2000/svg" fill="currentColor" width="16" height="16" viewBox="0 0 32 32"
-						aria-hidden="true">
-						<path d="M14 21.414L9 16.413 10.413 15 14 18.586 21.585 11 23 12.415 14 21.414z"></path>
-						<path d="M16,2A14,14,0,1,0,30,16,14,14,0,0,0,16,2Zm0,26A12,12,0,1,1,28,16,12,12,0,0,1,16,28Z"></path>
-					</svg>
-					<svg *ngIf="step.state.includes('current')" focusable="false" preserveAspectRatio="xMidYMid meet"
-						xmlns="http://www.w3.org/2000/svg" fill="currentColor" width="16" height="16" viewBox="0 0 32 32" aria-hidden="true">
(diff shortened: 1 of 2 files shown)


src/progress-indicator/progress-indicator.component.ts
```diff
@@ -29,16 +29,41 @@ import { Step } from "./progress-indicator-step.interface";
...
+							29.3z"></path>
+					</svg>
+				</ng-container>
+				<svg *ngIf="step.state.includes('error')" class="bx--progress__warning" focusable="false"
+					preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" fill="currentColor" width="16" height="16"
+					viewBox="0 0 16 16" aria-hidden="true">
+					<path d="M8,1C4.1,1,1,4.1,1,8s3.1,7,7,7s7-3.1,7-7S11.9,1,8,1z M8,14c-3.3,0-6-2.7-6-6s2.7-6,6-6s6,2.7,6,6S11.3,14,
+						8,14z"></path>
+					<path d="M7.5 4H8.5V9H7.5zM8 10.2c-.4 0-.8.3-.8.8s.3.8.8.8c.4 0 .8-.3.8-.8S8.4 10.2 8 10.2z"></path>
 				</svg>
```
Akshat55, line 66: Hey Stan, could you please use the ibmIcon directive with the following icon names:

* incomplete
* checkmark--outline
* warning
* circle-dash

An example of this for `incomplete/error` state would be:
```
<svg ibmIcon="incomplete" size="16" [ngClass]="{ 'bx--progress__warning': step.state.includes('error') }"></svg>
```

When testing, if you do NOT see the icon, you will need to import them in `icon.module.ts`.

Akshat55 (changes requested):

src/progress-indicator/progress-indicator.component.ts
stanislavgeorgiev: @Akshat55 I will give it a try - maybe the icons before were just not the right names

src/progress-indicator/progress-indicator.component.ts
```diff
@@ -85,7 +81,7 @@ export class ProgressIndicator implements OnChanges {
 	}
 	private _current: number;
 
-	constructor(protected experimental: ExperimentalService) {}
+	constructor(protected experimental: ExperimentalService) { }
```
Akshat55, line 84:
```suggestion
	constructor(protected experimental: ExperimentalService) {}
```

Akshat55 (changes requested):

Akshat55 (approved):
Approving as there is no breaking change. Icons are updated but behavior is still the same.

src/progress-indicator/progress-indicator.component.ts
stanislavgeorgiev: This was fixed by the linter, I think

zvonimirfras (approved):

stanislavgeorgiev:
@zvonimirfras @Akshat55 Do I submit a PR to `next` too or that's something that you're going to do?

PR #2493, opened 2023-03-21
fix(tabs): fix tab headers resize/scroll issue

yinglejia:

The tab headers component doesn't handle view resize event and reset scroll properly, which can cause display issues on view resize. This PR fixes that using `ResizeObserver`.

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}

### DIFF
--- a/src/tabs/tab-headers.component.ts
+++ b/src/tabs/tab-headers.component.ts
@@ -15,7 +15,6 @@ import {
 	OnInit,
 	ChangeDetectorRef
 } from "@angular/core";
-import { EventService } from "carbon-components-angular/utils";
 import { I18n } from "carbon-components-angular/i18n";
 import { Tab } from "./tab.component";
 
@@ -228,7 +227,6 @@ export class TabHeaders implements AfterContentInit, OnChanges, OnDestroy, OnIni
 	constructor(
 		protected elementRef: ElementRef,
 		protected changeDetectorRef: ChangeDetectorRef,
-		protected eventService: EventService,
 		protected i18n: I18n
 	) { }
 
--- a/src/tabs/tab-headers.component.ts
+++ b/src/tabs/tab-headers.component.ts
@@ -11,6 +11,7 @@ import {
 	TemplateRef,
 	OnChanges,
 	SimpleChanges,
+	OnDestroy,
 	OnInit,
 	ChangeDetectorRef
 } from "@angular/core";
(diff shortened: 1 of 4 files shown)


yinglejia:

yinglejia:
recheck

src/tabs/tab-headers.component.ts
```diff
@@ -290,8 +296,17 @@ export class TabHeaders implements AfterContentInit, OnChanges, OnInit {
...
-	ngOnInit() {
-		this.eventService.on(window as any, "resize", () => this.handleScroll());
+	ngOnInit(): void {
+		// Update scroll on resize
+		this.resizeObserver = new ResizeObserver(() => {
+			// Need to explicitly trigger change detection since this runs outside Angular zone
+			this.changeDetectorRef.detectChanges();
+		});
+		this.resizeObserver.observe(this.tabsScrollable.nativeElement);
+	}
```
Akshat55, line 306: Is `this.eventService.on(window as any, "resize", () => this.handleScroll());` not working? 🤔 

What browser are you using?

yinglejia: @Akshat55 It doesn't matter which browser; the problem is that we need to handle view resize, not window resize. For example, if we use a tabs component on a side panel which can be dynamically resized, the tabs will be resized accordingly. But we don't get resize event on window.

yinglejia: You can simulate the problem by manually setting the width of the parent container in DevTools. Since we set the width in DevTools, the Angular component tree will not receive any user event and therefore the Angular change detection won't be triggered. Therefore the tab scroll won't be updated properly.

image

In most cases, there's some user event happens on the component tree while the view is resized and the Angular change detection will be triggered, so this is not a problem.

However, in some cases, the view might be resized when there's not a user event on the component tree. For example, during open/expand/collapse animation. This requires us to handle view resize explicitly and trigger change detection.

Akshat55 (changes requested):
Could you remove the service import (line 18) & injection in the constructor (line 231):

`import { EventService } from "carbon-components-angular/utils";`

`protected eventService: EventService,`

yinglejia:
> Could you remove the service import (line 18) & injection in the constructor (line 231):
> 
> `import { EventService } from "carbon-components-angular/utils";`
> 
> `protected eventService: EventService,`

Done, thanks.

Akshat55 (approved):

zvonimirfras (approved):

PR #2497, opened 2023-03-23
fix: Improve grid performance by  preventing unnecessary change detection

Akshat55:
#### Changelog

**Changed**

* Assigning initial column classes after view initializes instead of ngOnInit to remove unnecessary SetTimeout which would trigger changeDetection for all child components.

### DIFF
--- a/src/grid/column.directive.ts
+++ b/src/grid/column.directive.ts
@@ -1,0 +1,161 @@
+import {
+	Directive,
+	HostBinding,
+	Input,
+	OnChanges,
+	OnDestroy,
+	OnInit,
+	Optional
+} from "@angular/core";
+import { Subscription } from "rxjs";
+import { GridService } from "./grid.service";
+
+@Directive({
+	selector: "[cdsCol], [ibmCol]"
+})
+export class ColumnDirective implements OnInit, OnChanges, OnDestroy {
+	@HostBinding("class")
+	get columnClasses(): string {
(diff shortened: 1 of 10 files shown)


Akshat55:
Test is failing because `NG0100: ExpressionChangedAfterItHasBeenCheckedError`. To fix this we would need to trigger manually trigger change detection. To avoid this, we are going to create a GridService & pass it as a provider in Grid Component.

We will take advantage of dependency injection & inject the GridService into columns. Then in columns `ngOnInit`, we subscribe & read the current status from the service & call appropriate function.

src/grid/column.directive.ts
```diff
@@ -0,0 +1,164 @@
...
+	protected _columnClasses: string[] = [];
+
+	private isCssGrid = false;
+	private subscription = new Subscription();
+
+	constructor(@Optional() private gridService: GridService) {}
+
+	ngOnInit() {
+		if (this.gridService) {
+			this.gridService.gridObservable.subscribe((isCssGrid: boolean) => {
```
eduardmarcinco, line 68: Isn't this missing `this.subscription = ` to unsubscribe on destroy?

zvonimirfras (approved):

src/grid/column.directive.ts
Akshat55: Thanks for pointing this out @eduardmarcinco. A fix has been merged to address this.

Issue #2499, opened 2023-03-27
Accessibility:Radio input and others with the name "tile-group-0" are not grouped together


ashishpawar12:
Elements: ibm-selection-tile

Related sets of radio buttons or checkboxes should be programmatically grouped

Why is this important?
Grouping related radio buttons and checkboxes enables assistive technology to convey the group relationship to users (e.g. "group of 5 radio buttons").

Element location
<input
    name="tile-group-0"
    value="Memory_Utilization-request_and_limit_not_set-18"
    type="radio"
    id="tile-12"
    class="bx--tile-input"
    tabindex="-1">

What to do
Group related controls of the same type by enclosing them within a <fieldset> element;
Or, group related radio buttons by using role="radiogroup";
Or, group related checkboxes by using role="group";
Or, disassociate unrelated controls or controls of different types by using different name attribute values;
Or, if the controls already have unique name attribute values and are in a data table, make sure the table is marked up using <table> or has a role of "grid" or "table".
Rule ID: WCAG20_Input_RadioChkInFieldSet
Reason ID: Fail_NotGroupedOtherNotGrouped

<img width="1439" alt="image" src="https://user-images.githubusercontent.com/18259137/227909101-5efd0dd1-c349-4112-acc7-92b0b165e8f0.png">

PR #2518, opened 2023-04-12
fix: Tile group accessibility

Akshat55:
Closes carbon-design-system/carbon-components-angular#2499

#### Changelog

**New**
* Can now add `legend` to tile group. It is a requirement in a fieldset.

**Changed**
* Tile group template has been updated, it is now being wrapped in a fieldset. The projected tiles have the same `name` attribute, hence they need to be wrapped in a fieldset. 
  * Can of worms! Fieldset need a legend!
* Selectable tile template has been updated, the input element has been moved outside of the label. This is because label is not a widget & was previously tabbable - now input is tabbable.


src/tiles/tile-group.component.ts
```diff
@@ -17,7 +18,14 @@ import { takeUntil } from "rxjs/operators";
 
 @Component({
 	selector: "ibm-tile-group",
-	template: `<ng-content select="ibm-selection-tile"></ng-content>`,
+	template: `
+		<fieldset>
+			<legend *ngIf="legend" class="bx--label">
+				<ng-container *ngIf="!isTemplate(legend)">{{legend}}</ng-container>
+				<ng-template *ngIf="isTemplate(legend)" [ngTemplateOutlet]="legend"></ng-template>
```
zvonimirfras, line 25: could be with `else`?

zvonimirfras (approved):

PR #2526, opened 2023-04-20
Fix: file uploader valueChanges event to fire for drag-and-drop

anemonetea:
Closes carbon-design-system/carbon-components-angular#2531

File Uploader had inconsistent behavior in how Forms events would fire between the drag-and-drop version and the standard version of the component. In Reactive Forms, for example, the `(valueChanges)` event wouldn't fire for the drag-and-drop version even though it would fire for the standard version.

#### Changelog

**Changed**

`valueChanges` API now behaves consistently across the `drop` version and non-drop version of the `ibm-file-uploader` component.

### DIFF
--- a/src/file-uploader/file-uploader.component.ts
+++ b/src/file-uploader/file-uploader.component.ts
@@ -172,7 +172,8 @@ export class FileUploader implements ControlValueAccessor {
 	}
 
 	/**
-	 * Specifies the property to be used as the return value to `ngModel`
+	 * Specifies the property to be used as the return value to `ngModel` and reactive forms.
+	 * Updates `this.files`.
 	 */
 	get value(): Set<FileItem> {
 		return this.files;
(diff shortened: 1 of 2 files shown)


src/file-uploader/file-uploader.component.ts
```diff
@@ -241,35 +242,37 @@ export class FileUploader implements ControlValueAccessor {
...
 				const fileItem = this.createFileItem(file);
-				this.files.add(fileItem);
+				newFiles.add(fileItem);
 			}
 		});
 
-		this.filesChange.emit(this.files);
-		this.value = this.files;
+		this.value = newFiles;
+		this.filesChange.emit(newFiles);
```
anemonetea, line 261: Here, I am copying the behaviour in `onFilesAdded()` function.

src/file-uploader/file-uploader.component.ts
```diff
@@ -241,35 +242,37 @@ export class FileUploader implements ControlValueAccessor {
...
 		this.dragOver = false;
 	}
 
 	removeFile(fileItem) {
-		let shouldEmit = true;
-		if (this.files) {
-			shouldEmit = this.files.has(fileItem);
-			this.files.delete(fileItem);
```
anemonetea, line 267: If `this.files.has(fileItem)` is `false`, then `this.files.delete(fileItem);` doesn't do anything, so it's redundant.

src/file-uploader/file-uploader.component.ts
anemonetea: The issue was that `onDrop(event)` would first set `this.files` and then set `this.value = this.files`, at which point the setter of `value` would check whether its input is different than `this.files`, find that it's not different, and not do anything (not fire the change callback).

Akshat55 (approved):

zvonimirfras (approved):

Issue #2558, opened 2023-05-16
combobox appenInline false creates many dom elements


huineng:
"carbon-components-angular": "^5.8.0",

when using combobox as a typeahead but with "appendInline": false then 
with each keystroke back and del a new dropdown list is appended

<img width="1080" alt="image" src="https://github.com/carbon-design-system/carbon-components-angular/assets/3482445/c24713cc-8629-4d37-a82a-1519fb532e00">

Akshat55:
Issue exists in v10 as well, going to resovle issue in v10 & merge changes into v11.

PR #2559, opened 2023-05-17
fix: Create only 1 listbox element per combobox

Akshat55:
Closes carbon-design-system/carbon-components-angular#2558

#### Changelog

**Update**
* Prevents new DOM element from being added on search string change


huineng:
can this also be merged in the @next version ?

PR #2579, opened 2023-05-26
feat: add additional wrapper components

Akshat55:
#### Changelog

**New**

* New wrapper components 
  * ibm-text-label & ibm-textarea-label - With new features, the label component is starting to get bloated & it's become hard to add new features without over complicating the html with templates. This approach will allow us to ensure we can make independent changes to the component to make it easier to maintain & use.

**Changed**

* We still offer `ibm-label` component, but now the ibm-label will wrap around ibm-text-label & ibm-textarea-label depending on the directive used. The point of this is to prevent a breaking change & ensure QOL. However, this approach does result in prop drilling, so we recommend directly using the new wrapper components if possible.

### DIFF
--- a/src/input/text-input-label.component.ts
+++ b/src/input/text-input-label.component.ts
@@ -10,7 +10,7 @@ import {
 } from "@angular/core";
 
 @Component({
-	selector: 'ibm-text-input',
+	selector: 'ibm-text-label',
 	template: `
 		<label
 			[for]="labelInputID"
--- a/src/input/text-input-label.component.ts
+++ b/src/input/text-input-label.component.ts
@@ -87,7 +87,6 @@ export class TextInputLabelComponent implements AfterViewInit {
 	@Input() skeleton = false;
 
 	/**
-	 * **Internal**
 	 * Helper input property for ease of migration
 	 * Since we cannot pass ng-content down easily from label component, we will accept the templates
 	 */
--- a/src/input/text-input-label.component.ts
+++ b/src/input/text-input-label.component.ts
@@ -1,0 +1,162 @@
+import {
+	Component,
+	Input,
+	AfterViewInit,
+	ElementRef,
+	HostBinding,
+	TemplateRef,
+	ViewChild,
+	ChangeDetectorRef
+} from "@angular/core";
+
+@Component({
+	selector: 'ibm-text-input',
+	template: `
+		<label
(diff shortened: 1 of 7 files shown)


src/input/text-input-label.component.ts
```diff
@@ -0,0 +1,162 @@
...
+	 * Set to `true` for a disabled label.
+	 */
+	@Input() disabled = false;
+	/**
+	 * Set to `true` for a loading label.
+	 */
+	@Input() skeleton = false;
+
+	/**
+	 * **Internal**
```
Akshat55, line 90:
```suggestion
	 *
```

src/input/text-input-label.component.ts
```diff
@@ -0,0 +1,162 @@
...
+	 * Creates an instance of Label.
+	 */
+	constructor(protected changeDetectorRef: ChangeDetectorRef) {}
+
+	/**
+	 * Sets the id on the input item associated with the `Label`.
+	 */
+	ngAfterViewInit() {
+		if (this.wrapper) {
+			// Prioritize setting id to `input` & `textarea` over div
```
Akshat55, line 136:
```suggestion
			// Prioritize setting id to `input` over div
```

src/input/textarea-label.component.ts
```diff
@@ -0,0 +1,168 @@
...
+	 * Set to `true` for a disabled label.
+	 */
+	@Input() disabled = false;
+	/**
+	 * Set to `true` for a loading label.
+	 */
+	@Input() skeleton = false;
+
+	/**
+	 * **Internal**
```
Akshat55, line 93:
```suggestion
	 *
```

src/input/textarea-label.component.ts
```diff
@@ -0,0 +1,168 @@
...
+	 * Creates an instance of Label.
+	 */
+	constructor(protected changeDetectorRef: ChangeDetectorRef) {}
+
+	/**
+	 * Sets the id on the input item associated with the `Label`.
+	 */
+	ngAfterViewInit() {
+		if (this.wrapper) {
+			// Prioritize setting id to `input` & `textarea` over div
```
Akshat55, line 142:
```suggestion
			// Prioritize setting id to `textarea` over div
```

zvonimirfras (dismissed):

PR #2580, opened 2023-05-26
fix: No longer mark components as experimental

Akshat55:
#### Changelog

**Removed**

* `modal-header`, `toast`, & `progress-indicator` components are no longer marked as experimental. Hence, `ExperimentalService` are no longer injected in constructor. 
* This also means that you will not need to import ExperimentalService on a root level when using Notification service with standalone components!

### DIFF
--- a/src/modal/modal-header.component.ts
+++ b/src/modal/modal-header.component.ts
@@ -5,7 +5,6 @@ import {
 	Input
 } from "@angular/core";
 import { I18n } from "carbon-components-angular/i18n";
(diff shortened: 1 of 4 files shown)


Akshat55:
@zvonimirfras In React v10, the components such as pagination have a directory called `experimental` & are also in experimental section in storybook. The components mentioned in PR no longer have such directory & are not mentioned in such section in storybook.

zvonimirfras (approved):

Issue #2552, opened 2023-05-09
Accessability Violation : The tabbable element's role 'none' is not a widget role（ibm-file-uploader）


GSSCDylan:
## Detailed description

> Describe in detail the issue you're having.

The tabbable element's role 'none' is not a widget role
> Is this a feature request (new component, new icon), a bug, or a general issue?

bug
> Is this issue related to a specific component?

ibm-file-uploader
> What did you expect to happen? What happened instead? What would you like to see changed?

add a widget role for the lable tag of ibm-file-uploader
> What browser are you working in?

Firefox
> What version of the Carbon Design System are you using?

"carbon-components": "^10.58.3",
"carbon-components-angular": "^4.71.8",

https://angular.carbondesignsystem.com/?path=/story/components-file-uploader--drag-and-drop

## Additional information

* Screenshots or code
<img width="1610" alt="label1" src="https://github.com/carbon-design-system/carbon-components-angular/assets/73825731/847b3e06-0d2d-4c71-bf41-7f80297c6cf6">

<img width="1404" alt="label2" src="https://github.com/carbon-design-system/carbon-components-angular/assets/73825731/ff8065f4-b79d-4c12-9a76-d144d9b8a529">

PR #2581, opened 2023-05-26
fix(file-uploader): Fix a11y role 'none'

stanislavgeorgiev:
Closes carbon-design-system/carbon-components-angular#2552

There's no need to have the `<label>` tabbable with no `role` and to have the `<div>` element not tabbable but with `role` so I consolidated both attributes onto the `<label>` element. Tested tabbing with the keyboard on the drag and drop area and opening the file dialog. Tested also just clicking with the mouse on the area.

The error is no longer showing in the IBM Equal Access Accessibility Checker:

image

### DIFF
--- a/src/file-uploader/file-uploader.component.ts
+++ b/src/file-uploader/file-uploader.component.ts
@@ -31,11 +31,11 @@ const noop = () => { };
 					(keyup.enter)="fileInput.click()"
 					(keyup.space)="fileInput.click()"
 					[ngClass]="{'bx--file-browse-btn--disabled': disabled}"
+					role="button"
 					tabindex="0">
 					<div
 						class="bx--file__drop-container"
 						[ngClass]="{'bx--file__drop-container--drag-over': dragOver}"
-						role="button"
 						(click)="fileInput.click()"
 						[attr.for]="fileUploaderId"
 						(dragover)="onDragOver($event)"


src/file-uploader/file-uploader.component.ts
```diff
@@ -31,11 +31,11 @@ const noop = () => { };
 					(keyup.enter)="fileInput.click()"
 					(keyup.space)="fileInput.click()"
 					[ngClass]="{'bx--file-browse-btn--disabled': disabled}"
+					role="button"
```
Akshat55, line 34: Label element cannot have the ARIA role 'button'.

A possible solution to this would be to move the tabindex to the child div element. We have the `for` property assigned to the div, so I think that would give screen readers context & it has the `role` set to `button` which makes sense to allow it to be focusable 🤔

Akshat55 (changes requested):

src/file-uploader/file-uploader.component.ts
stanislavgeorgiev: Sure, I can try that.
On the same note about that `<label>` - there's 2 label elements in the same component which might also be a part of the problem

Akshat55: Hmm, you're right, there should only be one.  I'll need to play around with it to figure out the best way to approach this. Removing one of the labels will still show the `a11y role 'none'` error unfortunately.

stanislavgeorgiev: I ended up removing that whole secondary label element as it's not really needed

stanislavgeorgiev: > Hmm, you're right, there should only be one. I'll need to play around with it to figure out the best way to approach this. Removing one of the labels will still show the `a11y role 'none'` error unfortunately.

Ah, you're right

stanislavgeorgiev: Wait, I was looking at the wrong branch. Here's what the checker is showing now - there's no 'none' complaints.

image

stanislavgeorgiev: @Akshat55 I made the change you've requested and the PR is ready for another review

Akshat55 (approved):

zvonimirfras (approved):

Issue #2551, opened 2023-05-09
Accessability Violation : The ARIA attributes "aria-label" are not valid for the element <div> with implicit ARIA role "generic"（ibm-table-toolbar）


GSSCDylan:
## Detailed description

> Describe in detail the issue you're having.

The ARIA attributes "aria-label" are not valid for the element

with implicit ARIA role "generic"
> Is this a feature request (new component, new icon), a bug, or a general issue?

bug
> Is this issue related to a specific component?

ibm-table-toolbar
> What did you expect to happen? What happened instead? What would you like to see changed?

Need to remove aria-label in div tag of  ibm-table-toolbar
> What browser are you working in?

Firefox
> What version of the Carbon Design System are you using?

"carbon-components": "^10.58.3",
"carbon-components-angular": "^4.71.8",

https://angular.carbondesignsystem.com/?path=/story/components-table--with-toolbar

## Additional information

* Screenshots or code
<img width="1752" alt="action tool bar" src="https://user-images.githubusercontent.com/73825731/237058731-1424f7ae-581f-425f-a119-327f9d5ee663.png">
<img width="1732" alt="action tool bar2" src="https://user-images.githubusercontent.com/73825731/237059423-dad8ec19-4128-4151-b2f1-f7c5772e2a95.png">

PR #2582, opened 2023-05-26
fix(table-toolbar): Fix a11y violations

stanislavgeorgiev:
Closes carbon-design-system/carbon-components-angular#2551

This PR fixes a11y violations when the toolbar is not showing and when it's showing. It also changes the toolbar behavior to be completely invisible (as in not in the DOM) when it's not showing which fixes one more violation without changing the functionality in any way.

Before

image

image

After

image

image

### DIFF
--- a/src/table/toolbar/table-toolbar.component.ts
+++ b/src/table/toolbar/table-toolbar.component.ts
@@ -50,12 +50,10 @@ import { TableRowSize } from "../table.types";
 		class="bx--table-toolbar"
 		[ngClass]="{'bx--table-toolbar--small' : size === 'sm'}">
 		<div
-			*ngIf="model"
-			class="bx--batch-actions"
-			[ngClass]="{
-				'bx--batch-actions--active': selected
-			}"
-			[attr.aria-label]="actionBarLabel.subject | async">
+			*ngIf="model && selected"
+			class="bx--batch-actions bx--batch-actions--active"
+			[attr.aria-label]="actionBarLabel.subject | async"
+			role="group">
 			<div class="bx--action-list">
 				<ng-content select="ibm-table-toolbar-actions"></ng-content>
 				<button


src/table/toolbar/table-toolbar.component.ts
```diff
@@ -50,12 +50,10 @@ import { TableRowSize } from "../table.types";
...
-			*ngIf="model"
-			class="bx--batch-actions"
-			[ngClass]="{
-				'bx--batch-actions--active': selected
-			}"
-			[attr.aria-label]="actionBarLabel.subject | async">
+			*ngIf="model && selected"
+			class="bx--batch-actions bx--batch-actions--active"
+			[attr.aria-label]="actionBarLabel.subject | async"
+			role="group">
```
Akshat55, line 56:
```suggestion
			*ngIf="model && selected" class="bx--batch-actions bx--batch-actions--active">
```

Could you move the aria-label to the section (line 49). This will ensure the entire section is accessible.

Akshat55 (changes requested):

stanislavgeorgiev:
@Akshat55 I made the change you've requested and the PR is ready for another review

src/table/toolbar/table-toolbar.component.ts
```diff
@@ -48,14 +48,12 @@ import { TableRowSize } from "../table.types";
...
+		role="group">
 		<div
-			*ngIf="model"
-			class="bx--batch-actions"
-			[ngClass]="{
-				'bx--batch-actions--active': selected
-			}"
-			[attr.aria-label]="actionBarLabel.subject | async">
+			*ngIf="model && selected"
+			class="bx--batch-actions bx--batch-actions--active">
```
Akshat55, line 56:
```suggestion
			*ngIf="model"
			class="bx--batch-actions"
			[ngClass]="{
				'bx--batch-actions--active': selected
			}">
```

We have animations that are applied when `bx--batch-actions--active`. When we remove the entire element when selected is false, this removes the animation.

Akshat55 (changes requested):

stanislavgeorgiev:
Thanks for the updates @Akshat55. I just reran the accessibility checker and I can see there are even less violations than before! :)

image


image

Akshat55 (approved):

Issue #2535, opened 2023-05-01
[Toggle button]: Name and value attributes of the input elements in toggle button comes as undefined  


RishabhSharma6:
## Detailed description

The name and value attributes of the input elements in the toggle button comes as undefined . The value field should have a binary (true/false) value.

## Additional information
<img width="1481" alt="Screenshot 2023-05-01 at 8 01 19 PM" src="https://user-images.githubusercontent.com/111842352/235468651-32aef646-772c-47c6-9452-48e9e3f7f9e2.png">

PR #2585, opened 2023-05-27
fix(toggle): set the name and value attributes

rkkp1023:
Closes https://github.com/carbon-design-system/carbon-components-angular/issues/2535

This Pr will only add the value of the attribute [value] and [name] to the input element in the toggle.

#### Changelog

**Before**
image

**After**

image

### DIFF
--- a/src/toggle/toggle.component.ts
+++ b/src/toggle/toggle.component.ts
@@ -103,6 +103,16 @@ export class Toggle extends Checkbox {
 	static toggleCount = 0;
 
 	/**
+	 * Variable set the value to input true/false.
+	 */
+	value = false;
+
+	/**
+	 * Variable set the name to input same as Id.
+	 */
+	name = "toggle-" + Toggle.toggleCount;
+
+	/**
 	 * Text that is set on the left side of the toggle.
 	 */
 	@Input()
@@ -193,6 +203,7 @@ export class Toggle extends Checkbox {
 		this.change.emit(event);
 		/* end deprecation */
 
+		this.value = this.checked;
 		this.checkedChange.emit(this.checked);
 		this.propagateChange(this.checked);
 	}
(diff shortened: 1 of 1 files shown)


src/toggle/toggle.component.ts
```diff
@@ -193,6 +203,7 @@ export class Toggle extends Checkbox {
 		this.change.emit(event);
 		/* end deprecation */
 
+		this.value = this.checked;
```
Akshat55, line 206:
```suggestion
```

src/toggle/toggle.component.ts
```diff
@@ -102,6 +102,16 @@ export class Toggle extends Checkbox {
...
+	/**
+	 * Variable set the value to input true/false.
+	 */
+	value = false;
+
+	/**
+	 * Variable set the name to input same as Id.
+	 */
+	name = "toggle-" + Toggle.toggleCount;
+
```
Akshat55, line 114: `value` & `name` are metadata (@Input) in Checkbox and are inherited. They appear as `undefined` because they are not initialized in checkbox component.

You can instead change the template (line 70 & 71) to the following:

image


```
			[attr.value]="value"
			[attr.name]="name"
```

This way, if they equal undefined or null, they will be removed from the element.

Akshat55: The below change will be unnecessary after & can be removed.

Akshat55 (changes requested):

src/toggle/toggle.component.ts
rkkp1023: addressed

src/toggle/toggle.component.ts
rkkp1023: ok removed

zvonimirfras (approved):

Issue #2591, opened 2023-06-02
searchbar icon


huineng:
v 5.8.0

you class input class="cds--search-input" has align-items center

but the icon itself is still right aligned, while a per design , it should be centered

<img width="276" alt="image" src="https://github.com/carbon-design-system/carbon-components-angular/assets/3482445/751ed550-19f5-4442-b817-4e8d76ba888e">

<img width="433" alt="image" src="https://github.com/carbon-design-system/carbon-components-angular/assets/3482445/da3eb5d7-ec12-4b49-bd69-7bdb7f7ad067">

Could you have a look at this , minor thing.. nevertheless we want an excelent library to stay excellent

PR #2592, opened 2023-06-02
fix: Align search icon to be in center

Akshat55:
Closes carbon-design-system/carbon-components-angular#2591

#### Changelog

* Make toolbar search size `lg` by default


src/table/toolbar/table-toolbar-search.component.ts
```diff
@@ -22,6 +22,8 @@ import { NG_VALUE_ACCESSOR } from "@angular/forms";
 export class TableToolbarSearch extends Search implements AfterViewInit {
 	tableSearch = true;
 
+	size: "sm" | "md" | "lg" = "lg";
```
Akshat55, line 25:
```suggestion
	size = "lg";
```

zvonimirfras (approved):

PR #2619, opened 2023-06-28
feat:Min-Max date range feature-date picker issue#2618

amit-gaba:

{{short description}}

This PR adds the Min/Max feature to existing Date picker component.
Where the dates outside Min/Max range will not be available for selection.

**Changed**

Added Min Max date range feature.

9d7ce878-742d-47b3-897f-89d8b8b9d396

8f532995-7709-4300-af21-64c250965f55

### DIFF
--- a/src/datepicker-input/datepicker-input.component.ts
+++ b/src/datepicker-input/datepicker-input.component.ts
@@ -1,5 +1,4 @@
-import {
-	Component,
+import {	Component,
 	Input,
 	Output,
 	EventEmitter,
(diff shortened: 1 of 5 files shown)


Akshat55 (commented):
I suggest you have a look at the following [code sandbox](https://codesandbox.io/s/silly-butterfly-78l4sn?file=/src/app/app.component.html) as this feature is available just not well documented.

Perhaps you can improve the documentation here and include an example for setting minDate/maxDate via flatpickrOptions input prop.

PR #2632, opened 2023-07-10
Checkbox aria label

### LINKED ISSUE #2623: Accessability Violation : The ARIA attributes "aria-label" are not valid for the element <label> with ARIA role "none"
[Title]: Accessability Violation : The ARIA attributes "aria-label" are not valid for the element <label> with ARIA role "none"

  IBM Accessibility Checker tool reported that "The ARIA attributes "aria-label" are not valid for the element <label> with ARIA role "none"" for checkbox component label tag
  
   Bug

    checkbox Component - <ibm-checkbox>

    <ibm-checkbox >  --> <label class="bx--checkbox-label" for="DD:DDRP_input" aria-label=""> </label> **<label>** should not have aria-label

   Mozila Firefox

 ```  Firefox 106.0.3 ```

   "carbon-components": "^10.58.1",
    "carbon-components-angular": "^4.59.14",

1. Run the IBM Accessibility Checker tool on component <ibm-checkbox>

## Additional information

image

Element Location :

    <label
    aria-label=""
    for="DD:DDRP_input"
    class="bx--checkbox-label">

sant-santhiya:
Closes #2623

**Changed**

* Set ariaLabel attribute to be undefined by default instead of an empty string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,9 +63,11 @@ Angular version support matrix:
 | 9       | ❌ | ✅ | ❌                   |
 | 10      | ❌ | ✅ | ❌                   |
 | 11      | ❌ | ✅ | ❌                   |
-| 12      | ❌ | ✅ | ✅                   |
-| 13      | ❌ | ✅ | ✅                   |
+| 12      | ❌ | ✅ | ❌                   |
+| 13      | ❌ | ✅ | ❌                   |
 | 14      | ❌ | ✅ | ✅                   |
+| 15      | ❌ | ✅ | ✅                   |
+| 16      | ❌ | ❌ | ✅                   |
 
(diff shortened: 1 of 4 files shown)


sant-santhiya:

Akshat55 (changes requested):
Thanks for the PR @sant-santhiya. I agree that we do not need aria label, but some teams _may_ be using it.

So could you undo the template change and instead of removing it, can you set Line 152 to undefined? So it would be:

```ts
@Input("ariaLabel") _ariaLabel;
```

We use attribute binding for aria label so if it is undefined then it will not add the attribute to the element.

Akshat55 (commented):
Forgot to approve - but it's good to go!

PR #2681, opened 2023-08-24
fix: access correct object property to resolve webpack import error

Akshat55:
#### Changelog

* Updating flatpickr l10n `en` (english) short hand value only once.
* Resolves *flatpickr webpack imported module is undefined* issue


zvonimirfras (approved):
Does not lgtm

Issue #2286, opened 2022-10-11
Accessibility: File upload desc and error not read by screen reader


makandre:
## Detailed description

> Describe in detail the issue you're having.

The description and error text in file uploader component are not read by screen reader.

> Is this a feature request (new component, new icon), a bug, or a general issue?

Bug

> Is this issue related to a specific component?

File uploader

> What did you expect to happen? What happened instead? What would you like to see changed?

The description and error text should be read by screen reader

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

`"carbon-components-angular": "^4.59.19"`

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

IBM ADDI

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

* Screenshots or code

![image](https://user-images.githubusercontent.com/17596240/195066236-ad56fa64-ed1d-4f18-beb5-6ee0f72e28c5.png)

* Notes

Akshat55:
Hey @makandre what screen reader are you using? Screen reader that comes with OS? If yes, which OS are you using?

makandre:
I tested with VoiceOver in Mac. But I believe the original reporter was using JAWS on Windows

Akshat55:
I've tested it with windows 11 screen reader & the helper/invalid text are both read 🤔 Will run it on Mac & with JAWS. Did you programmatically set for the role for the helper text? Or is everything default?

makandre:
I’m just using defaults

makandre:
It happens in the storybook example too.  I should clarify, the description for example is read if I highlight the text.  But there's no way to tab to that description; someone needing screen reader is likely also using keyboard navigation.

stanislavgeorgiev:
I'm not sure that text blocks are supposed to be tabbable - read by a screen reader yes though

andyl309:
This issue is being flagged as a failure with our accessibility compliance and we will need to file a deviation without a fix. Does carbon team have resources to address this issue soon?

andyl309:
After verifying with the accessibility team, for this defect a role label needs to be added to the invalid text. Tabbing to the description is not required.

PR #2683, opened 2023-08-25
fix: enable screen reader to read file upload desc and error

andyl309:
Closes carbon-design-system/carbon-components-angular#2286

Adding a role attribute to description and invalid text so screen reader such as JAWS can read them.

https://www.ibm.com/able/requirements/requirements/#4_1_3

This fix is for version 5

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}


andyl309:

Issue #2519, opened 2023-04-13
Accessability Violation : <ibm-panel> ARIA attributes "aria-label" are not valid for the element <div> with implicit ARIA role "generic" 


sant-santhiya:
## Detailed description

> Describe in detail the issue you're having.

The ARIA attributes "aria-label" are not valid for the element <div> with implicit ARIA role "generic". Need to remove 'aria-label' as it is not valid for div tag.

> Is this a feature request (new component, new icon), a bug, or a general issue?

Bug

> Is this issue related to a specific component?

`<ibm-header> ---> <ibm-panel> --> <div class="bx--panel--overlay" >`

> What did you expect to happen? What happened instead? What would you like to see changed?

Need to remove aria-label in div tag of `<ibm-panel>`

> What browser are you working in?

Mozila Firefox

> What version of the Carbon Design System are you using?

"carbon-components": "^10.56.0",
"carbon-components-angular": "^4.59.8",

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

within one week

## Steps to reproduce the issue

Run the IBM Accessibility Checker tool on component. [ui-shell--right-panel-with-router](https://angular.carbondesignsystem.com/?path=/story/components-ui-shell--right-panel-with-router)

## Additional information

* Screenshots or code

![image](https://user-images.githubusercontent.com/104894503/231660280-bf57201b-0570-4b1d-8867-1d7033b1a515.png)

```
<div
    aria-label="Header panel"
    class="bx--panel--overlay bx--panel--expanded">
```

makandre:
There's essentially the same issue reported on the react side, with more details and discussions https://github.com/carbon-design-system/carbon/issues/13464

andyl309:
This issue is being flagged as a failure with our accessibility compliance and we will need to file a deviation without a fix. Does carbon team have resources to address this issue soon?

keithcha:
It looks like removing [aria-label](https://github.com/carbon-design-system/carbon-components-angular/blob/4dcb72f3d34ee438a3134a489ae6c5bb1aaf21be/src/ui-shell/panel/panel.component.ts#L12C33-L12C33) could address this issue.

PR #2685, opened 2023-08-25
fix: remove invalid aria-label for panel

andyl309:
Closes carbon-design-system/carbon-components-angular#2519

Aria attributes must be valid for the element and ARIA role to which they are assigned

https://www.ibm.com/able/requirements/requirements/#4_1_2

### DIFF
--- a/src/ui-shell/panel/panel.component.ts
+++ b/src/ui-shell/panel/panel.component.ts
@@ -9,7 +9,6 @@ import { Component, Input } from "@angular/core";
 	template: `
 		<div
 			class="bx--panel--overlay"
-			[attr.aria-label]="ariaLabel"
 			[ngClass]="{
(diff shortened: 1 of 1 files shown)


andyl309:
Sorry for the urgent request but is it possible to get a review by today or tomorrow? Otherwise my team will need to file a deviation this week.

src/ui-shell/panel/panel.component.ts
```diff
@@ -23,5 +23,5 @@ export class Panel {
 	 */
 	@Input() expanded = false;
 
-	@Input() ariaLabel = "Header panel";
+	@Input() ariaLabel = null;
```
Akshat55, line 26:
```suggestion
	@Input() ariaLabel;
```

Akshat55 (changes requested):

PR #2686, opened 2023-08-25
fix: remove invalid aria-label for panel

andyl309:
Closes carbon-design-system/carbon-components-angular#2519

Aria attributes must be valid for the element and ARIA role to which they are assigned

https://www.ibm.com/able/requirements/requirements/#4_1_2

### DIFF
--- a/src/ui-shell/panel/panel.component.ts
+++ b/src/ui-shell/panel/panel.component.ts
@@ -23,5 +23,5 @@ export class Panel {
 	 */
 	@Input() expanded = false;
 
-	@Input() ariaLabel = null;
+	@Input() ariaLabel;
 }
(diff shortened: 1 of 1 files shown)


Akshat55 (commented):
Hey @andyl309 

This would be a breaking change. What you can do is pass in `null` or `undefined` for `ariaLabel` input. We are using attribute binding so if the the value is null, it will not add the aria-label attribute to the element.

src/ui-shell/panel/panel.component.ts
```diff
@@ -23,5 +23,5 @@ export class Panel {
 	 */
 	@Input() expanded = false;
 
-	@Input() ariaLabel = "Header panel";
+	@Input() ariaLabel = null;
```
Akshat55, line 26:
```suggestion
	@Input() ariaLabel;
```

Akshat55 (changes requested):

Issue #2698, opened 2023-09-19
CDS Header Action Click property not propagating 


ParkerMIBM:
## Detailed description

> Migrating from carbon-component-angular@4.59 to 5.14, both cds-header-action and ibm-header-action are no longer calling the function I specified in the click property. I see the "active" and "selected" properties being updated correctly, but I am unsure why the (click)="refreshDataOnPage($event)" that has long existed on this element would stop being called.

> Is this a feature request (new component, new icon), a bug, or a general issue?
This is a bug

> Is this issue related to a specific component?
This is related to the UIShell - specifically the cds-header-action component. 

> What did you expect to happen? What happened instead? What would you like to see changed?
The v4 to v5 migration document linked in this repository had me expecting to change from description to action. It said nothing about how events would propagate. I would have expected that my (click) property would have continued to work as normal.

> What browser are you working in?
Firefox, chrome, safari

> What version of the Carbon Design System are you using?
    "@carbon/icons": "11.26.0",
    "@carbon/styles": "1.37.0",
    "@carbon/themes": "11.24.0",
    "carbon-components-angular": "5.14.10",
    "@angular/core": "14.3.0",

## Steps to reproduce the issue
I tried to recreate on the code sandbox but it was very down level and updating it to try to recreate my environment was starting to take too long through that interface.
Basically
1. Have a UIShellHeader bar with a <ibm-header-action> component, what was displayed was an icon like in the story.
2. create a function within the component "public myOnClickFunction( event ) { console.log("we made it!"); }"
3. add a click property to propagate an onclick <ibm-header-action (click)="myOnClickFunction($event)">
4. Load page
5. Click button
6. notice in console log no "we made it!" message

## Additional information

* I am not getting any console errors
* The angular inspection tools on chrome do not show the click property on the element

Akshat55:
Please use the `selected` and an `activeChange` event.

PR #2699, opened 2023-09-19
updated to have a click event emitter in HeaderAction

ParkerMIBM:
Closes carbon-design-system/carbon-components-angular#2698

Since moving to the base icon button, the click event does not propagate up the chain. This adds the click event as an Output on the BaseIconButton and the emitter within the onClick function in the UIShellModule Header Action Component

#### Changelog

**Changed**

* BaseIconButton to have an Output
* Header Action click feature returned

### DIFF
--- a/src/ui-shell/header/header-action.component.ts
+++ b/src/ui-shell/header/header-action.component.ts
@@ -60,5 +60,6 @@ export class HeaderAction extends BaseIconButton {
 		this.active = !this.active;
 		this.selected.emit(this.active);
 		this.activeChange.emit(this.active);
+		this.click.emit(this.active);
 	}
 }
--- a/src/button/base-icon-button.component.ts
+++ b/src/button/base-icon-button.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input } from "@angular/core";
+import { Component, EventEmitter, Input, Output } from "@angular/core";
 
 /**
  * Base button with common input properties for configuring icon button.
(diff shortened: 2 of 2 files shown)


ParkerMIBM:

src/ui-shell/header/header-action.component.ts
```diff
@@ -60,5 +60,6 @@ export class HeaderAction extends BaseIconButton {
 		this.active = !this.active;
 		this.selected.emit(this.active);
 		this.activeChange.emit(this.active);
+		this.click.emit(this.active);
```
Akshat55, line 63: This is redundant. There is a `selected` and an `activeChange` event emitted with the same data. Instead listening for click event, just listen for `selected` or `activeChange` event. You can even double bind a variable with `[active]`.

ParkerMIBM: Then is that not an upgrade action that should be documented?
The (click) property on header-action-component seems like it is unused now, and code in the field binding to click breaks when moving from v4 to v5.

Akshat55: You're right, we must've missed this. I'll go in and add this to the v4 to v5 migration documentation.

Akshat55: Just checked, it never emitted a `click` event. Only the events I mentioned above were emitted. You can verify this here:

You can view the v10 files here: https://github.com/carbon-design-system/carbon-components-angular/blob/v10/src/ui-shell/header/header-action.component.ts

Akshat55:
@ParkerMIBM I'm closing this PR as we have alternatives available.

PR #2464, opened 2023-02-22
feat: add contained-list

maicongodinho:
New Contained List component 🥳

Reference: https://carbondesignsystem.com/components/contained-list/usage/
React team implementation: https://react.carbondesignsystem.com/?path=/story/components-containedlist--default

PR #2717, opened 2023-10-18
Fix/contained list sentence capitlization

hassniazi:

Closes  #2464 Fixed capitlization issues. Not addressing the comment about readonly instead of @Input? as if using @input would have to handle it and don't think it needs handling rn. 

#### Changelog

**New**

* {{new thing}}

**Changed**

* {{change thing}}

**Removed**

* {{removed thing}}


hassniazi:

Akshat55:
Thanks for contributing this, we recently merged contained list! So I'll be closing this PR!

Issue #2687, opened 2023-08-25
Unable to use ```lint-fix```and remove IBM prefix when working via the tutorials in Angular


ENate:
## Detailed description
I followed the tutorial on your website on how to design an angular carbon design using the tutorial guides provided on your website.
> Describe in detail the issue you're having.
- I noticed that a number of commands do not work on Angular 16.0+ For instance, 
- ``` ng g component header --lint-fix```does not work with the ```--lint-fix```, and I do not know what you meant by adding the following code 
- ```
 import Notification20 from '@carbon/icons/es/notification/20';
import UserAvatar20 from '@carbon/icons/es/user--avatar/20';
import AppSwitcher20 from '@carbon/icons/es/app-switcher/20';
```
inside ```src/app/header/app.modules.ts ```since we need 1 ``àpp.module.ts```in this case.  Does this mean we need another ```src/app/header/app.modules.ts

```after creating the header component?

Also, running step 1 of the tutorial using```npm start ```resulted in the following errors: 
```
Error: src/app/header/header.component.html:6:9 - error NG8001: 'ibm-header-action' is not a known element:
1. If 'ibm-header-action' is an Angular component, then verify that it is part of this module. 
2. If 'ibm-header-action' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message.
```
I added ```CUSTOM_ELEMENTS_SCHEMA```to the ``ǸgModules.schema```in the ```app.module.ts```folder but still go the error.
> Is this a feature request (new component, new icon), a bug, or a general issue?
General issue
> Is this issue related to a specific component?
Yes it is also related to UI Shell Header component. Unable to remove the ``ÌBM```prefix in the header element of this component.

It seems. But I imagine it has to do with the Angular version, maybe.

> What did you expect to happen? What happened instead? What would you like to see changed?

To see the output of the tutorial guide for step 1

> What browser are you working in?
Google-chrome, Firefox, Brave
> What version of the Carbon Design System are you using?
5.+
> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
Personal project

## Steps to reproduce the issue

<!--
Please try to re-create the issue using our Codesandbox template available here:
https://codesandbox.io/s/carbon-components-angular-0129r494ql

This helps reduce the issue to an isolated case that helps us out when debugging.
If you're able to reproduce using the codes-and-box link above, please include the
[...]

PR #2723, opened 2023-10-25
Issue #2687- remove ibm prefix from header-action component

nipun005:
Closes https://github.com/carbon-design-system/carbon-components-angular/issues/2687

This PR remove "ibm" prefix from the selector of "ibm-header-action" component. Now the user can able to directly call this component by <header-action>

#### Changelog

**Changed**

Previously to use header-action" component, we were calling it by <ibm-header-action>.
As part of this PR, We have removed "ibm" as prefix and this component can be directly use by <header-action>.
 According to discussion on Original issue #2687 , To remove the IBM prefix can increase the outreach.

### DIFF
--- a/src/ui-shell/header/header-action.component.ts
+++ b/src/ui-shell/header/header-action.component.ts
@@ -11,7 +11,7 @@ import { BaseIconButton } from "carbon-components-angular/button";
  * Contained by `HeaderGlobal`. Generally used to trigger `Panel`s
  */
 @Component({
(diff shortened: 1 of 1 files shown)


nipun005:

Akshat55 (commented):
Will not be merged as this is a breaking change.

Akshat55:
I'll be closing this PR, but feel free to respond and we can discuss why this change is a breaking change and cannot be merged according semantic release versioning.

Issue #2724, opened 2023-10-25
ActionableNotification close button action.


placzkowskim:
ActionableNotification close button action.

Hi I have created actionable notification like:

`    <cds-actionable-notification
        [notificationObj]="{
    inline: true,
    type: 'error',
    title: 'Error: Unable to procees your request this time.',
    actionsTemplate: customActionableTrigger,
    message: 'Please try again later',
    showClose:true,
    lowContrast: true,
}"
        (close)="onNotificationClose()"
    >
    </cds-actionable-notification>`

I need such default behaviour as close that component on 'x' click. 
My onNotificationClose() method is beeing properly invoked on 'x' click.
How can I achieve it ? 
Should I somehow register notification component created in template in NotificationService ? 
`NotificationDisplayService ` exposes `close()` method but it accepts ComponentRef. Does this mean that I should create that notification using ComponentResolverFactory ? Or there is some simple solution that I'm missing.

nipun005:
Hi @placzkowskim, To implement such functionality- When the user will click on close icon of the actionable notification, Then your method which you written should invoke, Fix is required in actionable-notification file.
Please assign this issue to me, I will fix it and would raise PR.

nipun005:
Hi @placzkowskim,  Below PR has been raised for above issue.
https://github.com/carbon-design-system/carbon-components-angular/pull/2727

After merging of this PR, On Click of X button of ActionableNotification, "onNotificationClose()" method will automaticaaly be invoked if you check with above code again that you used previously, and you would not be needed  to use any other approach like- ComponentResolverFactory or something.

Akshat55:
Hi, if you would like to `close` the notification. I highly recommend using the notification service to dynamically render your component. The component will be dynamically removed on `x` click. We have quite a few examples of this in our storybook: 

https://angular.carbondesignsystem.com/?path=/story/components-notification--dynamic-inline

PR #2727, opened 2023-10-27
#2724- Fix of close method invokation in ActionableNotification close…

nipun005:
… button action

Closes https://github.com/carbon-design-system/carbon-components-angular/issues/2724

When the user will click on close button of ActionableNotification , then the function was not invoked which user defined in html file against close method.
After this Fix- The function will be invoked.

#### Changelog

**New**

The close function will be invoked on clicking of close button of ActionableNotification.

### DIFF
--- a/src/notification/actionable-notification.component.ts
+++ b/src/notification/actionable-notification.component.ts
@@ -1,7 +1,9 @@
 import {
 	Component,
 	Input,
-	HostBinding
+	HostBinding,
+	Output,
+	EventEmitter
 } from "@angular/core";
 
 import { of } from "rxjs";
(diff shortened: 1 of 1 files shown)


placzkowskim:
Hi @nipun005 ,
I think it is not valid solution, `ActionableNotification `extends `BaseNotification `where 
@Output() close: EventEmitter<any> = new EventEmitter(); is defined.
In my [Issue](https://github.com/carbon-design-system/carbon-components-angular/issues/2724) I described that '_My onNotificationClose() method is beeing properly invoked on 'x' click._. The issue is that notification is not dissapearing on 'x' click.

Akshat55:
Closing this as functionality exists using dynamic notifications.

PR #2759, opened 2024-01-13
fix: display invalid state even when warn state is enabled

Deku-nattsu:
Minor story changes along with consistent state selection across all relevant form elements.

#### Changelog

**Changed**

* Add weight to `invalid` state over `warn` state when both states are enabled.
* Since `Text` is the default input type, it we now query for `ibmText` & `cdsText` _as well_ for projected content. This ensures flexibility and users no longer have to add `input="text"` attribute.

**Chore**

* Story control and option fixes for input and textarea.


Deku-nattsu:

src/input/text-area.directive.ts
```diff
@@ -27,4 +27,13 @@ export class TextArea {
...
 	}
+	constructor(private elementRef: ElementRef<HTMLTextAreaElement>, private renderer: Renderer2) {}
+
+	@Input() set cols(value: number) {
+		if (value) {
+			this.renderer.setStyle(this.elementRef.nativeElement, "width", "");
+		} else {
+			this.renderer.setStyle(this.elementRef.nativeElement, "width", "100%");
+		}
+	}
```
Akshat55, line 38: Not a big fan of this change for two reasons:

1. We should be avoiding inline styling as much as possible in favor of using classes. This makes the code more readable and professional.
2. I don't see any reason to include this as you can simply assign a class to a projected text area to set the width to 100%. In v4, the text area was set to 100% by default but removed in v5 to give teams more customization capability.

Other then this, I think everything makes sense to me. I agree that there should be some level of weight to the 3 states of form items. Setting the state to invalid when the other 2 states (warning/default) are enabled should set the state to invalid.

Akshat55: @zvonimirfras, it would make sense that we apply this change to all form components, not just label and text area. Thoughts on this? I don't think it makes sense to release a fix not applied to all components.

Deku-nattsu: > We should be avoiding inline styling as much as possible in favor of using classes. This makes the code more readable and professional.

you are right, also this can be done in CSS with less boilerplate, and I forgot to forward the cols property to the textarea 

> I don't see any reason to include this as you can simply assign a class to a projected text area to set the width to 100%. In v4, the text area was set to 100% by default but removed in v5 to give teams more customization capability.

true, this can be easily done on the consumer's end anyway, adding this looked reasonable to me because it is the default behavior in web components and react

src/input/input.stories.ts
```diff
@@ -26,6 +26,7 @@ const Template = (args) => ({
 		{{label}}
 		<input
 			cdsText
+			type="text"
```
Akshat55, line 29:
```suggestion
```

The type of the input is text by default, so we don't need to set the type here.

src/input/label.component.ts
```diff
@@ -98,7 +98,7 @@ import { TextInput } from "./input.directive";
 				[attr.data-invalid]="(invalid ? true : null)"
 				#wrapper>
 				<svg
-					*ngIf="!warn && invalid"
+					*ngIf="invalid"
```
Akshat55, line 101: Are you able to apply this change to all of other form components? This is so that the behavior is consistent for all form components.

You can search for `"!warn && invalid"` and you'll find numerous instances of this.

image

Akshat55 (changes requested):

src/input/input.stories.ts
Deku-nattsu: yes, but the textInputContent projection selector requires`input[type=text]`, not setting the type will cause the input element to be placed inside labelContent instead of textInputContent.

src/input/label.component.ts
Deku-nattsu: sure thing, I'll do it as soon I am free.

Akshat55: Great, thanks! 🎉

Deku-nattsu: done :)

src/input/input.stories.ts
Akshat55: You're right, totally forgot about that. Could you add `cdsText` and `ibmText` to the the select options for `text-input-label`? This is to make it more flexible.

So it would be:
`<ng-content select="[cdsText],[ibmText],input[type=text],div"></ng-content>`

Akshat55 (changes requested):
Apart from that minor request, it looks good to go!

src/input/input.stories.ts
Deku-nattsu: done! I did the same with textarea.

Akshat55 (approved):

zvonimirfras (approved):

PR #2789, opened 2024-02-14
fix : null check console error

### LINKED ISSUE #2782: Angular CodeSnippet component gives null check console error
## Title line template: [Angular CodeSnippet component gives null check console error]: 
There is a console error in Angular CodeSnippet component. You can notice that in the storybook as well

image

abiramcodes:
Closes #2782 

Added optional chaining to get rid of the null check error in the console.

### DIFF
--- a/src/code-snippet/code-snippet.component.ts
+++ b/src/code-snippet/code-snippet.component.ts
@@ -231,7 +231,7 @@ export class CodeSnippet extends BaseIconButton implements OnInit, AfterViewInit
 		let ref;
 		switch (this.display) {
 			case "multi":
-				ref = this.codeContent.nativeElement;
+				ref = this.codeContent?.nativeElement;
 				break;
 			case "single":
 				ref = this.codeContainer.nativeElement;
@@ -310,7 +310,7 @@ export class CodeSnippet extends BaseIconButton implements OnInit, AfterViewInit
 
 	protected canExpand() {
 		if (this.display === "multi") {
-			const height = this.codeContent.nativeElement.getBoundingClientRect().height;
+			const height = this.codeContent?.nativeElement.getBoundingClientRect().height;
 			if (
 				this.maxCollapsedNumberOfRows > 0 &&
 				(this.maxExpandedNumberOfRows <= 0 ||
--- a/src/code-snippet/code-snippet.component.ts
+++ b/src/code-snippet/code-snippet.component.ts
@@ -310,7 +310,7 @@ export class CodeSnippet extends BaseIconButton implements OnInit, AfterViewInit
 
 	protected canExpand() {
 		if (this.display === "multi") {
-			const height = this.codeContent?.nativeElement.getBoundingClientRect().height;
+			const height = this.codeContent.nativeElement.getBoundingClientRect().height;
 			if (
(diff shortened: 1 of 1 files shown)


src/code-snippet/code-snippet.component.ts
```diff
@@ -310,7 +310,7 @@ export class CodeSnippet extends BaseIconButton implements OnInit, AfterViewInit
 
 	protected canExpand() {
 		if (this.display === "multi") {
-			const height = this.codeContent.nativeElement.getBoundingClientRect().height;
+			const height = this.codeContent?.nativeElement.getBoundingClientRect().height;
```
Akshat55, line 313:
```suggestion
			const height = this.codeContent.nativeElement.getBoundingClientRect().height;
```

I think there is a bigger underlying issue than this, the view children aren't being picked up.

The biggest issue is that the `#code` directive is within a `ng-template` so it isn't eing picked up by Angular. Could you move that outside of the ng-template?

Before:
```html
		<ng-template #codeTemplate>
			<code #code><ng-content></ng-content></code>
		</ng-template>
```


After:
```
		<ng-template #codeTemplate>
			<ng-content></ng-content>
		</ng-template>
```

Then you will have to update all instances where that template is used and wrap it in `code` tag as such:
Inline with no copy:
```html
					<code #code>
						<ng-container *ngTemplateOutlet="codeTemplate"></ng-container>
					</code>
```

Multi:
```html
				<pre
					#codeContent
					*ngIf="!skeleton"
					(scroll)="(display === 'multi' ? handleScroll() : null)"><code #code><ng-container *ngTemplateOutlet="codeTemplate"></ng-container></code></pre>
```

Inline:
```html
					<code #code>
						<ng-container *ngTemplateOutlet="codeTemplate"></ng-container>
					</code>
```

src/code-snippet/code-snippet.component.ts
```diff
@@ -231,7 +231,7 @@ export class CodeSnippet extends BaseIconButton implements OnInit, AfterViewInit
 		let ref;
 		switch (this.display) {
 			case "multi":
-				ref = this.codeContent.nativeElement;
+				ref = this.codeContent?.nativeElement;
```
Akshat55, line 234:
```suggestion
				ref = this.codeContent.nativeElement;
```

Same as comment below. This _may_ be required.

Akshat55 (changes requested):
Component seemed to work but due to Angular update, `#code` directive isn't being picked up by Angular. So there is a bigger underlying issue. 

Let's make the following changes I've requested first, then we can improve it even further!

src/code-snippet/code-snippet.component.ts
Akshat55: Sorry, I was on vacation - I don't see any of the files changed in the push 😅

Akshat55: Could you push these changes and I can suggest additional changes/make updates

abiramcodes: Hey, hope you had a great vacation!
Sure, pushed my changes now.

Akshat55 (changes requested):

Akshat55 (approved):

PR #2791, opened 2024-02-16
chore: onboard carbon-components-angular to IBM Telemetry 🚀

francinelucca:
### Description

Adds config file and dependency necessary to start tracking telemetry data via [IBM Telemetry](https://github.com/ibm-telemetry/telemetry-js)

### Changelog

**New**

- Adds @ibm/telemetry-js dependency
- Adds [telemetry.yml](telemetry.yml) config file
    - This file was generated using [IBM Telemetry automated config generator](https://github.com/ibm-telemetry/telemetry-js-config-generator/pull/1)
- Adds readme telemetry disclosure
- Adds postinstall script to use IBM telemetry command

**Removed**
- Removes `@carbon/telemetry` dependency

### Review
Please look through package.json config file and ensure all necessary modifications have been made so that the "telemetry.yml" config file is included in the release version of this package.

**PLEASE NOTE:** In order for IBM Telemetry to start collecting data for this project, a new build must be published including these changes.


package.json
```diff
@@ -137,8 +138,8 @@
   "dependencies": {
     "@carbon/icon-helpers": "10.37.0",
     "@carbon/icons": "11.14.0",
-    "@carbon/telemetry": "0.1.0",
     "@carbon/utils-position": "1.1.4",
+    "@ibm/telemetry-js": "^1.2.1",
```
Akshat55, line 142:
```suggestion
    "@ibm/telemetry-js": "1.2.1",
```

Can we version lock this package? This is more of a safe measure in case there is a breaking change made without a major version bump.  Will need to run `npm install` to apply these changes to package-lock as well.

Akshat55 (changes requested):
This is exciting, we were looking forward to this data! Just a few questions and requested changes.

package.json
jdharvey-ibm: Hey, @Akshat55! You're certainly more than welcome to pin the IBM Telemetry version to a specific release if you'd like. I will say though that by design, we (the IBM Telemetry team) enforce that our code be zero-impact to any library instrumented with it. If our code breaks or fails, our fundamental design principle is to fail silently. We will not impact any build/CI/install processes under any circumstances even if we encounter errors when running. We accomplish this by running our data collection in a detached process which (regardless of its exit code) will always be silent. In the event that we do have a breaking change (maybe a config file format change or a collection regression bug), that will not impact you or your library's consumers at all.

Looking at your `package.json` file, it looks like your team's convention is to pin all deps, and if this is your preferred approach, it's totally fine by us. I just wanted to offer some peace of mind about how the telemetry tooling operates.

If you'd like @francinelucca or I to pin the dep, just let us know and we can make that adjustment, or if you'd like to make that change yourself, that's fine too!

Akshat55: Thank you for the clarification @jdharvey-ibm!

Akshat55 (approved):

mattrosno:
Hi @zvonimirfras - can we get a review on this PR so we can start collecting telemetry data?

zvonimirfras (approved):

Issue #2705, opened 2023-10-04
Tooltip displayed when mouse leaves and not when the mouse enters while using change detection OnPush


kcpradeep09:
## Detailed description

> Describe in detail the issue you're having.
If I use the tooltip in a component which has change detection strategy as OnPush

> Is this a feature request (new component, new icon), a bug, or a general issue?
Bug

> Is this issue related to a specific component?
Yes, Tooltip

> What did you expect to happen? What happened instead? What would you like to see changed?
The tooltip should be displayed when the mouse enters and not when the mouse leaves.

> What browser are you working in?
Chrome

> What version of the Carbon Design System are you using?
 "carbon-components-angular": "^5.14.1",
 "@carbon/styles": "^1.35.0",
 "@carbon/icons": "^11.27.0",
 "@angular/core": "^16.1.0",

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

## Steps to reproduce the issue

However over the action items at top right on the UI Shell header
https://codesandbox.io/p/sandbox/recursing-ardinghelli-9rwssc?file=%2Fsrc%2Fapp%2Fapp.component.html%3A12%2C42

Akshat55:
None of our components currently support OnPush strategy, this would require an overhaul of all components where all components would need to be moved to the strategy.

Geoffrey0:
@Akshat55 Would you accept a PR for adding OnPush to specifically the tooltip? As this is a component that can be used in other components and it heavily relies on events rather than inputs/outputs.

PR #2800, opened 2024-02-27
feat: allow tooltip and toggletip to be used in an OnPush component

Geoffrey0:
Closes carbon-design-system/carbon-components-angular#2705

This change allows the tooltip, definition-tooltip and toggletip to be used in OnPush enabled components. Components that are using the default change detection strategy should not be impacted.

#### Changelog

**Changed**

* Tooltip has been changed to work with OnPush change detection strategy.
* Definition Tooltip has been changed to work with OnPush change detection strategy.
* Toggletip has been changed to work with OnPush change detection strategy.

### DIFF
--- a/src/popover/popover.directive.ts
+++ b/src/popover/popover.directive.ts
@@ -96,11 +96,7 @@ export class PopoverContainer {
 		"left" | "left-bottom" | "left-top" |
 		"right" | "right-bottom" | "right-top" = "bottom";
 
-	private changeDetectorRef: ChangeDetectorRef;
-
-	constructor(ref?: ChangeDetectorRef) {
-		this.changeDetectorRef = ref;
-	}
+	constructor(private changeDetectorRef: ChangeDetectorRef) {}
 
 	handleChange(open: boolean, event: Event) {
 		if (this.isOpen !== open) {
@@ -113,6 +109,6 @@ export class PopoverContainer {
 			this.onClose.emit(event);
 		}
 		this.isOpen = open;
-		this.changeDetectorRef?.markForCheck();
+		this.changeDetectorRef.markForCheck();
 	}
 }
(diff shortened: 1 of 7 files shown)


Geoffrey0:

src/popover/popover.directive.ts
```diff
@@ -95,6 +96,12 @@ export class PopoverContainer {
 		"left" | "left-bottom" | "left-top" |
 		"right" | "right-bottom" | "right-top" = "bottom";
 
+	private changeDetectorRef: ChangeDetectorRef;
+
+	constructor(ref?: ChangeDetectorRef) {
+		this.changeDetectorRef = ref;
+	}
```
Akshat55, line 103: Why do we need to make the dependency injection optional?

```suggestion
	constructor(private changeDetectorRef: ChangeDetectorRef) {}
```

src/popover/popover.directive.ts
```diff
@@ -106,5 +113,6 @@ export class PopoverContainer {
 			this.onClose.emit(event);
 		}
 		this.isOpen = open;
+		this.changeDetectorRef?.markForCheck();
```
Akshat55, line 116: As per above changes.

```suggestion
		this.changeDetectorRef.markForCheck();
```

Akshat55 (changes requested):
Thanks for contributing this! 🚀

Apart from the optional DI, it seems good to me!

src/popover/popover.directive.ts
Geoffrey0: I could make it required for sure. All component that will extend this will then have to pass it. I left the option to make it optional if some component didn't need "want" to implement OnPush. On Angular 16 we could solve this in a nicer way by just using DI with the inject function not poluting the constructor. What do you think is the best approach here?

Akshat55: I think it makes sense to add it as a required dependency, Ill be injecting additional dependencies sometimes next week that will be required when I integrate floating-ui with popover.

Akshat55 (approved):

zvonimirfras (approved):

Issue #2811, opened 2024-03-12
Sidenav route implementation doesn't use routerLink


klaascuvelier:
## Detailed description

The current implementation of the sidenav component does not use `routerLink` (and `routerLinkActive`) when using the `route` input. Instead it add an eventListener on the `a` tag and uses the `router` to do navigation.

One downside of this is that the `a` tag in DOM does not get the right `href` but instead keeps `#`. This doesn't allow for functionality like opening in new tab, copying the link, ...

Proper usage of `routerLink` would solve this.

> Is this a feature request (new component, new icon), a bug, or a general issue?

Feature request.

> Is this issue related to a specific component?

Sidenav

PR #2812, opened 2024-03-12
feat: add useRouter property to sidenav

klaascuvelier:
Closes carbon-design-system/carbon-components-angular#2811

This MR adds a `useRouter` option to the SideNav component (which is `false` by default). 
When the `input` is `true`, `routerLink` will be used instead.

#### Changelog

**New**

* adds `useRouter` input to `SideNav`

### DIFF
--- a/src/ui-shell/sidenav/sidenav-item.component.ts
+++ b/src/ui-shell/sidenav/sidenav-item.component.ts
@@ -35,8 +35,7 @@ import { Router } from "@angular/router";
 				[routerLinkActive]="'cds--side-nav__item--active'"
 				[ariaCurrentWhenActive]="'page'"
 				[attr.title]="title ? title : null"
-				class="cds--side-nav__link"
-			>
+				class="cds--side-nav__link">
 				<ng-template [ngTemplateOutlet]="sidenavItemContentTpl"></ng-template>
 			</a>
 		</ng-template>
--- a/src/ui-shell/sidenav/sidenav-item.component.ts
+++ b/src/ui-shell/sidenav/sidenav-item.component.ts
@@ -32,8 +32,8 @@ import { Router } from "@angular/router";
 		<ng-template #sidenavItemRouterTpl>
 			<a
 				[routerLink]="route"
-				[routerLinkActive]="'cds--side-nav__item--active'"
-				[ariaCurrentWhenActive]="'page'"
+				routerLinkActive="cds--side-nav__item--active"
+				ariaCurrentWhenActive="page"
 				[attr.title]="title ? title : null"
 				class="cds--side-nav__link">
 				<ng-template [ngTemplateOutlet]="sidenavItemContentTpl"></ng-template>
--- a/src/ui-shell/sidenav/sidenav-item.component.ts
+++ b/src/ui-shell/sidenav/sidenav-item.component.ts
@@ -124,8 +124,7 @@ export class SideNavItem implements OnChanges {
 
 	protected _href = "#";
 
-	constructor(protected domSanitizer: DomSanitizer, @Optional() protected router: Router) {
-	}
+	constructor(protected domSanitizer: DomSanitizer, @Optional() protected router: Router) {}
 
 	ngOnChanges(changes: SimpleChanges) {
 		if (changes.active) {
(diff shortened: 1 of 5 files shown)


klaascuvelier:
recheck

klaascuvelier:
recheck

klaascuvelier:
recheck

Akshat55 (changes requested):
Apologies for the late review, I've been juggling multiple projects this month 😓 Overall PR is good to go, just minor linting changes.

Im targeting a release for this PR early next week.

klaascuvelier:
Thanks for the review @Akshat55!  Took care of your remarks. 

Quick question, did you do manual linting here? If so, is there a possibility to automate some things? Like adding eslint or prettier to the repo maybe?

Akshat55:
> Quick question, did you do manual linting here? If so, is there a possibility to automate some things? Like adding eslint or prettier to the repo maybe?

We do some manual linting here and there while reviewing the code. We do have a linter (tslint) and we run it prior to pushing the commit (using husky). 

We haven't updated to eslint in v5 because we noticed it was taking way too long. From what I've seen, the duration has decreased and we plan to replace our existing linter with eslint in v6!

src/ui-shell/sidenav/sidenav-item.component.ts
```diff
@@ -16,22 +17,37 @@ import { Router } from "@angular/router";
...
+			<ng-template [ngTemplateOutlet]="sidenavItemContentTpl"></ng-template>
+		</a>
+
+		<ng-template #sidenavItemRouterTpl>
+			<a
+				[routerLink]="route"
+				[routerLinkActive]="'cds--side-nav__item--active'"
+				[ariaCurrentWhenActive]="'page'"
+				[attr.title]="title ? title : null"
+				class="cds--side-nav__link">
```
Akshat55, line 38: I don't think we need to bind routerlinkactive & ariaCurrentWhenActive here. We should be able to just do:

```suggestion
				[routerLink]="route"
				routerLinkActive="cds--side-nav__item--active"
				ariaCurrentWhenActive="page"
				[attr.title]="title ? title : null"
				class="cds--side-nav__link">
```

src/ui-shell/ui-shell.stories.ts
```diff
@@ -303,6 +309,51 @@ const SideNavigationRouterTemplate = (args) => ({
...
+			    position: absolute;
+    			background-color: var(--cds-border-interactive, #0f62fe);
+    			content: "";
+    			inline-size: 3px;
+    			inset-block-end: 0;
+    			inset-block-start: 0;
+    			inset-inline-start: 0;
+    		}
+		}
+	`],
```
Akshat55, line 329: I don't think we need this. After looking at @carbon/styles, seems like the `active` class is assigned when `aria-current="page"`. Could you please verify this when router is enabled?

Akshat55 (changes requested):

src/ui-shell/sidenav/sidenav-item.component.ts
klaascuvelier: You're absolutely right

src/ui-shell/ui-shell.stories.ts
klaascuvelier: You are also correct, this was probably an issue before I correctly added the `aria-current` thanks for noticing!

klaascuvelier:
> > Quick question, did you do manual linting here? If so, is there a possibility to automate some things? Like adding eslint or prettier to the repo maybe?
> 
> We do some manual linting here and there while reviewing the code. We do have a linter (tslint) and we run it prior to pushing the commit (using husky).
> 
> We haven't updated to eslint in v5 because we noticed it was taking way too long. From what I've seen, the duration has decreased and we plan to replace our existing linter with eslint in v6!

Ah right, noticed a warning in my console when pushing/committing. Prolly I should look into resolving that.

Akshat55 (approved):

Licen-it:
This PR enabled the routing by default, but didn't take in consideration the `routeExtras` passed to the component (which is breaking anyone using the `relativeTo` option)

Issue #2832, opened 2024-03-28
DatePicker range is getting erase while closing calendar


kush-savani:
Steps to produce:
- Open date-picker range
- Select any dates for start and end date.
- click on start date input. 
- click outside of calendar dialog. (selected date will erase)

https://github.com/carbon-design-system/carbon-components-angular/assets/49565062/997afb61-ebfd-488e-9717-93d310ae3758

Expected behaviour:

https://github.com/carbon-design-system/carbon-components-angular/assets/49565062/470931b9-7e75-482b-81cf-91c0b19ac3da

PR #2845, opened 2024-04-02
fix(date-picker): manage to reselect date while closing calendar

kush-savani:
Closes https://github.com/carbon-design-system/carbon-components-angular/issues/2832

On calendar dialog close, if user didn't select any value or close accidentally, we are selecting empty value. 
With current changes, I remove that logic. now if we don't get any selected value we will pass last selected value.

## Before
https://github.com/carbon-design-system/carbon-components-angular/assets/49565062/997afb61-ebfd-488e-9717-93d310ae3758

## After

https://github.com/carbon-design-system/carbon-components-angular/assets/49565062/ce981f91-0458-4994-9858-62a4592ade28


src/datepicker/datepicker.component.ts
```diff
@@ -269,13 +269,6 @@ export class DatePicker implements
 			// This makes sure that the `flatpickrInstance selectedDates` are in sync with the values of
 			// the inputs when the calendar closes.
 			if (this.range && this.flatpickrInstance) {
-				if (this.flatpickrInstance.selectedDates.length !== 2) {
-					// we could `this.flatpickrInstance.clear()` but it insists on opening the second picker
-					// in some cases, so instead we do this
-					this.setDateValues([]);
```
Akshat55, line 275: Perhaps, if only 1 date is set of the 2, we should set the start and end date to be the same? This is the behavior I'm seeing in React.

https://react.carbondesignsystem.com/?path=/story/components-datepicker--range-with-calendar

1. Click on the date picker
2. Select a date
3. Click outside of the date picker and notice that the same date is assigned to the start/end range.

Akshat55 (changes requested):

src/datepicker/datepicker.component.ts
kush-savani: I update condition part. now it will look like this.

https://github.com/carbon-design-system/carbon-components-angular/assets/49565062/ab959005-a1cc-484a-87f0-293bc96ada01

rajeshkumaru (approved):

PR #2849, opened 2024-04-07
fix: resolve cannot read property of undefined when all list items are disabled

### LINKED ISSUE #2830: [BUG]:  Dropdown Component - 'Cannot read properties of undefined' when all the options are disabled
## Detailed description

When all the options (items) provided to dropdown are disabled, there is an error popping up in the console.

```
ERROR TypeError: Cannot read properties of undefined (reading 'nativeElement')
    at DropdownList.getNextElement (carbon-components-angular-dropdown.mjs:1391:37)
    at DropdownList.updateIndex (carbon-components-angular-dropdown.mjs:1522:18)
    at DropdownList.updateList (carbon-components-angular-dropdown.mjs:1316:14)
    at set items (carbon-components-angular-dropdown.mjs:1266:18)
```

The error is happening in method `getNextElement` in [dropdown-list.component.ts](https://github.com/carbon-design-system/carbon-components-angular/blob/master/src/dropdown/list/dropdown-list.component.ts#L335-L356)

## Cause of the bug

The initial value of `this.index` is `-1`, and since all the `this.displayItems` are having disabled to true, the index value is not getting modified. So, accessing `nativeElement` from `elemList[-1]` will be undefined, the error is happening.

Aru-Ku:
Closes #2830 

Resolves the error, on reading nativeElement when all the list items are disabled.

For details info on the error, please refer to the issue #2830

### DIFF
--- a/src/dropdown/list/dropdown-list.component.ts
+++ b/src/dropdown/list/dropdown-list.component.ts
@@ -352,7 +352,7 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 			}
 		}
 
-		return elemList[this.index].nativeElement;
+		return elemList[this.index]?.nativeElement;
 	}
 
 	/**
@@ -490,6 +490,7 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 	updateIndex() {
 		// initialize index on the first selected item or
 		// on the next non disabled item if no items are selected
+		// in case, if all items are disabled, the index value will remain same
 		const selected = this.getSelected();
 		if (selected.length) {
 			this.index = this.displayItems.indexOf(selected[0]);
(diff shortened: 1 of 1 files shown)


src/dropdown/list/dropdown-list.component.ts
```diff
@@ -519,7 +520,7 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 				}
 			} else if (event.key === "ArrowUp") {
 				if (this.hasPrevElement()) {
-					this.getPrevElement().scrollIntoView({  block: "nearest" });
+					this.getPrevElement()?.scrollIntoView({  block: "nearest" });
```
Akshat55, line 523: Based on the change you made to `getNextElement()`, shouldn't optional chaining be used on line 517 instead of 523?

Akshat55 (changes requested):

src/dropdown/list/dropdown-list.component.ts
Aru-Ku: Hi @Akshat55 
Yes, it went unnoticed. Fixed it now.

Akshat55 (approved):

Issue #2851, opened 2024-04-08
bug: SidenavItem component has useRouter input set to true by default


klaascuvelier:
`sidenav-item.component.ts` has `useRouter` `input` accidentally set to true by default. This breaks the current implementation.  
Introduced by #2812

PR #2852, opened 2024-04-08
fix: sidenavItem component useRouter should be false by default

klaascuvelier:
Closes carbon-design-system/carbon-components-angular#2851

fixed default value for `useRouter` in `sidenav-item.component.ts`

#### Changelog

**Changed**

* fixed default value for `useRouter` in `sidenav-item.component.ts`

### DIFF
--- a/src/ui-shell/sidenav/sidenav-item.component.ts
+++ b/src/ui-shell/sidenav/sidenav-item.component.ts
@@ -74,7 +74,7 @@ export class SideNavItem implements OnChanges {
 	/**
 	 * Use the routerLink attribute on <a> tag for navigation instead of using event handlers
 	 */
-	@Input() useRouter = true;
+	@Input() useRouter = false;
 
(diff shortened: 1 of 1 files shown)


Akshat55 (approved):

Akshat55:
Nice catch, that totally went past me 😓

klaascuvelier:
Thanks for the quick review @Akshat55 
I could make a test case for this but I don't know if there's much value in it. Let me know if you'd still want that.

Akshat55:
@klaascuvelier, it should be fine, no need for a test for this case.

klaascuvelier:
> Nice catch, that totally went past me 😓

Same. Noticed styles were broken in storybook and started investigating. Shouldn't have happened, apologies.

Akshat55:
All good! Thanks for quickly creating a PR to fix this.

Issue #2853, opened 2024-04-09
Tooltip component no longer exposing `data` property to provide data to the template


Licen-it:
## Detailed description

> Describe in detail the issue you're having.

Migrating from v4 to v5 I noticed that the component replacing `ibmTooltip` is not exposing the `data` input (that was inherited from `Dialog`) to provide some data to the template when `description` is a TemplateRef

> Is this a feature request (new component, new icon), a bug, or a general issue?

bug

> Is this issue related to a specific component?

<cds-tooltip>

> What did you expect to happen? What happened instead? What would you like to see changed?

to have the `data` input back and used to pass info to the TemplateRef when rendered

> What browser are you working in?

any

> What version of the Carbon Design System are you using?

latest v5

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

I'm migrating from v4 to v5 and this is causing a lot of duplicated code. I have till the end of this week to complete the migration.

PR #2854, opened 2024-04-10
feat(Tooltip): expose data input to be passed as template outlet context when rendering the tooltip

Licen-it:
Closes carbon-design-system/carbon-components-angular#2853

Reintroduce `data` input to be passed as `ngTemplateOutletContext` when `description` is a template

#### Changelog

**New**

* added `data` input to Tooltip and TooltipDesctiption

**Changed**

* fixed IconButton stories that were showing incorrect input label (`cdsButton` instead of `kind`)

### DIFF
--- a/src/tooltip/definition-tooptip.component.ts
+++ b/src/tooltip/definition-tooptip.component.ts
@@ -41,7 +41,7 @@ import { PopoverContainer } from "carbon-components-angular/popover";
 			role="tooltip">
 			<span class="cds--popover-content cds--definition-tooltip">
 				<ng-container *ngIf="!isTemplate(description)">{{description}}</ng-container>
-				<ng-template *ngIf="isTemplate(description)" [ngTemplateOutlet]="description"></ng-template>
+				<ng-template *ngIf="isTemplate(description)" [ngTemplateOutlet]="description" [ngTemplateOutletContext]="{ data }"></ng-template>
 				<span *ngIf="autoAlign" class="cds--popover-caret cds--popover--auto-align"></span>
 			</span>
 			<span *ngIf="!autoAlign" class="cds--popover-caret"></span>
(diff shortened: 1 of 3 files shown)


src/tooltip/definition-tooptip.component.ts
```diff
@@ -57,6 +57,10 @@ export class TooltipDefinition extends PopoverContainer {
 	 * The string or template content to be exposed by the tooltip.
 	 */
 	@Input() description: string | TemplateRef<any>;
+	/**
+	 * Optional data for templates
+	 */
+	@Input() data = {}
```
Akshat55, line 63: Let's use `context` or perhaps `templateContext` to make it more descriptive.

```suggestion
	/**
	 * Optional data for templates
	 */
	@Input() context: any;
```

src/tooltip/tooltip.component.ts
```diff
@@ -69,6 +69,10 @@ export class Tooltip extends PopoverContainer implements AfterContentChecked {
 	 * The string or template content to be exposed by the tooltip.
 	 */
 	@Input() description: string | TemplateRef<any>;
+	/**
+	 * Optional data for templates
+	 */
+	@Input() data = {}
```
Akshat55, line 75: Same as above

src/tooltip/definition-tooptip.component.ts
```diff
@@ -41,7 +41,7 @@ import { PopoverContainer } from "carbon-components-angular/popover";
 			role="tooltip">
 			<span class="cds--popover-content cds--definition-tooltip">
 				<ng-container *ngIf="!isTemplate(description)">{{description}}</ng-container>
-				<ng-template *ngIf="isTemplate(description)" [ngTemplateOutlet]="description"></ng-template>
+				<ng-template *ngIf="isTemplate(description)" [ngTemplateOutlet]="description" [ngTemplateOutletContext]="{ data }"></ng-template>
```
Akshat55, line 44: What are your thoughts on this implicit?

Akshat55 (changes requested):

src/tooltip/definition-tooptip.component.ts
Licen-it: I was following probably the old pattern, but it would make sense to have it `$implicit`

Akshat55 (approved):

Issue #2861, opened 2024-04-19
Toast notification no longer showing close button by default


Licen-it:
## Detailed description

> Describe in detail the issue you're having.

After migrating from v4 to v5, I noticed Toast notification is no longer filling missing configurations with default values, resulting in no longer showing close button by default

> Is this a feature request (new component, new icon), a bug, or a general issue?

bug

> Is this issue related to a specific component?

Toast notification

> What did you expect to happen? What happened instead? What would you like to see changed?

To inherit default configuration values when not provided

> What browser are you working in?

any

> What version of the Carbon Design System are you using?

latest

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Zelta

PR #2864, opened 2024-04-19
fix: merge default notification obj with toast content when set

Akshat55:
Closes carbon-design-system/carbon-components-angular#2861

#### Changelog

**Changed**
* Merge default notification obj with toast content when set. This matches the behavior of actionable and inline notification.

### DIFF
--- a/src/notification/toast.component.ts
+++ b/src/notification/toast.component.ts
@@ -5,6 +5,7 @@ import {
 	HostBinding
 } from "@angular/core";
 
+import { of } from "rxjs";
 import { ToastContent } from "./notification-content.interface";
 import { NotificationDisplayService } from "./notification-display.service";
 import { I18n } from "carbon-components-angular/i18n";
@@ -49,7 +50,17 @@ export class Toast extends BaseNotification implements OnInit {
 	 *
 	 * `type` can be one of `"error"`, `"info"`, `"info-square"`, `"warning"`, `"warning-alt"`, or `"success"`
 	 */
-	@Input() notificationObj: ToastContent;
+	@Input() set notificationObj(obj: ToastContent) {
+		if (obj.closeLabel) {
+			obj.closeLabel = of(obj.closeLabel);
+		}
+		this._notificationObj = Object.assign({}, this.defaultNotificationObj, obj);
+	}
+
+	get notificationObj(): ToastContent {
+		return this._notificationObj as ToastContent;
+	}
+
 
 	@HostBinding("attr.id") toastID = `toast-${Toast.toastCount++}`;
 	@HostBinding("class.cds--toast-notification") toastClass = true;


src/notification/toast.component.ts
```diff
@@ -49,7 +50,17 @@ export class Toast extends BaseNotification implements OnInit {
 	 *
 	 * `type` can be one of `"error"`, `"info"`, `"info-square"`, `"warning"`, `"warning-alt"`, or `"success"`
 	 */
-	@Input() notificationObj: ToastContent;
+	@Input() set notificationObj(obj: ToastContent) {
+		if (obj.closeLabel) {
+			obj.closeLabel = of(obj.closeLabel);
```
Licen-it, line 55: You would require the `| async` in the template where it's using the label...but tbh i tried to understand why it was introduced that way, and i can't get it...it seems and attempt to have closeLabel as observable as well as string, but in this way if i set an observable in closeLabel, it will only be an `Observable<Observable<string>>`

Akshat55: Good catch, thanks!

When we pass a non-observable parameter in `of`. It creates a observable and emits the non-observable value. When we pass an observable as a parameter in `of`, it subscribes to it and emits the value. 

So by using `of` operator, we don't have to check the content if closeLabel is an observable or a string.

Akshat55: You're right, scratch my last statement, it wraps the observable within another observable, so subscribing to it would emit an observable.

So from what I'm seeing in the v10 code, we do NOT support observable for `closeLabel`. We have the `of` in place to convert the user passed `closeLabel` value to an observable which is unwrapped by the `async` pipe.

We DO use the observable initially for defaultNotificationObj which fetches translations and returns a BehaviorSubject:
```ts
// Returns a STRING or an Observable
closeLabel: this.i18n.get("NOTIFICATION.CLOSE_BUTTON"),
```

Looking at the [PR that made this change](https://github.com/carbon-design-system/carbon-components-angular/pull/299), `closeLabel` initially had a type of `string` which was then set to `any` .

We can change the setter to the following to support observables, but the question is do we really want to support this?
```ts
	set notificationObj(obj: NotificationContent) {
		if (obj.closeLabel && !isObservable(obj.closeLabel)) {
			obj.closeLabel = of(obj.closeLabel);
		}
		this._notificationObj = Object.assign({}, this.defaultNotificationObj, obj);
	}
```

We can set the type of `closeLabel` to string and set the initial translated value in BaseNotification constructor.

@zvonimirfras thoughts?

Licen-it: any update on this @Akshat55 ?

Akshat55: Hey @Licen-it, this is planned for end of this week. I'll be picking it up sometime tomorrow. We plan on removing `async` pipe altogether for notification and setting the type in the interface to string. Just need to run a few tests before making a these changes.

Akshat55: Decided to check for observables, if user passes in an observable we don't use the `of` operator. We don't have to manually subscribe then unsubscribe if we use the `async` pipe. Depending on the usage of the translate service, we may remove it in distant future.... I don't see a reason to use it over ngx-translate or other libs.

Licen-it (approved):

zvonimirfras (approved):

PR #2875, opened 2024-05-08
fix: copy inner text or text content of code element

Akshat55:
Issue went unnoticed due to storybook creating code argument in global scope which was accessible by the component.

#### Changelog

**Changed**

* Now text content is correctly copied to clipboard from code element
* **MISC**: Docs now use different variable name than the one used in component to avoid referencing value in global scope.


Akshat55:
We cannot write tests for this since we run into issues related to permissions, (`Write permission denied.`). I'm assuming this is why there was no test case for this, Couldn't find any information on how to override this.

zvonimirfras (approved):

Issue #2904, opened 2024-05-27
Sidenav Item is ignoring routeExtras


Licen-it:
## Detailed description

> Describe in detail the issue you're having.

the change introduced in https://github.com/carbon-design-system/carbon-components-angular/pull/2812 is enabling the router by default, but is not passing the `routeExtras`, resulting in breaking links for who was relying on those extras (like `relativeTo`)

> Is this a feature request (new component, new icon), a bug, or a general issue?

bug

> Is this issue related to a specific component?

Sidenav Item

> What did you expect to happen? What happened instead? What would you like to see changed?

`routeExtras` should be properly consumed and passing the single option to the `routerLink` directive

> What browser are you working in?

any

> What version of the Carbon Design System are you using?

latest

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Zelta

PR #2905, opened 2024-05-27
fix: pass route extras in routerLink

Akshat55:
Closes carbon-design-system/carbon-components-angular#2904

#### Changelog

**Changed**

* Now calling `navigate` function on anchor link click which will pass in routeExtras.

### DIFF
--- a/src/ui-shell/sidenav/side-nav.component.spec.ts
+++ b/src/ui-shell/sidenav/side-nav.component.spec.ts
@@ -18,7 +18,7 @@ class FooComponent { }
 @Component({
 	template: `
 		<cds-sidenav [allowExpansion]="allowExpansion" [hidden]="hidden">
-			<cds-sidenav-menu title="Example Title"></cds-sidenav-menu>
+			<cds-sidenav-menu title="Example Title"></cds-sidenav-menu>
 			<cds-sidenav-item
 				[route]="route"
 				[useRouter]="useRouter"
@@ -38,7 +38,7 @@ class SideNavTest {
 	}
 }
 
-describe("SideNav", () => {
+fdescribe("SideNav", () => {
 	let fixture, wrapper, element;
 	beforeEach(() => {
 		TestBed.configureTestingModule({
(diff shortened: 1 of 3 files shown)


src/ui-shell/sidenav/side-nav.component.spec.ts
```diff
@@ -38,7 +38,7 @@ class SideNavTest {
 	}
 }
 
-describe("SideNav", () => {
+fdescribe("SideNav", () => {
```
Licen-it, line 41:
```suggestion
describe("SideNav", () => {
```

Licen-it (changes requested):

Licen-it (approved):

klaascuvelier:
Hey, I am a bit late here with feedback as this MR has been merged already. 

I think the current implementation effectively breaks the idea of the original change. The idea was to rely on the functionality of the directives solely to do navigation and navigation state checks, skipping any custom JS. This would allow for correct cmd + click behavior and so on.
I think the correct fix here would have been to add the navigation extras into the template where now the onclick handler was added.

Licen-it:
@klaascuvelier yeah it would make sense....i didn't think about the cmd + click, but the only pain would be that then passed into the template they wouldn't be a nice since object, but each property would need to be specified separately.
I have an idea @Akshat55, but not sure if that would be ok for carbon team...we could have a directive extending RouterLink directive which is accepting a single object of type NavigationExtras and in the constructor setting each property in this object to the parent class (in a loop way so that if they add/remove any, you won't have to update this code)

klaascuvelier:
Thanks for you reply @Licen-it.
I think adding _all_ the navigation extras in the template of this component should not be too much work, just a one time job.
I'll draft something up quickly asap.

Licen-it:
> Thanks for you reply @Licen-it. I think adding _all_ the navigation extras in the template of this component should not be too much work, just a one time job. I'll draft something up quickly asap. I'll start by creating a new issue so we can take the discussion there. Thanks for collaborating on this and my apologies for breaking stuff :)

Well I was mainly thinking about future proof solution, so that even if a later angular version is changing it will work fine...something to easily support different versions of angular without breaking

Issue #2911, opened 2024-06-05
[bug] routeExtras in useRouter of side navigation breaks default navigation


klaascuvelier:
The fix for #2904 done in #2905 breaks desired behavior. By adding a `click` listener javascript is added and the default browser behavior (like command + click and other behavior) is broken. 

The easiest way to restore this functionality and support navigation extras, is to add all individual extras in the template. 
This is however (as pointed out by @Licen-it in #2905) prone to bugs if Angular would introduce new properties in the navigation extras. 
I would however still go forward with this suggestion to unblock the current bug and then look at a more longterm solution afterwards

PR #2912, opened 2024-06-05
fix: remove click handler for the anchorlink when useRouter is true

klaascuvelier:
Closes carbon-design-system/carbon-components-angular#2911

Removes the click handler for the element when using `useRouter`. Adds in all available routeExtras so they get passed along.

### DIFF
--- a/src/ui-shell/sidenav/sidenav-item.component.ts
+++ b/src/ui-shell/sidenav/sidenav-item.component.ts
@@ -31,11 +31,19 @@ import { Router } from "@angular/router";
 
 		<ng-template #sidenavItemRouterTpl>
 			<a
-				[routerLink]="route"
-				routerLinkActive="cds--side-nav__item--active"
-				(click)="navigate($event)"
-				ariaCurrentWhenActive="page"
 				[attr.title]="title ? title : null"
+				[routerLink]="route"
+ 				[relativeTo]="routeExtras?.relativeTo"
+				[queryParams]="routeExtras?.queryParams"
+				[fragment]="routeExtras?.fragment"
+				[queryParamsHandling]="routeExtras?.queryParamsHandling"
+				[preserveFragment]="routeExtras?.preserveFragment"
+				[skipLocationChange]="routeExtras?.skipLocationChange"
+				[replaceUrl]="routeExtras?.replaceUrl"
+				[state]="routeExtras?.state"
+				[info]="routeExtras?.info"
+				routerLinkActive="cds--side-nav__item--active"
+				ariaCurrentWhenActive="page"
 				class="cds--side-nav__link">
 				<ng-template [ngTemplateOutlet]="sidenavItemContentTpl"></ng-template>
 			</a>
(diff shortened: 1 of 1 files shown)


src/ui-shell/sidenav/sidenav-item.component.ts
```diff
@@ -31,11 +31,20 @@ import { Router } from "@angular/router";
...
 		<ng-template #sidenavItemRouterTpl>
 			<a
+				[attr.title]="title ? title : null"
 				[routerLink]="route"
+ 				[relativeTo]="routeExtras?.relativeTo"
+				[queryParams]="routeExtras?.queryParams"
+				[fragment]="routeExtras?.fragment"
+				[queryParamsHandling]="routeExtras?.queryParamsHandling"
+				[preserveFragment]="routeExtras?.preserveFragment"
+				[onSameUrlNavigation]="routeExtras?.onSameUrlNavigation"
```
Licen-it, line 41: `onSameUrlNavigation` is the only one that is not defined by the `routerLink` directive...it seems to be a property only for `router.navigate` which is overriding the default configuration for the router

klaascuvelier: Good catch!

src/ui-shell/sidenav/sidenav-item.component.ts
```diff
@@ -31,11 +31,19 @@ import { Router } from "@angular/router";
...
 				[routerLink]="route"
+ 				[relativeTo]="routeExtras?.relativeTo"
+				[queryParams]="routeExtras?.queryParams"
+				[fragment]="routeExtras?.fragment"
+				[queryParamsHandling]="routeExtras?.queryParamsHandling"
+				[preserveFragment]="routeExtras?.preserveFragment"
+				[skipLocationChange]="routeExtras?.skipLocationChange"
+				[replaceUrl]="routeExtras?.replaceUrl"
+				[state]="routeExtras?.state"
+				[info]="routeExtras?.info"
```
Akshat55, line 44:
```suggestion
```

@klaascuvelier Considering we are using Angular 14, NavigationExtras does not have `info` option. We should remove it for now and add it if it is requested by users. This is a just in case measure to prevent breaking existing apps.

Akshat55 (changes requested):

src/ui-shell/sidenav/sidenav-item.component.ts
Licen-it: and this is why i was suggesting the directive that would add all the properties if they exists there

klaascuvelier: Ah that's why my editor gave a squiggly line, but I couldn't figure out why because the documentation was pointing to it.
Indeed also clearly confirms the concern @Licen-it voiced. 

For now I removed the info property. Will look into the directive suggestion after this. Want to fix this regression and unblock us first.

Licen-it (approved):

src/ui-shell/sidenav/sidenav-item.component.ts
Akshat55: Yea I understand, it would simplify porting between the versions. It's best to make another PR for the directive targeting the next release branch.

Akshat55 (approved):

Issue #2924, opened 2024-06-14
Future proof routerExtras implementation


klaascuvelier:
The fix done for #2911 is not the most future proof implementation. Whenever Angular decides to change the properties of the `routerExtra` the implementation might break. (As already happened during the implementation of the fix).

This issue serves to find a better approach.

@Licen-it it suggested to create new directive that extends the `routerLink` functionality to also add the routerExtras.
Can you describe how you see this? Is this a new directive like `routerLinkExtras` that should also be added on the a tag? Or do you want it to really `extend` the `routerLink`? Where would this directive live in the repo? 

Thanks 🙇

Licen-it:
@klaascuvelier just pushed the solution

PR #2925, opened 2024-06-14
fix(SideNavItem): implemented more future proof navigation extras without dropping routerlink support

Licen-it:
Closes carbon-design-system/carbon-components-angular#2924

This PR is introducing a more future proof solution to resolve the issue mentioned above. It's creating a directive extending RouterLink directive and setting the relevant navigation extras properties.

#### Changelog

**Changed**

* Implemented RouterLinkExtendedDirective


Licen-it:
@zvonimirfras @Akshat55 for the moment i didn't know where to put the directive, so i placed in the same module where we needed to use, so that it won't affect other routerLinks around carbon, but if you think this should be reused elsewhere as well, please let me know where it would be good to place it. I would avoid to export it in any module so that it won't affect consumers, but only carbon code.

Akshat55 (approved):

angular.json
```diff
@@ -71,7 +71,8 @@
 							"-t"
 						],
 						"configDir": ".storybook",
-						"outputDir": "dist/docs/storybook"
+						"outputDir": "dist/docs/storybook",
+            "enableProdMode": false
```
Akshat55, line 75: Need to set `enableProdMode` to `false` for the story to correctly display. _May_ be addressed by updating to storybook in v8?

 Tested the build in angular 18 and was able to pass navigation extras successfully.

zvonimirfras (approved):

Akshat55 (approved):

Issue #2929, opened 2024-07-01
PropagateSelected becoming slow when adding lot's of items to a dropdown in multiple batches


koenvd:
## Description

When adding items to a dropdown in mulitple batches the `propagateSelected` method takes considerable time to run and becomes slower with a higher number of items.

To reproduce create a dropdown and add items to the dropdown in batches. Amount of items something like 100, 200, 300, ... When you check the performance you will notice the `propagateSelected` taking more and more time to execute. The reason for this is that the `propagateSelected` method does a loop for every item in the `oldItems` list over the complete list of the `newItems` which makes this a O(n^2) algorithm.

There are different ways to improve this:
- Only match selected items would solve my use case since typically the amount of selected items is a lot lower than the total amount of items.
- Ideally we would like to do a fast lookup in a Dictionary based on a lookup key (eg the stringified object). But this doesn't look straightforward because matching items is currently being done on a **substring** of the stringified object.  However I don't get why this is done.  Maybe someone can explain the reasoning of this? If we could drop this substring requirement a very fast lookup in a Dictionary would become possible.

Carbon Design System: 5.26.0

PR #2930, opened 2024-07-01
fix: Optimize propagateSelected by limiting matching items to the selected ones (#2929)

koenvd:
Closes carbon-design-system/carbon-components-angular#2929

Optimizes `propogateSelected` as in the first bullet point explained in issue #2929.  

Ideally we still want to improve this doing a lookup with a specific key instead of looping over all items.  But as a start we only loop over selected items which makes the performance already a lot better since typically the amount of selected items is way less then the total number of items.

Since this my first MR to this project please let me know if I need to add tests or do something extra to get this one merged.

Thanks.


Akshat55:
@koenvd Could you please sign the DCO, simply comment the following text to the PR:

koenvd:

koenvd:
Done @Akshat55 .

Akshat55 (approved):

zvonimirfras (approved):

src/dropdown/list/dropdown-list.component.ts
```diff
@@ -445,24 +445,38 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
...
 				// stringify for compare
 				tempOldItem = JSON.stringify(tempOldItem);
-				for (let newItem of value) {
-					// copy the item
-					let tempNewItem: string | ListItem = Object.assign({}, newItem);
-					// deleted selected because it's what we _want_ to change
-					delete tempNewItem.selected;
-					// stringify for compare
-					tempNewItem = JSON.stringify(tempNewItem);
+				for (let selectedNewItem of selectedNewItems) {
```
Licen-it, line 476: i'm arriving late to the party, but here we could optimise even more as follows (even in terms of readability 😄 )

```ts
oldItem.selected = !!selectedNewItems.find(newItem => tempOldItem.includes(newItem))
```
avoiding the entire for block (for sure `.find` is better optimised than a for of...and we would avoid reassigning `oldItem.selected` to false multiple time when the array is relatively big

PR #2945, opened 2024-07-22
feat: Add readonly state

### LINKED ISSUE #2901: Add ReadOnly state for Text Input

abiramcodes:
Closes #2901 

Added readonly state for the text input. Introduced new  boolean variable as an `@Input()` for the readonly state, updated the storybook as well that reflects the changes.

### DIFF
--- a/src/input/label.component.ts
+++ b/src/input/label.component.ts
@@ -56,7 +56,6 @@ import { PasswordInput } from "./password.directive";
 					[warn]="warn"
 					[warnText]="warnText"
 					[ariaLabel]="ariaLabel"
-					[readonly]="readonly"
 					[labelTemplate]="labelContentTemplate"
 					[textAreaTemplate]="inputContentTemplate">
 				</cds-textarea-label>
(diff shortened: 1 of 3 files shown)


src/input/label.component.ts
```diff
@@ -56,6 +56,7 @@ import { PasswordInput } from "./password.directive";
 					[warn]="warn"
 					[warnText]="warnText"
 					[ariaLabel]="ariaLabel"
+					[readonly]="readonly"
```
Akshat55, line 59: Readonly value is being passed to the `cds-textarea-label` component instead of `cds-text-label`. Additionally, in `cds-text-label` component (see `text-input-label.component.ts`), the host should be assigned `cds--text-input-wrapper--readonly` class.

```ts
@Input() @HostBinding('class.cds--text-input-wrapper--readonly') readonlyWrapper = false;
```

With this class assigned, it should look like this:

image

Akshat55 (changes requested):
Hey @abiramcodes, thanks for taking this on! The issues did not have a lot of detail, sorry about that. I've suggested some changes to help resolve this issue.

src/input/label.component.ts
abiramcodes: Hey @Akshat55 , I have made the changes as per your suggestion.
Please let me know if I am missing anything :)

Akshat55: Sorry for the delay, had to think about this to reduce setting readonly twice. The solution I came up with was to use a getter and apply the class based on `readonly` attribute of the input.

Akshat55 (approved):

Akshat55 (approved):

Issue #2946, opened 2024-07-24
Disabled ComboBox can be cleared clicking on the `x` button


elder-cb:
## Disabled ComboBox can be cleared clicking on the `x` button

> Describe in detail the issue you're having.

Even if the `x` button in a single-select disabled ComboBox is styled as disabled and shows the disabled cursor, clicking it will clear the value.

> Is this a feature request (new component, new icon), a bug, or a general issue?

Bug

> Is this issue related to a specific component?

`ComboBox`

> What did you expect to happen? What happened instead? What would you like to see changed?

I expect that clicking the disabled `x` button will not do anything, but I see the value being cleared.

> What browser are you working in?

Edge, Chrome, Brave

> What version of the Carbon Design System are you using?

Latest

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Zelta Clinical Development: https://www.merative.com/documents/brief/zelta-cro-solution-brief

## Steps to reproduce the issue

github link: https://github.com/elder-cb/carbon-angular-disabled-combobox-bug

1. Create an angular project with carbon components
2. Create a page using a `cds-combo-box`, set its items and its value and mark it as disabled
3. Run the application
4. Click on the `x` button in the ComboBox
5. Observe that the value is being cleared despite the control being disabled

## Additional information

* Project replicating the issue: https://github.com/elder-cb/carbon-angular-disabled-combobox-bug
* Fork + branch containing the fix: https://github.com/elder-cb/carbon-components-angular/tree/fix-disabled-combobox-clear
* Project using the fixed fork showing the fix works: https://github.com/elder-cb/carbon-angular-disabled-combobox-bug/tree/fix-disabled-combobox-clear
* PR with fix: #2947

PR #2947, opened 2024-07-24
fix: prevent combobox from being cleared when disabled

elder-cb:
Closes carbon-design-system/carbon-components-angular#2946

When clicking the clear button, check if the component is disabled before clearing the input.

#### Changelog

**Changed**

* ComboBox: do not clear input when component is disabled


elder-cb:

Akshat55 (approved):

Issue #2965, opened 2024-08-20
Treeview component not fully cloning nodes


Licen-it:
## Detailed description

> Describe in detail the issue you're having.

Treeview component is shallowly cloning nodes passed as input which is then creating the issue of Angular not running change detection on the components.

> Is this a feature request (new component, new icon), a bug, or a general issue?

bug

> Is this issue related to a specific component?

Treeview

> What did you expect to happen? What happened instead? What would you like to see changed?

Angular should correctly run change detection when a node changes

> What version of the Carbon Design System are you using?

latest (5.30.0)

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Zelta

PR #2966, opened 2024-08-20
fix: deeply clone treeNodes to avoid Angular not detecting changes

Licen-it:
Closes carbon-design-system/carbon-components-angular#2965

Deeply clone treeNodes to avoid Angular not detecting changes. Not using lodash cloneDeep to avoid performance issue if any node contains a templateRef.

#### Changelog

**Changed**

* Treeview now deeply cloning treeNodes received as input

### DIFF
--- a/src/treeview/treeview.component.ts
+++ b/src/treeview/treeview.component.ts
@@ -70,7 +70,7 @@ export class TreeViewComponent implements AfterViewInit, OnInit, OnDestroy {
 	 * Passing value will disregard projected content
 	 */
 	@Input() set tree(treeNodes: Node[]) {
-		this._tree = treeNodes.map((node) => Object.assign({}, node));
+		this._tree = treeNodes.map((node) => this.copyNode(node));
 		this.treeViewService.contentProjected = false;
 	}
 
@@ -169,5 +169,13 @@ export class TreeViewComponent implements AfterViewInit, OnInit, OnDestroy {
 
 	public isProjected() {
 		return this.treeViewService.contentProjected;
+	}
+
+	private copyNode(node: Node): Node {
+		const copiedNode = Object.assign({}, node)
+		if (node.children) {
+		  copiedNode.children = node.children.map(child => this.copyNode(child))
+		}
+		return copiedNode
 	}
 }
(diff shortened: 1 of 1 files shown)


src/treeview/treeview.component.ts
```diff
@@ -170,4 +170,13 @@ export class TreeViewComponent implements AfterViewInit, OnInit, OnDestroy {
...
 	}
+
+	private copyNode(node: Node): Node {
+		// making a recursive shallow copy to avoid performance issues when deeply cloning templateRefs if defined in the node
+		const copiedNode = Object.assign({}, node)
+		if (node.children) {
+		  copiedNode.children = node.children.map(child => this.copyNode(child))
+		}
+		return copiedNode
+	}
```
Akshat55, line 181: This works, but any chance we can use cloneDeep from lodash? I believe we have the dependency as part of CCA

Akshat55 (changes requested):

src/treeview/treeview.component.ts
Licen-it: @Akshat55 the answer is in the comment i left there in line 175...i have this solution locally in my project and i was using cloneDeep before, but when it comes to having angular template references in the object...then the cloneDeep is messing up with angular making it not refreshing correctly or extremely low performat...like i had huge wait times before the UI was refreshing

Akshat55, line 181:
```suggestion

	private copyNode(node: Node): Node {
		// making a recursive shallow copy to avoid performance issues when deeply cloning templateRefs if defined in the node
		const copiedNode = Object.assign({}, node)
		if (node.children) {
		  copiedNode.children = node.children.map(child => this.copyNode(child))
		}
		return copiedNode
	}
```
```suggestion

	private copyNode(node: Node): Node {
		// making a recursive shallow copy to avoid performance issues when deeply cloning templateRefs if defined in the node
		const copiedNode = Object.assign({}, node);
		if (node.children) {
		  copiedNode.children = node.children.map(child => this.copyNode(child));
		}
		return copiedNode;
	}
```

Akshat55 (changes requested):

Akshat55 (approved):

Issue #2967, opened 2024-08-20
Treeview not exposing expansion output


Licen-it:
## Detailed description

> Describe in detail the issue you're having.

Treeview component is not exposing node expansion output forcing any user who would need to listen for that output to opt for a `contentProjected` approach involving more code and more handlers

> Is this a feature request (new component, new icon), a bug, or a general issue?

bug

> Is this issue related to a specific component?

Treeview

> What did you expect to happen? What happened instead? What would you like to see changed?

Treeview should output the node expansion event the same way it's doing for the node selection

> What version of the Carbon Design System are you using?

latest (5.30.0)

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Zelta

PR #2968, opened 2024-08-20
feat: Exposing toggle output for treeview

Licen-it:
Closes carbon-design-system/carbon-components-angular#2967

Esposed `toggle` output from Treeview component

#### Changelog

**Changed**

* Treeview component now exposing `toggle` output which emits when a node toggle event occurs

### DIFF
--- a/src/treeview/tree-node.component.ts
+++ b/src/treeview/tree-node.component.ts
@@ -122,6 +122,7 @@ export class TreeNodeComponent implements AfterContentChecked, OnInit, OnDestroy
 	@Input() id = `tree-node-${TreeNodeComponent.treeNodeCount++}`;
 	@Input() active = false;
 	@Input() disabled = false;
+	@Input() selectable = true;
 	@Input() expanded = false;
 	@Input() label: string | TemplateRef<any>;
 	@Input() labelContext: any;
@@ -148,6 +149,7 @@ export class TreeNodeComponent implements AfterContentChecked, OnInit, OnDestroy
 		this.id = node.id ?? this.id;
 		this.active = node.active ?? this.active;
 		this.disabled = node.disabled ?? this.disabled;
+		this.selectable = node.selectable ?? this.selectable;
 		this.expanded = node.expanded ?? this.expanded;
 		this.label = node.label ?? this.label;
 		this.labelContext = node.labelContext ?? this.labelContext;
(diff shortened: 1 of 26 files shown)


src/treeview/tree-node.component.ts
```diff
@@ -207,8 +207,10 @@ export class TreeNodeComponent implements AfterContentChecked, OnInit, OnDestroy
 			this.selected = true;
 			this.active = true;
 			event.target.parentElement.focus();
+			const node = { id: this.id, label: this.label, value: this.value }
 			// Passes event to all nodes to update highlighting & parent to emit
-			this.treeViewService.selectNode({ id: this.id, label: this.label, value: this.value });
+			this.treeViewService.selectNode(node);
+			this.nodeSelect.emit({node, event})
```
Akshat55, line 213:
```suggestion
			this.nodeSelect.emit({...node, event});
```
Could you use the spread operator on node? This is a breaking change for existing users.

Akshat55: Would need to update the interface as well

src/treeview/tree-node.component.ts
```diff
@@ -207,8 +207,10 @@ export class TreeNodeComponent implements AfterContentChecked, OnInit, OnDestroy
 			this.selected = true;
 			this.active = true;
 			event.target.parentElement.focus();
+			const node = { id: this.id, label: this.label, value: this.value }
```
Akshat55, line 210:
```suggestion
			const node = { id: this.id, label: this.label, value: this.value };
```

Akshat55 (changes requested):

src/treeview/tree-node.component.ts
Licen-it: at the moment the interface is returning an object with a property `node` and another `event`...if i use the spread operator there i would return a node with inside an event, but this would be a breaking change with the current interface

Licen-it: oh sorry...now i see the breaking change for the select...yes, i'll push just the node without the event

Akshat55 (changes requested):

src/treeview/treeview.component.ts
```diff
@@ -57,7 +57,8 @@ import { TreeViewService } from "./treeview.service";
 			<ng-template #notProjected>
 				<cds-tree-node
 					*ngFor="let node of tree"
-					[node]="node">
+					[node]="node"
+					(nodetoggle)="onNodeToggle($event)">
```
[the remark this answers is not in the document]
Licen-it: Oh right, similar to the focus node i had to do in the other PR....anyway yes feel free to push directly to the PR.

About renaming it would make more sense indeed, or even better `expandedChange` so that it could be also used the 2 way binding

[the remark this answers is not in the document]
Licen-it: actually on a second thought having an observable in the service is not needed. Listening for `nodeToggle` from the top component only makes sense if we don't use projected content, otherwise the user would just listen on the nodeComponent itself. So i believe we just need to add the listener to `nodetoggle` in the nodeComponent itself for its children and propagate that value through the parent output. (Let me know if it's not clear from my explantation here...i'll try to phrase it better with an example)

Akshat55 (approved):

Issue #2969, opened 2024-08-21
As Developer I'd like to have Treeview nodes actionable but not selectable


Licen-it:
## Detailed description

> Describe in detail the issue you're having.

In my usecase expandable nodes can't be selected as there wouldn't be anything associated with that selection, but the current state of the component is not allowing to have some nodes not selectable. I'm currently running a workaround where on select i just trigger the selection on the previously selected node when the user clicks on an expandable node. This workaround is not ideal and it would be preferable to have an internal solution to set the node not selectable (and maybe when a node is not selectable clicking on it would just expand the node)

> Is this a feature request (new component, new icon), a bug, or a general issue?

feature

> Is this issue related to a specific component?

Treeview

> What version of the Carbon Design System are you using?

latest (5.30.0)

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Zelta

PR #2970, opened 2024-08-22
Treeview nodes can now be not selectable

Licen-it:
Closes carbon-design-system/carbon-components-angular#2969

Implementing the ability to have a node enabled but not selectable, making it expand/compress when clicking on it (only acting applied on nodes with children)

#### Changelog

**New**

* A node with children can now be unselectable, making any interaction to expand/compress only

**Changed**

* moved the assignment of the `treewalker.currentNode` on focus rather than on click to make sure it would still work when the node is not selectable

**Removed**

* removed the assignment of the `treewalker.currentNode` during the selection events

### DIFF
--- a/src/treeview/tree-node.component.ts
+++ b/src/treeview/tree-node.component.ts
@@ -122,6 +122,7 @@ export class TreeNodeComponent implements AfterContentChecked, OnInit, OnDestroy
 	@Input() id = `tree-node-${TreeNodeComponent.treeNodeCount++}`;
 	@Input() active = false;
 	@Input() disabled = false;
+	@Input() selectable = true;
 	@Input() expanded = false;
 	@Input() label: string | TemplateRef<any>;
 	@Input() labelContext: any;
@@ -147,6 +148,7 @@ export class TreeNodeComponent implements AfterContentChecked, OnInit, OnDestroy
 		this.id = node.id ?? this.id;
 		this.active = node.active ?? this.active;
 		this.disabled = node.disabled ?? this.disabled;
+		this.selectable = node.selectable ?? this.selectable;
 		this.expanded = node.expanded ?? this.expanded;
 		this.label = node.label ?? this.label;
 		this.labelContext = node.labelContext ?? this.labelContext;
@@ -186,8 +188,8 @@ export class TreeNodeComponent implements AfterContentChecked, OnInit, OnDestroy
 	ngOnInit(): void {
 		// Highlight the node
 		this.subscription = this.treeViewService.selectionObservable.subscribe((value: Map<string, Node>) => {
-			this.selected = value.has(this.id);
-			this.active = this.selected;
+			this.selected = this.selectable && value.has(this.id);
+			this.active = this.selectable && this.selected;
 		});
 	}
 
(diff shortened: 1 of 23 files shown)


src/treeview/tree-node.component.ts
```diff
@@ -204,11 +206,15 @@ export class TreeNodeComponent implements AfterContentChecked, OnInit, OnDestroy
...
-			// Passes event to all nodes to update highlighting & parent to emit
-			this.treeViewService.selectNode({ id: this.id, label: this.label, value: this.value });
+			if (this.selectable || this.children.length === 0) {
+				this.selected = true;
+				this.active = true;
+				// Passes event to all nodes to update highlighting & parent to emit
+				this.treeViewService.selectNode({ id: this.id, label: this.label, value: this.value });
+			} else {
+				this.toggleExpanded(event)
+			}
```
Akshat55, line 217:
```suggestion
			if (this.selectable) {
				this.selected = true;
				this.active = true;
				// Passes event to all nodes to update highlighting & parent to emit
				this.treeViewService.selectNode({ id: this.id, label: this.label, value: this.value });
			} else if(this.children.length > 0) {
				this.toggleExpanded(event)
			}
```

src/treeview/tree-node.component.ts
```diff
@@ -236,7 +242,9 @@ export class TreeNodeComponent implements AfterContentChecked, OnInit, OnDestroy
 	}
 
 	emitFocusEvent(event) {
-		this.nodeFocus.emit({ node: { id: this.id, label: this.label, value: this.value }, event });
+		const node = { id: this.id, label: this.label, value: this.value }
+		this.nodeFocus.emit({ node, event });
+		this.treeViewService.focusNode(node)
```
Akshat55, line 247:
```suggestion
		this.treeViewService.focusNode(node);
```

Akshat55 (changes requested):

src/treeview/tree-node.component.ts
Licen-it: This would make the leaf node with `selectable: false` not be actionable...i was thinking more about having this option for expandable nodes as i don't know what could be a valid usecase for leaf nodes not being selectable and being enabled...but if this is a usecase you would want to have...i'm happy to go that way

Akshat55 (changes requested):
Should be good to go after this.

Seems like I didn't submit the requested change message. Will post as comment.

src/treeview/tree-node.types.ts
```diff
@@ -7,6 +7,7 @@ export interface Node {
 	id?: string;
 	active?: boolean;
 	disabled?: boolean;
+	selectable?: boolean;
```
Akshat55, line 10: Could you add a comment to what `selectable` is here? This is so devs don't have to trace the code to figure out what this does.

src/treeview/tree-node.types.ts
```diff
@@ -7,6 +7,8 @@ export interface Node {
 	id?: string;
 	active?: boolean;
 	disabled?: boolean;
+	// selectable is only valid for nodes with childre. If true, the node won't become selected and if clicked it would only expand/compress itself
```
Akshat55, line 10:
```suggestion
	// selectable is only valid for nodes with children. If true, the node won't become selected and if clicked it would only expand/compress itself
```

PR #2978, opened 2024-08-30
feat: readonly state for slider

eduardmarcinco:
Closes carbon-design-system/carbon-components-angular#2898

### DIFF
--- a/src/slider/slider.component.ts
+++ b/src/slider/slider.component.ts
@@ -72,14 +72,20 @@ import { EventService } from "carbon-components-angular/utils";
 				<ng-container *ngIf="!isTemplate(label)">{{label}}</ng-container>
 				<ng-template *ngIf="isTemplate(label)" [ngTemplateOutlet]="label"></ng-template>
 			</label>
-			<div class="cds--slider-container">
+			<div
+			class="cds--slider-container"
+			[ngClass]="{ 'cds--slider-container--readonly': readonly }"
+			>
 				<label [id]="bottomRangeId" class="cds--slider__range-label">
 					<ng-content select="[minLabel]"></ng-content>
 				</label>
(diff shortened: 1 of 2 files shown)


src/slider/slider.component.ts
```diff
@@ -72,14 +72,20 @@ import { EventService } from "carbon-components-angular/utils";
 				<ng-container *ngIf="!isTemplate(label)">{{label}}</ng-container>
 				<ng-template *ngIf="isTemplate(label)" [ngTemplateOutlet]="label"></ng-template>
 			</label>
-			<div class="cds--slider-container">
+			<div
+			class="cds--slider-container"
+			[ngClass]="{ 'cds--slider-container--readonly': readonly }"
+			>
```
Akshat55, line 78:
```suggestion
				class="cds--slider-container"
				[ngClass]="{ 'cds--slider-container--readonly': readonly }">
```

src/slider/slider.component.ts
```diff
@@ -276,6 +282,18 @@ export class Slider implements AfterViewInit, ControlValueAccessor {
...
+		this._readonly = v;
+		// for some reason `this.input` never exists here, so we have to query for it here too
+		const inputs = this.getInputs();
+		if (inputs && inputs.length > 0) {
+			inputs.forEach(input => input.readOnly = v);
+		}
+	}
+	get readonly() {
+		return this._readonly;
+	}
```
Akshat55, line 296: > // for some reason `this.input` never exists here, so we have to query for it here too

That's expected, input probably hasn't been rendered when it was called. The entire slider component needs a rewrite imo. :/ Most likely will be done in next major version.

Akshat55 (approved):

PR #2993, opened 2024-09-08
fix: replace 'lodash' with 'lodash-es' to reduce bundle size

danielsogl:
#### Changelog

* Replace lodash with 'lodash-es' dependency to reduce bundle size

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "@floating-ui/dom": "1.6.3",
         "@ibm/telemetry-js": "^1.5.0",
         "flatpickr": "4.6.13",
+        "lodash-es": "4.17.21",
         "tslib": "2.3.0"
       },
       "devDependencies": {
@@ -68,7 +69,6 @@
         "karma-coverage-istanbul-reporter": "3.0.2",
         "karma-jasmine": "4.0.0",
         "karma-jasmine-html-reporter": "1.7.0",
-        "lodash-es": "4.17.21",
         "ng-packagr": "14.2.2",
         "postcss": "8.4.14",
         "postcss-loader": "7.0.0",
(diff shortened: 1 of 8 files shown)


danielsogl:

danielsogl:
recheck

Akshat55:
Hi @danielsogl, any specific reason for removing lodash? I feel like it can bring value to a lot of the components where we need to do array manipulation or deep copies. I agree we can implement our own utilities, but using lodash would be more safer and stable. 

The tests seem to pass so it's definitely safe to remove, just want to know your reasoning and any complications you have run into.

Akshat55 (commented):
Hi @danielsogl, any specific reason for removing lodash? I feel like it can bring value to a lot of the components where we need to do array manipulation or deep copies. I agree we can implement our own utilities, but using lodash would be more safer and stable. 

The tests seem to pass so it's definitely safe to remove, just want to know your reasoning and any complications you have run into.

danielsogl:
@Akshat55 `lodash ` adds a huge footprint to the library bundle. Another issue is that `lodash-es` is also added as a dependency to the project. You don’t need to use two implementations of the same library, or you could opt for no implementation at all.

Akshat55:
> @Akshat55 `lodash ` adds a huge footprint to the library bundle. Another issue is that `lodash-es` is also added as a dependency to the project. You don’t need to use two implementations of the same library, or you could opt for no implementation at all. `lodash-es` would keep the bundle small.

You're right, my initial plan was to replace `lodash-es` with just `lodash` but did not take the overall size into account. `lodash-es` is not exported with the code, only `lodash` is, but we can replace this. For now, I think it would be better to replace `lodash` with `lodash-es` instead of completely removing it. What are your thoughts?

Akshat55:
@danielsogl Is it alright if I make changes to this PR to replace `lodash` with `lodash-es`?

danielsogl:
@Akshat55 yes feel free to make changes :)

Akshat55 (approved):

PR #2997, opened 2024-09-18
fix: add rel attribute to clickable tile anchor tag

### LINKED ISSUE #2996: anchor rel tag missing in clickable tile for security
## Detailed description

cds-clickable-tile has href target properties, but it is missing the Security HTML anchor rel when target is set

general issue

Clickable Tile

rel attribute is added

Chrome, Firefox, safari

latest

grishma-shah7:
closes #2996

This PR adds the rel attribute to the anchor tag in the clickable tile for security

Similar security also observed in the anchor tag for overflow menu options component

### DIFF
--- a/src/tiles/clickable-tile.component.ts
+++ b/src/tiles/clickable-tile.component.ts
@@ -8,6 +8,11 @@ import {
 import { Router } from "@angular/router";
 
 /**
+ * Security HTML anchor rel when target is set
+ */
+const REL = "noreferrer noopener";
+
+/**
  * Build application's clickable tiles using this component. Get started with importing the module:
  *
  * ```typescript
@@ -34,6 +39,7 @@ import { Router } from "@angular/router";
 		(click)="navigate($event)"
 		[attr.href]="disabled ? null : href"
 		[attr.target]="target"
+		[attr.rel]="rel"
 		[attr.aria-disabled]="disabled">
 		<ng-content></ng-content>
 	</a>`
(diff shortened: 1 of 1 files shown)


grishma-shah7:

src/tiles/clickable-tile.component.ts
```diff
@@ -7,6 +7,11 @@ import {
 } from "@angular/core";
 import { Router } from "@angular/router";
 
+/**
+ * Security HTML anchor rel when target is set
+ */
+const REL = "noreferrer noopener";
```
yinglejia, line 13: We shouldn't have a default like this, it needs to be explicitly passed by caller.

src/tiles/clickable-tile.component.ts
```diff
@@ -34,6 +39,7 @@ import { Router } from "@angular/router";
 		(click)="navigate($event)"
 		[attr.href]="disabled ? null : href"
 		[attr.target]="target"
+		[attr.rel]="rel"
```
yinglejia, line 42: Define `rel` as an input property and use something like `[attr.rel]="rel ? rel : null"`

yinglejia (changes requested):

src/tiles/clickable-tile.component.ts
grishma-shah7: Sure, have added this
I had thought that having the rel attribute for all links would be helpful for usage when rel is not mentioned

yinglejia (approved):

Akshat55 (approved):

PR #3008, opened 2024-09-25
feat: fluid state for dropdown

eduardmarcinco:
Closes carbon-design-system/carbon-components-angular#2887

### DIFF
--- a/src/dropdown/dropdown.component.ts
+++ b/src/dropdown/dropdown.component.ts
@@ -69,7 +69,7 @@ import { hasScrollableParents } from "carbon-components-angular/utils";
 	<div
 		class="cds--list-box"
 		[ngClass]="{
-			'cds--dropdown': type !== 'multi',
+			'cds--dropdown': type !== 'multi' && !(skeleton && fluid),
 			'cds--multiselect': type === 'multi',
 			'cds--multi-select--selected': type === 'multi' && getSelectedCount() > 0,
 			'cds--dropdown--light': theme === 'light',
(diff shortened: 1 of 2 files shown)


src/dropdown/dropdown.component.ts
```diff
@@ -319,6 +324,28 @@ export class Dropdown implements OnInit, AfterContentInit, AfterViewInit, OnDest
...
+	/**
+	 * Experimental: enable fluid state
+	 */
+	@HostBinding("class.cds--list-box__wrapper--fluid") @Input() fluid = false;
+
+	/**
+	 * Style needed to properly display label in fluid layout.
+	 * React implementation is using a div that is a block-level element.
+	 */
+	@HostBinding("style.display") display = "block";
```
Akshat55, line 339: We normally make a PR to @carbon/styles package and include `display` property with block. We can also use `:host` pseudo class, but I think we should avoid adding styles to maintain on our end. 

I'll make a PR in @carbon/styles and manually update the version here when it is merged. Will also look at other PRs to see if they contain display properties.

eduardmarcinco: Okay, it makes sense to have the style updated upstream. Thanks

Other PRs dont update the display property. Just this one.

Akshat55: Sorry for the delay, waiting on one more approval on this. Hopefully it gets merged by end of this week!

Akshat55: It's been merged! Will be released at the end of this week @1.71.0 🎉

PR #3009, opened 2024-09-25
feat: fluid state for select

eduardmarcinco:
Closes carbon-design-system/carbon-components-angular#2890

### DIFF
--- a/src/select/select.component.ts
+++ b/src/select/select.component.ts
@@ -132,16 +132,18 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from "@angular/forms";
 					size="16"
 					class="cds--select__invalid-icon cds--select__invalid-icon--warning">
 				</svg>
-				<hr *ngIf="fluid" class="cds--select__divider" />
-				<div
-					*ngIf="invalid && invalidText && fluid" role="alert" class="cds--form-requirement" aria-live="polite">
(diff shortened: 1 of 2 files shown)


src/select/select.component.ts
```diff
@@ -116,13 +132,23 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from "@angular/forms";
...
+				<hr *ngIf="fluid" class="cds--select__divider" />
+				<div
+					*ngIf="invalid && invalidText && fluid" role="alert" class="cds--form-requirement" aria-live="polite">
+					<ng-container *ngIf="!isTemplate(invalidText)">{{invalidText}}</ng-container>
+					<ng-template *ngIf="isTemplate(invalidText)" [ngTemplateOutlet]="invalidText"></ng-template>
+				</div>
+				<div *ngIf="!invalid && warn && fluid" class="cds--form-requirement">
+					<ng-container *ngIf="!isTemplate(warnText)">{{warnText}}</ng-container>
+					<ng-template *ngIf="isTemplate(warnText)" [ngTemplateOutlet]="warnText"></ng-template>
+				</div>
```
Akshat55, line 144: I think we should simplify this and wrap it in a `ng-container` with a condition that checks for `fluid`. Thoughts?

eduardmarcinco: I agree, it will group elements related to fluid layout and reduce cognitive load when going through the code.

Issue #2892, opened 2024-05-16
Add fluid state for Text Input/Password


Akshat55:
ref: https://github.com/carbon-design-system/carbon-components-angular/issues/2769

PR #3011, opened 2024-09-25
feat: add fluid state for text input and password

eduardmarcinco:
Closes carbon-design-system/carbon-components-angular#2892

Notes: 
- @carbon/styles version update is required. Fixes for password input were released recently.
- a way of displaying skeleton was updated as now it needs to handle standard skeleton and fluid skeleton depending on the input property

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -25,16 +25,33 @@
 	document.title = "Carbon Components Angular";
 </script>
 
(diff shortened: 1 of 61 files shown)


brampeirs:
Q: does storybook still compile for you on version 1.65 for the styles? For me I get a webpack error. When changing version to 1.67 everything works again.

Akshat55 (approved):

Issue #2891, opened 2024-05-16
Add fluid state for Text Area


Akshat55:
Ref: https://github.com/carbon-design-system/carbon-components-angular/issues/2769

Issue #2886, opened 2024-05-16
Date picker


Akshat55:

coreset:
@Akshat55  I wanted to check in to see if this issue is still open for fixes. Please let me know if there's anything I can do to help.

Akshat55:
@coreset Doesn't seem to be the case, would you like me to be assigned to it?

coreset:
HI @Akshat55  Yes, please go ahead and assign me to the issue. I’d be happy to start working on it. Looking forward to contributing!

eduardmarcinco:
Hello, I actually tried to implement fluid state for date picker couple week ago but hit a wall due to bloated DOM structure that is described in #2822 
It looks like PR #2823 needs to be merged before fluid state can be implemented.

Just my 2 cents. Maybe you'll find a way to implement it.

coreset:
> Hello, I actually tried to implement fluid state for date picker couple week ago but hit a wall due to bloated DOM structure that is described in [#2822](https://github.com/carbon-design-system/carbon-components-angular/issues/2822) It looks like PR [#2823](https://github.com/carbon-design-system/carbon-components-angular/pull/2823) needs to be merged before fluid state can be implemented.
> 
> Just my 2 cents. Maybe you'll find a way to implement it.

@eduardmarcinco  thank you for sharing your insights! I appreciate the context regarding the challenges you faced with the fluid state for the date picker. I'll take a closer look at PR #2823 and see how it can help in moving forward with this implementation. Your input is valuable, and I'll keep you posted on any progress or discoveries I make. **My goal is fix this without breaking changes.**

PR #3012, opened 2024-09-25
feat: fluid state for text area

eduardmarcinco:
Closes carbon-design-system/carbon-components-angular#2891

A way of displaying skeleton was updated as now it needs to handle standard skeleton and fluid skeleton depending on the input property.

### DIFF
--- a/src/input/text-area.directive.ts
+++ b/src/input/text-area.directive.ts
@@ -28,8 +28,8 @@ export class TextArea {
 	@HostBinding("class.cds--text-area--light") get isLightTheme() {
 		return this.theme === "light";
 	}
-	
+
 	@HostBinding("attr.data-invalid") get getInvalidAttr() {
 		return this.invalid ? true : undefined;
-	};
+	}
 }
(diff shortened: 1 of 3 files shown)


Akshat55:
Removed `cds--text-area--warn` for the time being, want to avoid adding API that we may have trouble removing in future. Want to find a way we can streamline setting warning in ONLY the wrapper level instead of both the wrapper and the textarea. We can probably release this for now as experimental and come back to it.

Akshat55 (approved):

coreset:
HI @eduardmarcinco,
I’ve been working on [issue #2886](https://github.com/carbon-design-system/carbon-components-angular/issues/2886), which focuses on implementing a fluid state for the Date picker. While doing so, I came across your PR to better understand the patterns and practices followed in this repository.😊

In the process, I noticed that when changing the text area to a fluid variant in Storybook, the background of the label also changes. I just wanted to confirm— is this the intended behavior?
I also wanted to ask if there are any specific guidelines or rules I should follow when adding a new style class to this repository (excluding the carbon/style package).

eduardmarcinco:
hi @coreset ,
you can check Carbon React implementation here, to see how fluid state looks like -> https://react.carbondesignsystem.com/?path=/docs/experimental-fluid-components-unstable-fluiddatepicker--overview

Also check it's source code at https://github.com/carbon-design-system/carbon

Based on that you would know what classes to use in what scenarios.

Issue #2726, opened 2023-10-26
Breadcrumb overflow menu icon misaligned


rhythm55:
## Detailed description

> Describe in detail the issue you're having.

Breadcrumb oveflow menu icon is not aligned correctly with other breadcrumb items

> Is this a feature request (new component, new icon), a bug, or a general issue?
styling issue

> Is this issue related to a specific component?
Breadcrumb

> What did you expect to happen? What happened instead? What would you like to see changed?
Correct alignment

> What browser are you working in?
Chrome/firefox

> What version of the Carbon Design System are you using?
   "carbon-components-angular": "^4.72.2"

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

## Steps to reproduce the issue

1. check overflow menu icon for breadcrumb

## Additional information
<img width="1792" alt="Screenshot 2023-10-26 at 1 18 01 PM" src="https://github.com/carbon-design-system/carbon-components-angular/assets/36883992/5370a1f1-2b9a-4bbf-a97e-d8ea5958d60e">

nipun005:
Hi @rhythm55 , Can you please provide the screenshot of the expected behaviour? Do we need to display Breadcrumb overflow icon at end ??

rhythm55:
> Hi @rhythm55 , Can you please provide the screenshot of the expected behaviour? Do we need to display Breadcrumb overflow icon at end or Do you just want to change the view of Breadcrumb overflow icon display from vertical to horizontal?? Please provide some more details and assign it to me, I will fix and raise PR.

@nipun005 expected is it should be vertically aligned with other breadcrumb items

abiramcodes:
Hello @Akshat55 , I am able to fix the vertical align issue in the storybook : https://angular.carbondesignsystem.com/?path=/story/components-breadcrumb--model . 

However, there are `no css/scss files in this repository for any components.,` 
**could you please help me / navigate on how to contribute for this issue ?** 

`Before (changing the css in dev tools) : `

![Screenshot 2023-12-11 214651](https://github.com/carbon-design-system/carbon-components-angular/assets/131433061/8c752ec4-3231-4e75-97bb-fce926f15a5f)

`After (changing the css in dev tools) : `

![Screenshot 2023-12-11 214703](https://github.com/carbon-design-system/carbon-components-angular/assets/131433061/44c28583-3a2d-467e-a0fe-5bfb369117b3)


`I would be really happy to contribute : )`

[...]

PR #3020, opened 2024-10-02
fix: breadcrumb overflow menu icon alignment

eduardmarcinco:
Closes carbon-design-system/carbon-components-angular#2726

### DIFF
--- a/src/breadcrumb/breadcrumb.component.ts
+++ b/src/breadcrumb/breadcrumb.component.ts
@@ -71,22 +71,34 @@ const MINIMUM_OVERFLOW_THRESHOLD = 4;
 				</ng-template>
 			</cds-breadcrumb-item>
 			<cds-breadcrumb-item>
-				<cds-overflow-menu>
-					<li class="cds--overflow-menu-options__option"
-						*ngFor="let item of overflowItems">
-						<a class="cds--overflow-menu-options__btn"
-							href="{{item?.href}}"
-							(click)="navigate($event, item)"
-							style="text-decoration: none;">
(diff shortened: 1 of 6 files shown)


src/breadcrumb/breadcrumb.component.ts
```diff
@@ -71,22 +71,34 @@ const MINIMUM_OVERFLOW_THRESHOLD = 4;
...
+								<ng-container *ngIf="!item?.template">{{item?.content}}</ng-container>
+								<ng-template
+									*ngIf="item?.template"
+									[ngTemplateOutlet]="item?.template"
+									[ngTemplateOutletContext]="{ $implicit: item }">
+								</ng-template>
+							</a>
+						</li>
+					</cds-overflow-menu>
+				</cds-icon-button>
```
Akshat55, line 101: Functionality works, but this fails accessibility. This is because we have a button within a button. Still thinking of how we can fix this... may need a custom solution or need to extend overflow menu perhaps where we project content for the overflow menu trigger.

I'll look into this and get back to you on how we can approach this.

Akshat55 (changes requested):

src/breadcrumb/breadcrumb.component.ts
Akshat55: I'm going to make adjustments to the PR. It seems like OverflowMenu now supports tooltip, so we take the same approach as icon-button for it.

Akshat55 (approved):

zvonimirfras (approved):

Issue #3013, opened 2024-09-25
[Bug] Wrong page options number in pagination


Golosay:
I use table together with pagination for displaying search results.
When I have the **X** number of pages for results, and then I'm setting new data with **X** number of rows (same number as number of pages in previous search results), I'm seeing still X number of pages in pagination page select.

The issue is with this code:
https://github.com/carbon-design-system/carbon-components-angular/blob/6f064962787e801d02f992ca1732525507ed3903/src/pagination/pagination.component.ts#L323

Because this._pageOptions.length !== this.totalDataLength is false, as totalDataLength is the same number as a number of pages before. So, this getter is never recalculated.

<img width="1556" alt="image" src="https://github.com/user-attachments/assets/03985c49-f445-4f06-9905-4c63edf4a8eb">

PR #3021, opened 2024-10-02
fix: update pagination pageOptions calculation logic

eduardmarcinco:
Closes carbon-design-system/carbon-components-angular#3013

**Changed**

The logic that calculates number of pages based on `totalDataLenght` and `itemsPerPage` was updated.
It covers the case in the #3013 
Also covers a scenario when `totalDataLenght` is `0` then `numberOfPages` should be `1`.
New array will be constructed only if number of pages changes. It improves performance as previously `Array` was constructed every time getter was called.

**Added**

* a test to cover this scenario

### DIFF
--- a/src/pagination/pagination.component.spec.ts
+++ b/src/pagination/pagination.component.spec.ts
@@ -152,7 +152,7 @@ describe("Pagination", () => {
 	});
 
 	/**
-	 * GH issue #3013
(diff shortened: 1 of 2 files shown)


Golosay:
Could someone review/approve this PR?)
And thanks to @eduardmarcinco for this fix!

Akshat55 (changes requested):

Akshat55 (approved):

PR #3035, opened 2024-10-25
feat: enhance increment and decrement functionality number component

brampeirs:
Notes:
- Added logic to `onIncrement` and `onDecrement` to constrain values within min and max bounds
- Updated tests to verify min/max constraints
- Conform with React implementation

### Increment
Value | Step | Min | Max | Result before | Result after
-- | -- | -- | -- | -- | --
0 | 10 | 15 |   | 10 | 15
95 | 10 |   | 100 | 95 | 100
105 | 10 |   | 100 | 105 | 100

### Decrement

Value | Step | Min | Max | Result before | Result after
-- | -- | -- | -- | -- | --
20 | 2 |   | 15 | 18 | 15
6 | 2 | 5 |   | 6 | 5
10 | 2 | 15 |   | 10 | 15

https://github.com/user-attachments/assets/dc16654b-c70e-4071-ad19-e6f9de0efbb6

### DIFF
--- a/src/number-input/number.component.spec.ts
+++ b/src/number-input/number.component.spec.ts
@@ -152,9 +152,9 @@ describe("Number", () => {
 		expect(component.value).toEqual(0);
 	});
 
-	it("should decrement and set value to max if value - step is less than max", () => {
+	it("should decrement and set value to max if value - step exceeds max", () => {
 		fixture.detectChanges();
-		buttonUp = fixture.debugElement.query(By.css(".up-icon")).nativeElement;
+		buttonUp = fixture.debugElement.query(By.css(".down-icon")).nativeElement;
 		component.value = 20;
 		component.step = 2;
 		component.max = 15;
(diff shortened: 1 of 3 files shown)


src/number-input/number.component.spec.ts
```diff
@@ -129,6 +152,29 @@ describe("Number", () => {
 		expect(component.value).toEqual(0);
 	});
 
+	it("should decrement and set value to max if value - step is less than max", () => {
+		fixture.detectChanges();
+		buttonUp = fixture.debugElement.query(By.css(".up-icon")).nativeElement;
```
Akshat55, line 157: Should this be `.down-icon`? 🤔

Akshat55 (changes requested):
LGTM! Just a question about one of the tests.

src/number-input/number.component.spec.ts
brampeirs: Good catch, I'll update the test. Oversight from my side.

brampeirs: fixed, could you review once more?

Akshat55 (approved):

zvonimirfras (approved):

Issue #3003, opened 2024-09-23
Memory leak while using an overflow menu inside the tree view component while navigating


Ornco:
## Detailed description

> Describe in detail the issue you're having.

The design I am working on needs an overflow menu inside the tree view component which seems to be causing a memory leak coming along with navigation

> Is this a feature request (new component, new icon), a bug, or a general issue?

Seems to be a bug

> Is this issue related to a specific component?

Yes - the  overflow menu component

> What did you expect to happen? What happened instead? What would you like to see changed?

I would expect there to be no memory leak.

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

5.28.3

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Zelta

## Steps to reproduce the issue
Create a tree view with a node list and render an overflow menu with each node.

1. Check the sandbox link below and open the preview in it's own tab
2. Start profiling the performance 
3. Click the 'Click me to reset table button' a few times (the more you press the better, maybe 20/30+).
4. Stop profiling and view the leak.

<img width="1005" alt="image" src="https://github.com/user-attachments/assets/cdaf5538-506a-4430-8347-fa384a9883c8">

https://codesandbox.io/p/sandbox/carbon-components-angular-forked-mpx7kf

## Additional information
A similar leak was also seen for overflow menus inside the cds-table-toolbar-content component and for overflow menus in nested tables

[Update]
I'm not sure if it is related but I attached another sandbox for a similar issue noticed with combo boxes containing a large number of options (>500) in our case. This one isn't just a memory leak but also has performance implications. The more there are the worse the performance gets. The memory leak can be seen by adding more or simply toggling them on/off via the buttons at the top. If i need to open another issue please let me know. Otherwise I'll just leave this here for now. Thanks in advance.

https://codesandbox.io/p/sandbox/carbon-components-angular-forked-mz84kd

eduardmarcinco:
This one is interesting. 
I tracked the memory leak down to `EventService`. This service is used to track subscriptions to event listeners for DOM events, but service never unsubscribes from them and garbage collector cannot clean them up properly.

[EventService](https://github.com/carbon-design-system/carbon-components-angular/blob/master/src/utils/event.service.ts) implements `ngOnDestroy` but `EventService` is in providers array for `UtilsModule` [(defined here)](https://github.com/carbon-design-system/carbon-components-angular/blob/master/src/utils/utils.module.ts#L42).

[...]

PR #3038, opened 2024-10-29
fix: mark event service as deprecated and unsubscribe from fromEvent observables

Akshat55:
Closes carbon-design-system/carbon-components-angular#3003

Event service will be removed in v6 in favor of using `fromEvent` (which is what EventService is using under the hood). In v6, dialog directive won't be used for overflow menu when we switch to using `floating-ui` for positioning.

#### Changelog

**Changed**

* Removed usage of `EventService` in dialog in favor of simply calling FromEvent and unsubscribing on destroy.

### DIFF
--- a/src/dialog/dialog.directive.ts
+++ b/src/dialog/dialog.directive.ts
@@ -17,6 +17,7 @@ import { DialogService } from "./dialog.service";
 import { CloseMeta, CloseReasons, DialogConfig } from "./dialog-config.interface";
 import { EventService } from "carbon-components-angular/utils";
 import { Dialog } from "./dialog.component";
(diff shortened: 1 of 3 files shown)


Akshat55:
image

Licen-it (approved):

src/dialog/dialog.directive.ts
```diff
@@ -172,6 +178,7 @@ export class DialogDirective implements OnInit, OnDestroy, OnChanges {
 		this.updateConfig();
 	}
 
+
```
zvonimirfras, line 181:
```suggestion
```

zvonimirfras (approved):

Issue #3059, opened 2024-11-15
[ Enhancement ] : Add minDate and maxDate check in the Datepicker


rkkp1023:
This validation is available on the carbon react components. 
FYI: react carbon components

![Image](https://github.com/user-attachments/assets/c3aa4fb0-c6ec-4b54-98e3-f113f1e441fe)

Note: This is required for one of our use cases.

PR #3060, opened 2024-11-15
chore: added minDate and maxDate checks in the datepicker

rkkp1023:
Closes https://github.com/carbon-design-system/carbon-components-angular/issues/3059

This PR will add the minDate and maxDate to the flatpicker options for the story book

#### Changelog

**New**

image


src/datepicker/datepicker.component.ts
```diff
@@ -226,6 +226,16 @@ export class DatePicker implements
...
+	/**
+	 * The minimum date that a user can start picking from.
+	 */
+	@Input() minDate: string | number;
+
+	/**
+	 * The maximum date that a user can pick to.
+	 */
+	@Input() maxDate: string | number;
+
```
Akshat55, line 238: Hey @rkkp1023, this was intentionally excluded. At the time of the component creation, it was decided that we weren't going to create explicit wrapper inputs due to the possibility of changes upstream (flatpickr). If we start adding flatpickrOptions as inputs, we will then need to add ALL.

Instead, please pass the maxDate/minDate as part of the flatpickrOptions.

rkkp1023: Thanks, @Akshat55, Added flatpickerOptions in the storybook

src/datepicker/datepicker.component.ts
```diff
@@ -225,7 +225,7 @@ export class DatePicker implements
 	@Input() skeleton = false;
 
 	@Input() plugins = [];
-
+
```
Akshat55, line 228:
```suggestion

```

Akshat55 (approved):

src/datepicker/datepicker.component.ts
Akshat55: We tend to follow React implementation for the most part, however there are certain components that we decided to not 100% follow in terms of implementation. Date picker is one of those components. 

Date picker is dependent on a third-party library (flatpickr), so it doesn't make sense for us to add all of those inputs/outputs in case there is a change in flatpickr.

In such case, to resolve this, we would need to phase it out via breaking changes, so the reasonable thing to do on our end was allow for a user to pass in a flatpickr options object.

That being said, I definitely think we do need to **_improve_** our storybook documentation to better showcase this.

Issue #3049, opened 2024-11-06
Tooltip error thrown when switching the disabled state for a cds-icon-button from true to false.


koenvd:
To reproduce create a template containing this:

```
<cds-tooltip [autoAlign]="true" [description]="description">
   <cds-icon-button [disabled]="disabled">
   </cds-icon-button>
</cds-tooltip>
```

Then switch disabled to `true`.  Then to `false`. At that moment the following error is thrown:
![Image](https://github.com/user-attachments/assets/66900549-3c00-45b2-9ced-5bdf33146406)

Location where it fails:
![Image](https://github.com/user-attachments/assets/2cc85ca8-3cb7-44ba-acfa-6058653ae787)

In both cases (enabled and disabled) the tooltip still renders fine though.  But somehow the `Tooltip` doesn't find the popover content when switching to the enabled state.

Fix can be to just play safe and only call `setAttribute` when the popover content was found. So in `tooltip.component.ts` just do
```
this.popoverContentRef = this.elementRef.nativeElement.querySelector(".cds--popover-content");
this.popoverContentRef?.setAttribute("style", "");
```

But I think something more fundamental is going wrong if I read the comments above those lines. 

Probably there's a better fix?

Akshat55:
@koenvd What you're currently doing is wrapping a tooltip within a tooltip and I believe that is causing this issue. For accessibility purpose, starting v11, `cds-icon-button` component is wrapped in `cds-tooltip`.

To resolve this, you can change the above code to the following:

```html
<cds-icon-button
   [disabled]="disabled"
   [autoAlign]="true"
   [description]="description">
   <svg class="cds--btn__icon" cdsIcon="copy" size="16"></svg>
</cds-icon-button>
```

koenvd:
Indeed that resolves the error @Akshat55 .

But I noticed before the tooltip was appearing when the button was disabled. So we could explain in the tooltip why the button was disabled.  Now the tooltip is not appearing when the button is disabled. 

Not sure there's a way to do that?

klaascuvelier:
@Akshat55 when you say 

> I'll see if I can add ...

Does this mean you need to check if it's feasible, or that you want to do it yourself? 
Because I (or anyone else) could be up to take a look at this.

Akshat55:
Check if it's feasible + compare it to v4 implementation.

[...]

PR #3062, opened 2024-11-19
Fix: allow tooltip disabled icon button

klaascuvelier:
Closes carbon-design-system/carbon-components-angular#3049

Fix regression from v4: this will not hide the tooltip for a disabled icon-button.

#### Changelog

**Changed**

* fix: allow tooltip for disabled icon-button

### DIFF
--- a/src/button/icon-button.component.ts
+++ b/src/button/icon-button.component.ts
@@ -27,7 +27,7 @@ import { ButtonSize, ButtonType } from "./button.types";
 	<cds-tooltip
 		class="cds--icon-tooltip"
 		[description]="description"
-		[disabled]="disabled"
+		[disabled]="showTooltipWhenDisabled ? false : disabled"
 		[caret]="caret"
 		[dropShadow]="dropShadow"
 		[highContrast]="highContrast"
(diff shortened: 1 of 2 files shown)


src/button/icon-button.component.ts
```diff
@@ -27,7 +27,6 @@ import { ButtonSize, ButtonType } from "./button.types";
 	<cds-tooltip
 		class="cds--icon-tooltip"
 		[description]="description"
-		[disabled]="disabled"
```
Akshat55, line 30: I think we should continue to keep this, this was intentional. So we should create a an input 
```ts
@Input() showTooltipWhenDisabled = false
```


```html
[disabled]="showTooltipWhenDisabled ? true : disabled"
```

src/button/icon-button.stories.ts
```diff
@@ -134,3 +134,33 @@ WithAutoAlign.args = {
...
+			<svg class="cds--btn__icon" cdsIcon="copy" size="16"></svg>
+		</cds-icon-button>
+	`
+});
+export const DisabledWithTooltip = DisabledWithTooltipTemplate.bind({});
+DisabledWithTooltip.args = {
+	autoAlign: true,
+	description: "Icon Description",
+	align: "top"
+};
```
Akshat55, line 166: Probably don't need a separate story to showcase this

Akshat55 (changes requested):

src/button/icon-button.stories.ts
klaascuvelier: Was for me to make testing easier, but will remove :)

src/button/icon-button.component.ts
klaascuvelier: The logic in your example was off btw, `showTooltipWhenDisabled` should be `false` in the ternary :D

klaascuvelier:
Both remarks have been addressed @Akshat55, not sure if you want me to resolve them or you rather do it yourself after checking it out.

PR #3069, opened 2024-11-21
fix: refresh view after content is initialized

Akshat55:
Resolve expression has changed (NG0100) for input readonly state 

#### Changelog

**Changed**

* Manually trigger change detection when content is initialized

### DIFF
--- a/integration/ng17/angular.json
+++ b/integration/ng17/angular.json
@@ -40,7 +40,7 @@
                 {
                   "type": "initial",
                   "maximumWarning": "500kb",
-                  "maximumError": "1mb"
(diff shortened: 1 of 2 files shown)


integration/ng17/angular.json
```diff
@@ -40,7 +40,7 @@
                 {
                   "type": "initial",
                   "maximumWarning": "500kb",
-                  "maximumError": "1mb"
+                  "maximumError": "2mb"
```
Akshat55, line 43: Had to increase maximum error. We were over budget by 1kb

Issue #3075, opened 2024-11-27
Pagination icon doesn't show when pages > 1000


launderex619:
Hello team, there is a hardcoded number in `pagination.component.ts` in line `140`

![Image](https://github.com/user-attachments/assets/2c942d25-1131-4315-8e12-14b1687ed010)

that makes pagination inconsistent when large amount of data:

Working example:
![Image](https://github.com/user-attachments/assets/0a62718a-723c-446f-bd92-caa38f184e3f)

When large amount of data:
![Image](https://github.com/user-attachments/assets/04c644bd-288b-4826-8a2b-b10edb63567d)

I think the solution might be remove that hardcoded 1000 and use the `pageSelectThreshold` variable

klaascuvelier:
Seems like a mistake indeed. I created a PR to fix this and added some unit tests for the switch between select/input too.

PR #3076, opened 2024-11-28
fix: Remove hardcoded check in pagination select icon condition

klaascuvelier:
Closes carbon-design-system/carbon-components-angular#3075

The check to show/hide the cdsIcon had a hardcoded value, whereas the select box next to it was using `pageSelectThreshold`. The hardcoded value was replaced by the input variable and unit tests where added for this part of the component

#### Changelog

**Changed**
- Replace hardcoded value with `pageSelectThreshold`


Akshat55 (approved):
Good catch!

Issue #3066, opened 2024-11-20
v6: Migrate components to OnPush strategy


Akshat55:
As of now, only a few components support OnPush strategy, most components should be fairly straight forward to migrate. There is a likely hood of Angular moving away from `zone.js` in future releases, this step allows us to move closer to zone-less.

Components to migrate:
 - [ ] Accordion
 - [ ] Breadcrumb
 - [ ] Button (icon)
 - [ ] Checkbox
 - [ ] Combobox
 - [ ] Content switcher
 - [ ] Context-menu
 - [ ] Date picker
 - [ ] Datepicker input
 - [ ] Dropdown
 - [ ] File uploader
 - [ ] Inline loading
 - [ ] Input (Input/Password/Textarea)
 - [ ] Loading
 - [ ] Modal
 - [ ] Notification
 - [ ] Number Input
 - [ ] Pagination
 - [ ] progress bar
 - [ ] Progress Indicator
 - [ ] Radio
 - [ ] Search
 - [ ] Select
 - [ ] Slider (Possible rewrite)
 - [ ] Table
 - [ ] Tabs (Possible rewrite)
 - [ ] Tag
 - [ ] Tiles
 - [ ] Time picker
 - [ ] Time picker select
 - [ ] Toggle
 - [ ] Treeview
 - [ ] UI-shell

Since many components are going to be straight forward, we will accept `batch` PRs.

eduardmarcinco:
Checkbox component is already using on push strategy.
https://github.com/carbon-design-system/carbon-components-angular/blob/5c03216d438e845f78f79f72fabecc76d688a1ed/src/checkbox/checkbox.component.ts#L72

PR #3093, opened 2025-01-23
feat: convert components to use OnPush change detection strategy 

Rrothschild18:

#### Changelog

**Changed**

Select + fix unit tests
Loader
Progress
Number input + fix unit tests

While migrating components to use OnPush change detection, some unit tests started failing. The issue was that the fixture wasn't updating to new input values even when using `fixture.detectChanges()`.

My research revealed that Angular v14 introduced a new API for setting input properties for `fixture.componentRef`: `setInput` ([commit](https://github.com/angular/angular/commit/96c6139c9ab35aa6ab2330a5a79a5906d5c2e8be)) that seems to work as before.

However, using `fixture.componentRef.setInput(propertyName, value)` didn't resolve all test failures. I had to manually trigger change detection in those cases.

If my approach is incorrect, please let me know the best way to handle these unit testing scenarios.

### DIFF
--- a/src/i18n/i18n.service.ts
+++ b/src/i18n/i18n.service.ts
@@ -135,7 +135,9 @@ export class Overridable {
 /**
  * An object of strings, should follow the same format as src/i18n/en.json
  */
-export type TranslationStrings = { [key: string]: string };
+export type TranslationStrings = {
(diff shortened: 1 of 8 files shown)


Rrothschild18:

PR #3094, opened 2025-01-24
feat: convert components to use OnPush change detection strategy 

Rrothschild18:

#### Changelog

**Changed**

Select + fix unit tests
Loader
Progress
Number input + fix unit tests

While migrating components to use OnPush change detection, some unit tests started failing. The issue was that the fixture wasn't updating to new input values even when using `fixture.detectChanges()`.

My research revealed that Angular has an API for setting input properties for `fixture.componentRef`: `setInput` ([commit](https://github.com/angular/angular/commit/96c6139c9ab35aa6ab2330a5a79a5906d5c2e8be)) that seems to work as before.

However, using `fixture.componentRef.setInput(propertyName, value)` didn't resolve all test failures. I had to manually trigger change detection in those cases.

If my approach is incorrect, please let me know the best way to handle these unit testing scenarios.

### DIFF
--- a/src/loading/loading.component.ts
+++ b/src/loading/loading.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, HostBinding } from "@angular/core";
+import { ChangeDetectionStrategy, Component, HostBinding, Input } from "@angular/core";
 import { I18n } from "carbon-components-angular/i18n";
 
 /**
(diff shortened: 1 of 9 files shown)


klaascuvelier:
Nice work!

src/loading/loading.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, Input, HostBinding } from "@angular/core";
+import { ChangeDetectionStrategy, Component, HostBinding, Input } from "@angular/core";
```
Akshat55, line 1:
```suggestion
import {
	ChangeDetectionStrategy,
	Component,
	HostBinding,
	Input
} from "@angular/core";
```

Akshat55 (changes requested):
Awesome work! Thanks for linting and fixing the migrated components as well 🚀 

Just minor linting errors. I'll commit the suggested changes!

Akshat55 (approved):

PR #3100, opened 2025-02-05
feat: convert components to use OnPush change detection strategy 2

Rrothschild18:

#### Changelog

**Changed**

Dropdown + fix unit tests
File Uploader
Inline loading

### DIFF
--- a/src/dropdown/dropdown.component.spec.ts
+++ b/src/dropdown/dropdown.component.spec.ts
@@ -28,7 +28,7 @@ class DropdownTest {
 	model = null;
 	items = [{ content: "one", id: 0, selected: false }, { content: "two", id: 1, selected: false }];
 	selected: ListItem;
-	@Input() isOpen = true;
+	@Input() isOpen = false;
 	onSelect() { }
 }
 
@@ -80,7 +80,7 @@ describe("Dropdown", () => {
 		spyOn(wrapper, "onSelect");
 		fixture.detectChanges();
 		element = fixture.debugElement.query(By.css("cds-dropdown"));
-		fixture.componentRef.setInput("isOpen", false);
+		fixture.componentRef.setInput("isOpen", true);
 		fixture.detectChanges();
 		element.nativeElement.querySelector(".cds--list-box__menu-item__option").click();
 		fixture.detectChanges();
--- a/src/dropdown/dropdown.component.spec.ts
+++ b/src/dropdown/dropdown.component.spec.ts
@@ -18,7 +18,7 @@ import { IconModule } from "../icon/index";
 	<cds-dropdown
 		placeholder="test"
 		class="custom-class"
-		[menuIsClosed]="menuIsClosed"
+		[isOpen]="isOpen"
 		(selected)="onSelect()"
 		[(ngModel)]="model">
 		<cds-dropdown-list [items]="items"></cds-dropdown-list>
@@ -28,7 +28,7 @@ class DropdownTest {
 	model = null;
 	items = [{ content: "one", id: 0, selected: false }, { content: "two", id: 1, selected: false }];
 	selected: ListItem;
-	@Input() menuIsClosed = true;
+	@Input() isOpen = true;
 	onSelect() { }
 }
 
@@ -80,7 +80,7 @@ describe("Dropdown", () => {
 		spyOn(wrapper, "onSelect");
 		fixture.detectChanges();
 		element = fixture.debugElement.query(By.css("cds-dropdown"));
-		fixture.componentRef.setInput("menuIsClosed", false);
+		fixture.componentRef.setInput("isOpen", false);
 		fixture.detectChanges();
 		element.nativeElement.querySelector(".cds--list-box__menu-item__option").click();
 		fixture.detectChanges();
--- a/src/dropdown/dropdown.component.spec.ts
+++ b/src/dropdown/dropdown.component.spec.ts
@@ -1,6 +1,6 @@
-import { Component } from "@angular/core";
+import { Component, Input } from "@angular/core";
 import { TestBed } from "@angular/core/testing";
-import { By	 } from "@angular/platform-browser";
+import { By } from "@angular/platform-browser";
 
 import { Dropdown } from "./dropdown.component";
 import { DropdownList } from "./list/dropdown-list.component";
@@ -18,6 +18,7 @@ import { IconModule } from "../icon/index";
 	<cds-dropdown
 		placeholder="test"
 		class="custom-class"
+		[menuIsClosed]="menuIsClosed"
 		(selected)="onSelect()"
 		[(ngModel)]="model">
 		<cds-dropdown-list [items]="items"></cds-dropdown-list>
@@ -25,9 +26,10 @@ import { IconModule } from "../icon/index";
 })
 class DropdownTest {
 	model = null;
-	items = [{content: "one", id: 0, selected: false}, {content: "two", id: 1, selected: false}];
+	items = [{ content: "one", id: 0, selected: false }, { content: "two", id: 1, selected: false }];
 	selected: ListItem;
-	onSelect() {}
+	@Input() menuIsClosed = true;
+	onSelect() { }
 }
 
 describe("Dropdown", () => {
@@ -47,7 +49,7 @@ describe("Dropdown", () => {
 				IconModule,
 				UtilsModule
 			],
-			providers: [ DropdownService ]
+			providers: [DropdownService]
 		});
 	});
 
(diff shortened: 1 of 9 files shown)


src/dropdown/dropdown.component.spec.ts
```diff
@@ -18,16 +18,18 @@ import { IconModule } from "../icon/index";
...
 		<cds-dropdown-list [items]="items"></cds-dropdown-list>
 	</cds-dropdown>`
 })
 class DropdownTest {
 	model = null;
-	items = [{content: "one", id: 0, selected: false}, {content: "two", id: 1, selected: false}];
+	items = [{ content: "one", id: 0, selected: false }, { content: "two", id: 1, selected: false }];
 	selected: ListItem;
-	onSelect() {}
+	@Input() menuIsClosed = true;
```
Rrothschild18, line 31: The unit tests are using a mock component called `DropdownTest` as a wrapper to the real `Dropdown(cds-dropdown)` component. 

The unit test below is failing due not trigger change detection correctly. It happens because the wrapper component mutate a specific "non-reactive class property" inside Dropdown component and with `changeDetection.Default` it would work as intended, but with OnPush this property is inaccessible.

The property `menuIsClosed` seems to control the state if the dropdown closed or not. Its not a derived state and not specificly bounded to any user event change that would make sense to explicit call `marckForCheck()` somewhere. 


```ts
	it("should propagate the change in selected option back to the form and emit a selected event", () => {
		fixture = TestBed.createComponent(DropdownTest); // An angular PseudoApplicationComponentFixture, access to real component 
		wrapper = fixture.componentInstance;
		spyOn(wrapper, "onSelect");
		fixture.detectChanges();
		element = fixture.debugElement.query(By.css("cds-dropdown")); // Typeof DebugElement has only access to the DOM
		element.componentInstance.menuIsClosed = false; // Directly mutating CarbonComponent component class property 
		fixture.detectChanges();
		element.nativeElement.querySelector(".cds--list-box__menu-item__option").click();
		fixture.detectChanges();
		expect(element.nativeElement.querySelector(".cds--list-box__label").textContent).toEqual("one");
		expect(wrapper.onSelect).toHaveBeenCalled();
		expect(wrapper.model.content).toEqual("one");
		expect(wrapper.model.selected).toBe(true);
	});
```

Solution 1 (implemented) - Update `Dropdown` "non reactive class property"  to be an angular reactive `@Input` and rely on angular change detection by receiving an new @Input reference change. Also updated the docs.

```ts
@Component({
	template: `
	<cds-dropdown
                ...
		[menuIsClosed]="menuIsClosed" //
                ...
                >
		<cds-dropdown-list [items]="items"></cds-dropdown-list>
	</cds-dropdown>`
})
class DropdownTest {
        ...
	@Input() menuIsClosed = true; // Add mock @Input
        ...
}

```

Solution 2 - Re-write all unit tests for not using "wrapper" components since they make testing a little harder with OnPush + class properties that are not `@Inputs`.

Any other solutions are welcome. Im open to suggestions since the component is big and have few unit tests.

src/dropdown/dropdown.stories.ts
```diff
@@ -66,7 +66,8 @@ export default {
 		dropUp: false,
 		size: "md",
 		theme: "dark",
-		fluid: false
+		fluid: false,
+		menuIsClosed: true
```
Rrothschild18, line 70: I've added the new @Input property menuIsClosed to the documentation. However, during my code review, it appeared to be used as internal state, and I'm unsure whether it should be exposed at the docs. @Akshat55

Rrothschild18 (commented):
I've added a few explanations of my thought process.

src/dropdown/list/dropdown-list.component.ts
```diff
@@ -292,6 +299,7 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 			if (this.displayItems) {
 				this.index = 0;
 			}
+			this.changeDetectorRef.markForCheck();
```
Rrothschild18, line 302: Added the `markForCheck()` here because the component `ComboBox` uses `DropdownList` and there is a unit test that checks "should display dropdown list when new items are passed that match the current search string".

Rrothschild18 (commented):
.

src/dropdown/dropdown.component.spec.ts
zvonimirfras: `isOpen` makes more sense than `menuIsClosed`, for sure.
I don't see a need for a setter when we can just use `!isOpen` instead. Possibly do vice-versa during deprecation period.

src/dropdown/dropdown.component.ts
```diff
@@ -377,7 +379,7 @@ export class Dropdown implements OnInit, AfterContentInit, AfterViewInit, OnDest
 	/**
 	 * Set to `true` if the dropdown is closed (not expanded).
 	 */
-	menuIsClosed = true;
+	@Input() menuIsClosed = true;
```
Akshat55, line 382: We can change the following to something like this & update all instaces of `menuIsClosed` calls to use `isOpen`.

```suggestion
	@Input() isOpen = false;

	set menuIsClosed (isClosed: boolean) {
		this.isOpen = !isClosed;
		this.changeDetectorRef.markForCheck();
	}
```

src/dropdown/dropdown.component.ts
```diff
@@ -407,7 +409,7 @@ export class Dropdown implements OnInit, AfterContentInit, AfterViewInit, OnDest
 		protected elementRef: ElementRef,
 		protected i18n: I18n,
 		protected dropdownService: DropdownService,
-		protected elementService: ElementService) {}
+		protected elementService: ElementService) { }
```
Akshat55, line 412: With the above changes, we would also have to inject changeDetection. Also requires some linting here:

```suggestion
		protected elementService: ElementService,
		private changeDetectionRef: ChangeDetectionRef
	) {}
```

src/dropdown/dropdown.stories.ts
Akshat55:
```suggestion
		isOpen: false
```

src/dropdown/dropdown.stories.ts
```diff
@@ -105,6 +106,7 @@ const Template = (args) => ({
 			[disabled]="disabled"
 			[readonly]="readonly"
 			[fluid]="fluid"
+			[menuIsClosed]="menuIsClosed"
```
Akshat55, line 109:
```suggestion
			[isOpen]="isOpen"
```

src/dropdown/dropdown.stories.ts
```diff
@@ -139,6 +141,7 @@ const MultiTemplate = (args) => ({
...
```
Akshat55, line 144:
```suggestion
			[isOpen]="isOpen"
```

src/dropdown/dropdown.component.ts
Akshat55: I still want to run markForCheck here so it doesn't break anything on release. Teams are/could be using ViewChild to update this variable

src/dropdown/list/dropdown-list.component.ts
```diff
@@ -228,7 +231,11 @@ export class DropdownList implements AbstractDropdownView, AfterViewInit, OnDest
 	/**
 	 * Creates an instance of `DropdownList`.
 	 */
-	constructor(public elementRef: ElementRef, protected i18n: I18n) {}
+	constructor(
+		public elementRef: ElementRef,
+		protected i18n: I18n,
+		protected changeDetectorRef: ChangeDetectorRef
+	) { }
```
Akshat55, line 238:
```suggestion
	) {}
```

src/file-uploader/file.component.ts
```diff
@@ -93,7 +95,8 @@ export class FileComponent implements OnDestroy {
 	@HostBinding("class.cds--file__selected-file--lg") get fileSizeLarge() {
 		return this.size === "lg";
 	}
-	constructor(protected i18n: I18n) {}
+
+	constructor(protected i18n: I18n) { }
```
Akshat55, line 99:
```suggestion
	constructor(protected i18n: I18n) {}
```

Akshat55 (changes requested):
Requires some minor changes, really liked the intro of the menuIsClosed input (To be converted to a setter). Thanks for contributing this @Rrothschild18

src/dropdown/dropdown.component.ts
Akshat55: A notice of deprecation should also improve dev experience if they are using it.

src/dropdown/list/dropdown-list.component.ts
Rrothschild18: Thats why -> "typescript.format.insertSpaceAfterOpeningAndBeforeClosingEmptyBraces": false.

Now i think everything will work

src/dropdown/dropdown.component.spec.ts
Rrothschild18: Rename `menuIsClosed` to `isOpen` is way better, it sounded weird for me too. While migrating components im trying to change less code possible. 

Ill implement the suggestion of creating an input setter as suggested.

Akshat55 (approved):

PR #3113, opened 2025-03-02
fix: pass EnvironmentInjector to provide current module services to modals

Akshat55:
#### Changelog

* In CCA v4, modals were able to access ALL providers where the modal was declared & called from. In v5 this is not possible and is considered an issue requiring services to be provided in root. To address this, we are passing in `EnvironmentInjector` which is created for each module and due to Angular's hierarchical inheritance model, root level providers & Injections will also be available. 

**New**

**Changed**

* Pass  `EnvironmentInjector`  to allow services provided in same module where calling component is declared to be accessed.

### DIFF
--- a/src/modal/base-modal.service.ts
+++ b/src/modal/base-modal.service.ts
@@ -1,7 +1,9 @@
 import {
 	ComponentRef,
 	Injector,
-	Injectable
+	Injectable,
+	inject,
+	EnvironmentInjector
 } from "@angular/core";
 import { PlaceholderService } from "carbon-components-angular/placeholder";
 import { tap, delay } from "rxjs/operators";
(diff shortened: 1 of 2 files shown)


klaascuvelier:
Hey, sorry to reach out after this MR was merged, but this changed caused some functionality to break for us.

When you have things that are provided on a  _deeper_ route than where the modal service is provided, those services are no longer accessible in the instantiated modal.
eg.
```Typescript
// app.routes.ts
providers: [ModalService]
```
```Typescript
// nested-routes.ts
component: MyComponent,
providers: [MyCustomService]
```
```Typescript
// my.component.ts
openMyModal() {
   this.modalService.create({ component: MyModalComponent })
}
```
```Typescript
// my-modal.component.ts
myCustomService = inject(MyCustomService);
```

This setup no longer works. 
Probably this change should have been considered a breaking change

Akshat55:
Hi @klaascuvelier , just saw your comment on this [PR #3133](https://github.com/carbon-design-system/carbon-components-angular/pull/3113). Yes you're right, this is a breaking change, this should've been done from the get go when we upgraded (release of v5). Unfortunately, we missed this behavior and had to re-add it.

A solution would be to provide `ModalService` again which will create a new instance with updated `EnvironmentInjector`.

klaascuvelier:
Yes, that's indeed the way we're solving it now. Sadly it's a runtime error, so it's not easy to find occurrences :)

PR #3125, opened 2025-04-02
fix(dialog): Remove the incorrect descendant element

Thejashwini123:
Violation : The element with role "button" contains descendants with roles "dialog" which are ignored by browsers

accessibility-violation

Fix : The accessibility violation occurs because aria-owns moves the dialog inside the button in the accessibility tree, which is not valid for a button element.


Thejashwini123:

PR #3155, opened 2025-07-04
Move to standalone components

tom-wegener:
Updates all components to standalone components

It got a bit big but I've tried to move all components to standalone components. My test-setup is a bit confused with the changes but it works with the tests and storybook.

#### Changelog

**Changed**

* updates all components to standalone components
* updates all tests to work with the standalone components
* updates storybook to work with standalone components
* keeps the module functionality

I'm on vacation for the next two weeks but will look into this from time to time.


src/accordion/accordion.module.ts
```diff
@@ -6,17 +6,15 @@ import { AccordionItem } from "./accordion-item.component";
...
-		AccordionItem
-	],
 	exports: [
 		Accordion,
 		AccordionItem
 	],
 	imports: [
 		CommonModule,
-		IconModule
+		IconModule,
```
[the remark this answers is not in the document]
tom-wegener: Yes, I wanted to keep the modules for backwards compatibility since angular 18 still allows the functionality and I didn’t want to break my own apps by upgrading.

in the end we don’t need them anymore and I/we/soneone could look into the angular migrations to ease the transition from modules to standalone.

georgmu:
@Akshat55 what is your stance or policy regarding backwards compatibility on the next branch? There are a lot of interesting features in current angular releases, also with regards to signals and zoneless.

Would it be okay to make the next branch angular 20+ and drop backwards compatibility? The version numbers here (v5, ...) does not align with carbon upstream (which is at 11) nor with angular (20), so if you have no internal policy of supporting all versions of angular, maybe a clear cut and upgrade with the next branch would do it as well?

Akshat55:
Hey @georgmu, we need backwards compatibility, because many teams have massive applications and are slow to upgrade to latest angular versions + features. However, this should be taken care of with v5 - just means we will have to support v5 until Carbon 12 is released or have enough teams migrated to use v6.

I would be okay with Angular 20 as next branch, it would have to be another PR though.

This PR will take some time to review/merge - I want to have some teams test it possible. I'll also be testing locally via linking.

tom-wegener:
> Hey @georgmu, we need backwards compatibility, because many teams have massive applications and are slow to upgrade to latest angular versions + features. However, this should be taken care of with v5 - just means we will have to support v5 until Carbon 12 is released or have enough teams migrated to use v6.
> 
> I would be okay with Angular 20 as next branch, it would have to be another PR though.

To further support angular with zone.js > 0.14 without dependency resolution problems we maybe have to use the next branch (and version 6) for angular 19/20+ support. I'm also working on this atm and will be opening up at least tow PRs more in the next few days/weeks with the changes seperated in logical chunks.

> This PR will take some time to review/merge - I want to have some teams test it possible. I'll also be testing locally via linking.

Thank you very much for taking your time testing and reviewing it.

src/tag/index.ts
```diff
@@ -3,4 +3,4 @@ export * from "./tag.component";
 export * from "./tag-icon.directive";
 export * from "./tag-selectable.component";
 export * from "./tag-operational.component";
-export * from "./tag.module";
+// export * from "./tag.module";
```
Akshat55, line 6:
```suggestion
export * from "./tag.module";
```

We need to export from index to make it easier to import module

src/tag/tag.module.ts
```diff
@@ -3,26 +3,23 @@ import { CommonModule } from "@angular/common";
...
+	exports: [
 		Tag,
 		TagFilter,
 		TagIconDirective,
 		TagSelectableComponent,
 		TagOperationalComponent
 	],
-	exports: [
-		Tag,
+	imports: [CommonModule, IconDirective, Tag,
```
Akshat55, line 19: Do we still need CommonModule?

Also imports should be on new line

src/theme/index.ts
```diff
@@ -1,2 +1,2 @@
 export * from "./theme.directive";
-export * from "./theme.module";
```
Akshat55, line 2: Could we still keep Theme module? Same as change as the other index files

Akshat55: I actually had an issue with the following code since index file doesn't export Theme module:

```ts
import { ThemeModule } from 'carbon-components-angular';
```

src/timepicker-select/timepicker-select.component.ts
```diff
@@ -1,13 +1,12 @@
 import {
 	Component,
 	Input,
-	Output,
-	EventEmitter,
-	HostBinding,
-	TemplateRef
+	HostBinding
+
```
Akshat55, line 5:
```suggestion
```

src/ui-shell/header/header-item.component.ts
```diff
@@ -25,7 +26,7 @@ import { Router } from "@angular/router";
 				[ngClass]="{'cds--header__menu-item--current' : isCurrentPage}"
 				[href]="href"
 				(click)="navigate($event)">
-				<ng-container *ngTemplateOutlet="content" />
+				<ng-content></ng-content>
```
Akshat55, line 29:
```suggestion
				<ng-container *ngTemplateOutlet="content" />
```

Undo, we shouldn't use ng-content twice in a condition.

src/ui-shell/header/header-item.component.ts
```diff
@@ -35,15 +36,18 @@ import { Router } from "@angular/router";
 				[ngClass]="{'cds--header__menu-item--current' : isCurrentPage}"
 				[routerLink]="route"
 				[routerLinkActive]="activeLinkClass">
-				<ng-container *ngTemplateOutlet="content" />
+					<ng-content></ng-content>
```
Akshat55, line 39:
```suggestion
					<ng-container *ngTemplateOutlet="content" />
```

Undo

src/accordion/accordion-item.component.ts
```diff
@@ -48,7 +51,10 @@ import {
 			display: list-item;
 		}
 	`],
-	changeDetection: ChangeDetectionStrategy.OnPush
+	changeDetection: ChangeDetectionStrategy.OnPush,
+	standalone: true,
+	providers: [ICON_SERVICE_PROVIDER],
```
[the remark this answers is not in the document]
tom-wegener: I see the point of providing it in root (the current approach is complicated) but on the other hand that would defeat the idea of standalone components, where everything is standalone and I don't have to import it in root. 

We somewhat use the provide in root approach in house but a colleague complained that the file where all icons are declared gets really long in the mid and long term. So I would rather keep the complicated system. But in the end I'll follow your directions to get this merged.

I'll correct the other issues first.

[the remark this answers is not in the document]
tom-wegener: I've slept about it and tbh I'm not sure what I wanted to say with my first sentence.

I agree that providing it in root is the better option for all these services but would keep the icon registration in the components to keep the standalone.

georgmu:
We (including @tom-wegener) have a local version of CCA ported to Angular 21, which has multiple issues solved. This is not a pull request yet, since even this pull request makes no progress.

Locally, we have done the following:
- switch to angular 21 and vitest + jsdom (including minor reworks to tsconfig handling)
- update components and tests to pass the existing test cases
  - use signals (input, output, viewChild, contentChild) where ExpressionChangedAfterItHasBeenChecked was thrown
- type-safe modals (also modal inputs, but only for input signals) through some typescript magic

With some effort, we can share this, but if there is no progress in 6 months, it was not worth the effort to put this in clean commits and ask for pull requests. Or maybe as a fork somewhere else (what we are trying to avoid) if nothing happens...

arctumn:
> We (including @tom-wegener) have a local version of CCA ported to Angular 21, which has multiple issues solved. This is not a pull request yet, since even this pull request makes no progress.
> 
> Locally, we have done the following:
> 
> * switch to angular 21 and vitest + jsdom (including minor reworks to tsconfig handling)
> * update components and tests to pass the existing test cases
>   
>   * use signals (input, output, viewChild, contentChild) where ExpressionChangedAfterItHasBeenChecked was thrown
> * type-safe modals (also modal inputs, but only for input signals) through some typescript magic
> 
> With some effort, we can share this, but if there is no progress in 6 months, it was not worth the effort to put this in clean commits and ask for pull requests. Or maybe as a fork somewhere else (what we are trying to avoid) if nothing happens...

In our case we redid some carbon components to adjust issues as they showed up on our tests, but internally people are talking on dropping CCA in favor of primeNG like some of our pages are already have, we have like 3 pages still on CCA that were updated to 21 because redoing everything in primeNG might cost us to much $$$$ but if the library keeps getting neglected we might have no other option than do the switch and drop CCA

arctumn:
@Akshat55 with those new commits you added, are we close to a possible merge from this PR to next?

PR #3163, opened 2025-07-24
fix: fix switch case for inline loading not working

tom-wegener:
switch case compares the value in the case via === which does not work with || or &&

inline loading did not work since the switch case had some problems but it should be fixed now

Found it while doing visual tests for #3162 

#### Changelog

**Changed**

* inline loading did not work since the switch case had some problems but it should be fixed now.

### DIFF
--- a/src/inline-loading/inline-loading.component.ts
+++ b/src/inline-loading/inline-loading.component.ts
@@ -34,47 +34,40 @@ export enum InlineLoadingState {
 	template: `
 		@if (state !== InlineLoadingState.Hidden) {
 			<div class="cds--inline-loading__animation">
-				@switch (state) {
-					@case (InlineLoadingState.Inactive || InlineLoadingState.Active) {
-						<div
-							class="cds--loading cds--loading--small"
-							[ngClass]="{
-								'cds--loading--stop': state === InlineLoadingState.Inactive
-							}">
-							<svg class="cds--loading__svg" viewBox="0 0 100 100">
-								<circle class="cds--loading__background" cx="50%" cy="50%" r="44" />
-								<circle class="cds--loading__stroke" cx="50%" cy="50%" r="44" />
-							</svg>
-						</div>
-					}
-					@case (InlineLoadingState.Finished) {
-						<svg
-							cdsIcon="checkmark--filled"
-							size="16"
-							class="cds--inline-loading__checkmark-container">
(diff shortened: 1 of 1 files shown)


src/inline-loading/inline-loading.component.ts
```diff
@@ -34,27 +34,27 @@ export enum InlineLoadingState {
 	template: `
 		@if (state !== InlineLoadingState.Hidden) {
 			<div class="cds--inline-loading__animation">
-				@switch (state) {
-					@case (InlineLoadingState.Inactive || InlineLoadingState.Active) {
+				@switch (true) {
+					@case (state === InlineLoadingState.Inactive || state === InlineLoadingState.Active) {
```
Akshat55, line 38: We can probably just use a @if/else if condition instead. If you want to continue to use switch case, you can create a template for the content of `state === InlineLoadingState.Inactive || state === InlineLoadingState.Active` condition and use that condition for both case.

src/inline-loading/inline-loading.component.ts
```diff
@@ -34,27 +34,27 @@ export enum InlineLoadingState {
...
 						<div
 							class="cds--loading cds--loading--small"
 							[ngClass]="{
-								'cds--loading--stop': state === InlineLoadingState.Inactive
+								'cds--loading--stop': state === InlineLoadingState.Inctive
```
Akshat55, line 42:
```suggestion
								'cds--loading--stop': state === InlineLoadingState.Inactive
```

src/inline-loading/inline-loading.component.ts
```diff
@@ -64,14 +64,14 @@ export enum InlineLoadingState {
 				}
 			</div>
 		}
-		@switch(state) {
-			@case(InlineLoadingState.Inactive || InlineLoadingState.Active) {
+		@switch(true) {
+			@case(state === InlineLoadingState.Inactive || state === InlineLoadingState.Active) {
```
Akshat55, line 68: Same as above

Akshat55 (changes requested):

src/inline-loading/inline-loading.component.ts
tom-wegener: Yeah, fair. I've switched to if else because it is less hidden than working that into a condition.

PR #3183, opened 2025-10-01
fix: prevent dropdown from closing when querying

Akshat55:
Prevent dropdown list in combobox from closing when an item is already selected

#### Changelog

**Changed**

* Only close dropdown on organic events
* Update combobox query example to include selected element in list to prevent auto clearing

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -691,8 +691,15 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
 		});
 		this.view.items = this.items;
 		this.updatePills();
+		/**
+		 * @todo - In next major version update to the following:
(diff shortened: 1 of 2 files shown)


src/combobox/combobox.component.ts
```diff
@@ -691,9 +691,8 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
 		});
 		this.view.items = this.items;
 		this.updatePills();
-		// clearSelected can only fire on type=multi
```
Akshat55, line 694: This is false, this function was being called when clear input (x) is clicked. Previously would set the combobox selected element to an empty array which is incorrect for `single`.

Now it sets it to undefined which the dropdown-list emits on clear.

rajeshkumaru (approved):

PR #3201, opened 2025-11-25
fix: styles of context menu item content

krava11i:
Previously when no icons or only select icons are used the item name was diplayed with the 1rem icon width. This fix enables the correct use of context-menu-items with and without icons.

Screenshot From 2025-11-25 11-40-25
Screenshot From 2025-11-25 11-40-49

#### Changelog

**New**

**Changed**

* context-menu-item styles

**Removed**


krava11i:
recheck

PR #3250, opened 2026-02-12
feat(IconService): added missing size fallback to support pictograms

### LINKED ISSUE #3249: IconService not handling pictograms due to an omission of the size property
## Description

When trying to register a pictogram with the icon service it runs in to an error as it is trying to run .toString() on a size property that is non existent on pictograms

## Expected behavior

The icon service should be able to register pictograms as well as icons

## Root cause

Pictorgrams are generated without a size property

## Proposed fix

   If there is no size property to assume the size from the height/width attribute of the icon

Licen-it:
Closes carbon-design-system/carbon-components-angular#3249

Added a fallback if the size property is missing as in `@carbon/pictograms`, so that they can be loaded and displayed correctly with the IconService and IconDirective

#### Changelog

**Changed**

* IconService now support pictograms from `@carbon/pictograms` as well

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,6 +33,7 @@
         "@angular/platform-browser-dynamic": "14.3.0",
         "@angular/router": "14.3.0",
         "@babel/core": "7.9.6",
+        "@carbon/pictograms": "^12.69.0",
         "@carbon/styles": "1.88.0",
         "@carbon/themes": "11.24.0",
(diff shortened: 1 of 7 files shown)


Akshat55 (approved):

Akshat55:
I added a pictogram storybook as an example + testing, should be merged by EOD Thursday!

PR #3377, opened 2026-05-20
fix(Dropdown,Combobox): fixed selection lost updating ListItems when using reactive forms or ngModel

### LINKED ISSUE #3376: Dropdown component racing condition issue with reactive forms
I'm experiencing a racing condition when I use the dropdown component with a reactive form: since I'm relying on the value in the form to perform the selection among the dropdown items, all my ListItems have the property `selected` set to `false`. The issue appears after my save flow, where I need to rebuild the options as some might be new, but at the same time keep the selection that is stored in the reactive form. This flow is triggering a race condition where if the `writeValue` function in the dropdown component is called before the new input of the ListItems is passed to the dropdown-list component, nothing appear selected, as when the options change (even if they didn't really change), they don't re-select the currently selected value if available in the new list

### LINKED ISSUE #3389: RadioGroup not updating selected value if options change
The RadioGroup component is keeping the old reference of the selected radio when new references are passed in the ContentChildren.
This is making the selection disappear

Licen-it:
Closes carbon-design-system/carbon-components-angular#3376
Closes carbon-design-system/carbon-components-angular#3389

Fixed the selection lost when updating the ListItems

#### Changelog

**Changed**

* the selection is restored if the selected items still match items in the option list and the new list doesn't have any selected item already

### DIFF
--- a/src/combobox/combobox.component.ts
+++ b/src/combobox/combobox.component.ts
@@ -481,6 +481,9 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
 
 	protected _isFocused = false;
 
+	private _writtenValue: any;
+	private _isUsingReactiveForms = false;
+
 	/**
 	 * Creates an instance of ComboBox.
 	 */
@@ -524,10 +527,10 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
 					if (!isUpdate(event)) {
 						if (this.itemValueKey && this.view.getSelected()) {
 							const values = this.view.getSelected().map(item => item[this.itemValueKey]);
-							this.propagateChangeCallback(values);
+							this._propagateChange(values);
 						// otherwise just pass up the values from `getSelected`
 						} else {
-							this.propagateChangeCallback(this.view.getSelected());
+							this._propagateChange(this.view.getSelected());
 						}
 						this.selected.emit(event);
 					}
@@ -539,15 +542,15 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
 
 						if (!isUpdate(event)) {
 							if (this.itemValueKey) {
-								this.propagateChangeCallback(event.item[this.itemValueKey]);
+								this._propagateChange(event.item[this.itemValueKey]);
 							} else {
-								this.propagateChangeCallback(event.item);
+								this._propagateChange(event.item);
 							}
 						}
 					} else {
 						this.selectedValue = "";
 						if (!isUpdate(event)) {
-							this.propagateChangeCallback(null);
+							this._propagateChange(null);
 						}
 					}
 					// not guarding these since the nativeElement has to be loaded
@@ -558,6 +561,8 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
 						this.view.filterBy("");
 						this.selected.emit(event.item);
 						this.closeDropdown();
+					} else if (this._isUsingReactiveForms) {
+						this.writeValue(this._writtenValue);
 					}
 				}
 			});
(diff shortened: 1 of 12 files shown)


src/combobox/stories/app-dynamic-list-combobox.component.ts
```diff
@@ -5,6 +5,7 @@ import { Component, AfterViewInit } from "@angular/core";
 	template: `
 		<cds-combo-box
 			[(items)]="items"
+			[itemValueKey]="content"
```
Akshat55: "content" doesn't exist

```suggestion
```

src/combobox/stories/app-mock-query-search.component.ts
```diff
@@ -1,11 +1,12 @@
-import { Component } from "@angular/core";
+import { Component, Input } from "@angular/core";
 
 @Component({
 	selector: "app-mock-query-search",
 	template: `
 		<cds-combo-box
 			appendInline="true"
 			[items]="filterItems"
+			[itemValueKey]="content"
```
Akshat55: variable "content" doesn't exist for binding

```suggestion
```

src/combobox/combobox.component.ts
```diff
@@ -481,6 +481,9 @@ export class ComboBox implements OnChanges, AfterViewInit, AfterContentInit, OnD
 
 	protected _isFocused = false;
 
+	private _writtenValue: any;
+	private _isUsingReactiveForms = false;
```
Akshat55: Based on the changes in this file, I'm confused why this variable is called `_isUsingReactiveForms`. The changes made also impact Template driven forms.

So the question is, do we even need this check/variable? Shouldn't it be fixed for when we have one-way binding as well (non-reactive/non-template-driven -- passing in value and listening for change, updating value)?

src/dropdown/dropdown.component.ts
```diff
@@ -389,6 +394,10 @@ export class Dropdown implements OnInit, AfterContentInit, AfterViewInit, OnDest
 	// primarily used to capture and propagate input to `writeValue` before the content is available
 	private _writtenValue: any = [];
 
+	private _isUsingReactiveForms = false;
```
Akshat55: Same comment as above `_isUsingReactiveForms`.

src/dropdown/dropdown.component.ts
```diff
@@ -514,7 +526,7 @@ export class Dropdown implements OnInit, AfterContentInit, AfterViewInit, OnDest
 					this.view.propagateSelected(newValues);
 				} else {
 					// we can safely assume we're passing an array of `ListItem`s
-					this.view.propagateSelected(value);
+					this.view.propagateSelected(value || [value]);
```
Akshat55, line 529: What is the point of `allowNullValues`? Seems like it's propagating the same value when falsey (similar to line 503). Is this line a fail safe?

Akshat55 (changes requested):

src/combobox/stories/app-dynamic-list-combobox.component.ts
Licen-it: Right...it should have been without `[]`. Great catch

src/dropdown/dropdown.component.ts
Licen-it: At the moment `0` can't be used as the value of an item, as it would fall into the falsy path and everything will be deselected. In my project we used to also use `null` to select the "Select a Value" option and give the user the ability to deselect a value when in single selection mode

src/combobox/combobox.component.ts
Licen-it: Well...considering there is no input that can be used to pass the selection, yes we do need it. My suggestion for the next major is to remove the selected property from the listItem and have it handled as a separate input that would work in tandem with the writeValue flow bringing back the component to only have one source of truth for the selected values

Akshat55: Maybe instead of `_isUsingReactiveForms`, should name it as `_isUsingNgControl` since it's being applied to both template-driven & reactive forms.

Akshat55 (approved):

zvonimirfras (approved):

